PR #1, opened 2018-05-21
Publish :conjure-java-lib to bintray

iamdanfox:
Note, these are new maven coordinates compared to the internal version:
```diff
-com.palantir.conjure:conjure-java-lib
+com.palantir.conjure.java:conjure-lib
```

Running this locally gets:

```
./gradlew publish --dry-run
:conjure-lib:generatePomFileForNebulaPublication SKIPPED
:conjure-lib:processorPath SKIPPED
:conjure-lib:compileJava SKIPPED
:conjure-lib:processResources SKIPPED
:conjure-lib:classes SKIPPED
:conjure-lib:jar SKIPPED
:conjure-lib:sourceJar SKIPPED
:conjure-lib:publishNebulaPublicationToMavenLocal SKIPPED
:conjure-lib:bintrayUpload SKIPPED
:conjure-lib:publish SKIPPED
```

./gradlew publishToMavenLocal:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.palantir.conjure.java</groupId>
  <artifactId>conjure-lib</artifactId>
  <version>ef6ff68.dirty</version>
  <dependencies>
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-databind</artifactId>
      <scope>compile</scope>
      <version>2.7.4</version>
    </dependency>
    <dependency>
      <groupId>com.palantir.ri</groupId>
      <artifactId>resource-identifier</artifactId>
      <scope>compile</scope>
      <version>1.0.1</version>
    </dependency>
    <dependency>
      <groupId>com.palantir.tokens</groupId>
      <artifactId>auth-tokens</artifactId>
      <scope>compile</scope>
      <version>3.0.1</version>
    </dependency>
    <dependency>
      <groupId>com.palantir.remoting-api</groupId>
      <artifactId>errors</artifactId>
      <scope>compile</scope>
      <version>1.8.0</version>
    </dependency>
  </dependencies>
  <name>conjure-java-lib</name>
  <description/>
</project>
```

Also verified it publishes a source jar:
```
[...]

PR #2, opened 2018-05-21
Only publish from clean tags

iamdanfox:
Follow-up to https://github.com/palantir/conjure-java/pull/1

I think we should tag 0.1.0 to test this, as the internal repo has no 0.X releases.

### DIFF
--- a/conjure-lib/build.gradle
+++ b/conjure-lib/build.gradle
@@ -1,8 +1,8 @@
-apply plugin: 'maven-publish'
 apply plugin: 'com.jfrog.bintray'
+apply plugin: 'maven-publish'
 apply plugin: 'nebula.maven-publish'
-apply plugin: 'nebula.source-jar'
 apply plugin: 'nebula.publish-verification'
+apply plugin: 'nebula.source-jar'
 
 dependencies {
     compile 'com.fasterxml.jackson.core:jackson-databind'
@@ -22,7 +22,7 @@ bintray {
     key = System.env.BINTRAY_PASSWORD
     publish = true
     pkg {
-        repo = System.env.CIRCLE_TAG ? 'releases' : 'snapshots'
+        repo = 'releases'
         name = project.name
         userOrg = 'palantir'
         licenses = ['Apache-2.0']
@@ -31,3 +31,6 @@ bintray {
 }
 
 publish.dependsOn bintrayUpload
+bintrayUpload.onlyIf {
+    versionDetails().isCleanTag
+}


ferozco (approved):

conjure-lib/build.gradle
```diff
@@ -1,8 +1,8 @@
-apply plugin: 'maven-publish'
 apply plugin: 'com.jfrog.bintray'
```
uschi2000, line 1: should this project be called conjure-java-lib, similar to the other projects in this repository?

ferozco: we chose to follow the convention introduce by conjure-typescript, where the supporting libraries do not include the name of language they are implementing. i.e. we publish `conjure-client` rather than `conjure-typescript-client`

qinfchen: i think Rob is referring to the project layout consistency? We can rename this repo to `conjure-java-lib` and update this [line](https://github.com/palantir/conjure-java/blob/develop/conjure-lib/build.gradle#L26) to publish artifact as `conjure-lib`

uschi2000: i'm arguing for *some* consistency. if the maven group contains "com.palantir.conjure.java", then I'm fine with subprojects (and thus artifact names) "cli", "core", "lib", or alternatively, "conjure-java-cli", "conjure-java-core", "conjure-java-lib". the mix'n'match is what I don't quite follow. (@qinfchen , please don't depart from the subproject-name == artifact-name convention.)

ferozco: The names i propose are `conjure-java`, `conjure-java-core`, `conjure-lib`. Although still slightly inconsistent within the repo I think it makes consumption clearer. `conjure-lib` is named differently since it is the java implementation of some canonical "conjure-lib". The only language of usage is java  and therefore java does not need to be included in the package name. Meanwhile `conjure-java` and `conjure-java-core` are the implementation of a conjure generator for the java language, that just so happens to be written in java. They are intended to be used in polyglot environments or as part of a larger language agnostic tool. Therefore they should include the language in the package name.

uschi2000: Is there really a canonical conjure-lib? Looks like typescript doesn't have one and the python one is currently called conjure-python-lib and its API is significantly different from the java one ( https://github.palantir.build/foundry/conjure-python/blob/develop/conjure-python-lib/python/conjure/__init__.py )

iamdanfox: @uschi2000 Here are the naming conventions we've been pushing towards:

- one verbose maven group per repo: e.g.
  - `com.palantir.conjure.java`
  - `com.palantir.conjure.typescript`
  - `com.palantir.conjure.python`
- generator clis _include_ target language name because this is the whole reason for the CLI's existence (and would be convenient to invoke manually):
  - `conjure-java`
  - `conjure-typescript`
  - `conjure-python`
- any runtime libraries or clients _don't include_ the language name because it can be assumed that when a language X project pulls in a dependency, the dependency is compatible with language X.  Likewise for typescript/javascript. e.g. 
  - `compile "com.palantir.conjure.java:conjure-lib"`
  - `npm install conjure-client` not `npm install conjure-typescript-client`

The only things that don't comply with this structure right now are http-remoting and the python libraries.  We're intending to decouple a bunch of the python code anyway, so I think we can achieve nice `conjure_lib/conjure_client` naming for python.  I don't think we should spend time trying to rename http-remoting -> conjure-client though!

iamdanfox:
Still getting 403s after tagging 0.1.0:

```
Caused by: org.gradle.api.GradleException: Could not create package 'palantir/releases/conjure-lib': HTTP/1.1 403 Forbidden [message:forbidden]
```

https://circleci.com/gh/palantir/conjure-java/13

conjure-lib/build.gradle
uschi2000: Re naming: The rationale for this convention is pretty involved when compared to a pretty walk-up-friendly `conjure-{lang}-{lib,client,generator}`. I would make a different choice, but am happy to agree if you have a way to document the convention and keep it consistent across repositories.

Re http-remoting: I think the goal should be to make Conjure self-contained, i.e., including generators, clients and servers (if available). Referencing a seemingly disconnected "http-remoting" is pretty confusing. Minimally, I'd like to see a plan for eventually integration http-remoting.

PR #3, opened 2018-05-21
rename conjure-java-cli -> conjure-java

ferozco:


iamdanfox:
I actually quite like the current gradle project name (`:conjure-java-cli`), because it makes it very apparent that you shouldn't really be depending on this project.

What about just keeping this name and publishing an artifact from it with a different name: `com.palantir.conjure.java:conjure-java`?

ferozco:
I think best practice is to keep the project name and published name the same

iamdanfox (approved):

uschi2000:
is there a reason for keeping those branches around post-merge?

PR #9, opened 2018-05-30
Make optional setter covariant

JacekLach:
### Before this PR

```
// imagine this comes from a function and you don't know if it's present or absent
Optional<String> specificValue = Optional.of("foo");
RequestArgs args = RequestArgs.builder()
  // optionalArgument is `optional<any>` because different requests take different types
  .optionalArgument(specificValue)
  .build();
args.getOptionalArgument() // returns Optional.of(Optional.of("foo"))
```

### After this PR

```
Optional<String> specificValue = Optional.of("foo");
RequestArgs args = RequestArgs.builder()
  // optionalArgument is `optional<any>` because different requests take different types
  .optionalArgument(specificValue)
  .build();
args.getOptionalArgument() // returns Optional.of("foo")
```

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
@@ -139,14 +139,14 @@ public final class ListExample {
         }
 
         @JsonSetter("items")
-        public Builder items(Iterable<String> items) {
+        public Builder items(Iterable<? extends String> items) {
             this.items.clear();
             ConjureCollections.addAll(
                     this.items, Objects.requireNonNull(items, "items cannot be null"));
             return this;
         }
 
-        public Builder addAllItems(Iterable<String> items) {
+        public Builder addAllItems(Iterable<? extends String> items) {
             ConjureCollections.addAll(
                     this.items, Objects.requireNonNull(items, "items cannot be null"));
             return this;
@@ -158,7 +158,7 @@ public final class ListExample {
         }
 
         @JsonSetter("primitiveItems")
-        public Builder primitiveItems(Iterable<Integer> primitiveItems) {
+        public Builder primitiveItems(Iterable<? extends Integer> primitiveItems) {
             this.primitiveItems.clear();
             ConjureCollections.addAll(
                     this.primitiveItems,
@@ -166,7 +166,7 @@ public final class ListExample {
             return this;
         }
 
-        public Builder addAllPrimitiveItems(Iterable<Integer> primitiveItems) {
+        public Builder addAllPrimitiveItems(Iterable<? extends Integer> primitiveItems) {
             ConjureCollections.addAll(
                     this.primitiveItems,
                     Objects.requireNonNull(primitiveItems, "primitiveItems cannot be null"));
@@ -179,7 +179,7 @@ public final class ListExample {
         }
 
         @JsonSetter("doubleItems")
-        public Builder doubleItems(Iterable<Double> doubleItems) {
+        public Builder doubleItems(Iterable<? extends Double> doubleItems) {
             this.doubleItems.clear();
             ConjureCollections.addAll(
                     this.doubleItems,
@@ -187,7 +187,7 @@ public final class ListExample {
             return this;
         }
 
-        public Builder addAllDoubleItems(Iterable<Double> doubleItems) {
+        public Builder addAllDoubleItems(Iterable<? extends Double> doubleItems) {
             ConjureCollections.addAll(
                     this.doubleItems,
                     Objects.requireNonNull(doubleItems, "doubleItems cannot be null"));
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
@@ -139,14 +139,14 @@ public final class ListExample {
         }
 
         @JsonSetter("items")
-        public Builder items(Iterable<? extends String> items) {
+        public Builder items(Iterable<String> items) {
             this.items.clear();
             ConjureCollections.addAll(
                     this.items, Objects.requireNonNull(items, "items cannot be null"));
             return this;
         }
 
-        public Builder addAllItems(Iterable<? extends String> items) {
+        public Builder addAllItems(Iterable<String> items) {
             ConjureCollections.addAll(
                     this.items, Objects.requireNonNull(items, "items cannot be null"));
             return this;
@@ -158,7 +158,7 @@ public final class ListExample {
         }
 
         @JsonSetter("primitiveItems")
-        public Builder primitiveItems(Iterable<? extends Integer> primitiveItems) {
+        public Builder primitiveItems(Iterable<Integer> primitiveItems) {
             this.primitiveItems.clear();
             ConjureCollections.addAll(
                     this.primitiveItems,
@@ -166,7 +166,7 @@ public final class ListExample {
             return this;
         }
 
-        public Builder addAllPrimitiveItems(Iterable<? extends Integer> primitiveItems) {
+        public Builder addAllPrimitiveItems(Iterable<Integer> primitiveItems) {
             ConjureCollections.addAll(
                     this.primitiveItems,
                     Objects.requireNonNull(primitiveItems, "primitiveItems cannot be null"));
@@ -179,7 +179,7 @@ public final class ListExample {
         }
 
         @JsonSetter("doubleItems")
-        public Builder doubleItems(Iterable<? extends Double> doubleItems) {
+        public Builder doubleItems(Iterable<Double> doubleItems) {
             this.doubleItems.clear();
             ConjureCollections.addAll(
                     this.doubleItems,
@@ -187,7 +187,7 @@ public final class ListExample {
             return this;
         }
 
-        public Builder addAllDoubleItems(Iterable<? extends Double> doubleItems) {
+        public Builder addAllDoubleItems(Iterable<Double> doubleItems) {
             ConjureCollections.addAll(
                     this.doubleItems,
                     Objects.requireNonNull(doubleItems, "doubleItems cannot be null"));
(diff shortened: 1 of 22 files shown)


JacekLach:
Oh, I need to handle primitives specially here. Will fix later

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -217,6 +224,14 @@ private static CodeBlock typeAwareAssignment(EnrichedField enriched, Type type,
                             ByteBuffer.class)
                     .addStatement("this.$1N.rewind()", spec.name)
                     .build();
+        } else if (type.accept(TypeVisitor.IS_OPTIONAL)) {
```
markelliot, line 321: Why is this change required?

JacekLach: You can't assign a covariant type to an invariant type, like `Optional<? extends Object> bar = getBar(); Optional<Object> foo = bar;`. You need to either explicitly recast and suppress compiler warnings, or transform the generic type with `Optional.<Object>of(bar.get())`

(actually, I checked and it looks like `Optional<Object> foo = (Optional<Object>) bar;` works without java warnings so maybe that's better)
(oh, actually, it does produce a warning but it seems to be suppressed in generated code by default)

markelliot: I see. And you didn't want to widen the type everywhere?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -217,6 +224,14 @@ private static CodeBlock typeAwareAssignment(EnrichedField enriched, Type type,
...
+            Type innerType = type.accept(TypeVisitor.OPTIONAL).getItemType();
+            CodeBlock nullCheckedValue = Expressions.requireNonNull(
+                    spec.name, enriched.fieldName().get() + " cannot be null");
+            return CodeBlock.builder()
+                    .addStatement("this.$1N = $2L.flatMap($3T::<$4T>::of)",
```
markelliot: just filling in the syntax to make it clearer:

```java
foo.flatMap(Optional::<Object>::of)
```

doesn't seem valid to me/locally it does not compile for me.

markelliot: there needs to be some testing around both the code gen and the logic of the codegen in this PR

uschi2000:
what's a concrete use case for this? typically, our conjure objects have concrete types and there's no class hierarchy.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
JacekLach: Yeah, the extra `::` was a typo that I already had fixed locally but forgot to push. I added a minor test but currently the infra for adding new cases is missing; will come back to it after https://github.com/palantir/conjure-java/issues/10 is fixed

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
JacekLach: We could widen the type for optionals everywhere, but that feels a bigger / more scary change to me. If maintainers think that's a better approach, I'm happy to follow suit.

Intuitively I feel like it might require changes in many places; if `ConjureOptional.getItem()` starts returning `Optional<? extends FOO>` then a lot of code that currently expects `Optional<FOO>` will break; and I don't see an advantage to getting a wildcard type when reading

JacekLach:
@uschi2000 the biggest benefit is for `Optional<any>`, which sees some usage.

With current bindings, you get this gotcha:

```
 // imagine this comes from a function and you don't know if it's present or absent
Optional<String> specificValue = Optional.of("foo");
RequestArgs args = RequestArgs.builder()
  // optionalArgument is `optional<any>` because different requests take different types
  .optionalArgument(specificValue)
  .build();
args.getOptionalArgument() // returns Optional.of(Optional.of("foo"))
```

With wildcard builder you get `Optional.of("foo")` in the last example, which I think you more often expect. Of course it means if you actually want a directly nested optional in an untyped field you need to put in some more boilerplate; but that seems a very rare occurrence.

conjure-java-core/src/test/java/com/palantir/conjure/java/CovariantsTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.Test;
+
+public class CovariantsTest {
+
+    @Test
+    public void covariantOptionalBuilderTest() {
+        String value = "foo";
+        Optional<String> maybeValue = Optional.of(value);
+
+        // FIXME: this doesn't compile right now because changes to example-types.yml aren't picked up
```
uschi2000: there's a test here that re-generates the code when you comment out a `if (REGEN)` - ish statement

JacekLach: I believe that regenerates the java code based on the `.json` definitions, but there's nothing that regenerates the `.json` definitions right now, see #10

JacekLach:
Green!

markelliot:
Given the use-case, I wonder if we only want to be covariant for `any` to limit abuse.

JacekLach:
I think it's also possibly handy for things like collections (being able to put an `Optional<SortedSet>` into a `optional<set<string>>`), but it is only in the case of `any` that without this change you transparently use the wrong overload, instead of getting a compilation error. So I'm OK with just doing this for `any` at first, and extending if real use-case shows up.

JacekLach:
Another related annoyance is the static builders for collections:

```
    public static CovariantListExample of(List<Object> items) {
        return builder().items(items).build();
    }
```

but at a quick glance it doesn't seem trivial to widen the parameters there (the `BeanBuilderGenerator` and `BeanGenerator` seem to operate differently)

conjure-java-core/src/integrationInput/java/com/palantir/product/OptionalExample.java
```diff
@@ -99,7 +99,7 @@ public Builder from(OptionalExample other) {
 
         @JsonSetter("item")
         public Builder item(Optional<String> item) {
-            this.item = Objects.requireNonNull(item, "item cannot be null");
+            this.item = (Optional<String>) Objects.requireNonNull(item, "item cannot be null");
```
dansanduleac: this cast doesn't seem necessary

JacekLach: yep, missed out on fixing tests after 5435159

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
dansanduleac: I don't think there's much value in _returning_ `Optional<? extends FOO>`, means you now have to propagate `?` everywhere, and you can't assign it to a variable of type `Optional<FOO>` without a cast.
The value is there only when it's a parameter, since the compiler knows that `FOO` or e.g. `BAR <: FOO` both satisfy `? extends FOO`.

conjure-java-core/src/integrationInput/java/com/palantir/product/OptionalExample.java
JacekLach: 530b826 should address this and fix tests

conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
```diff
@@ -179,15 +179,15 @@ public Builder primitiveItems(int primitiveItems) {
         }
 
         @JsonSetter("doubleItems")
-        public Builder doubleItems(Iterable<Double> doubleItems) {
+        public Builder doubleItems(Iterable<? extends Double> doubleItems) {
```
iamdanfox: As a rule of thumb, we've tried to make conjure-java generate code that is as close as possible to what a human might write.  Can we avoid the `? extends` for known classes like `String`, `Double`, `Integer` etc?

conjure-java-core/src/main/java/com/palantir/conjure/spec/ConjureDefinition.java
```diff
@@ -158,14 +158,14 @@ public Builder version(int version) {
         }
 
         @JsonSetter("errors")
-        public Builder errors(Iterable<ErrorDefinition> errors) {
+        public Builder errors(Iterable<? extends ErrorDefinition> errors) {
```
iamdanfox: I'm also not a fan of this. When a builder accepts a conjure type, we know it is final so there's no point adding `? extends`

iamdanfox:
@JacekLach Dan and I had a couple of CR comments above - what do you think?

JacekLach:
I agree with the sentiment behind them, just didn't have time to look into how to make it happen. The answer is probably to inspect if the type is final somehow.

JacekLach:
@iamdanfox I think this does it.

I'll let the `.of` constructors be for now even though they make me sad, since at least you can now use the builder when you need to :P

conjure-java-core/src/main/java/com/palantir/conjure/spec/EndpointDefinition.java
```diff
@@ -290,14 +290,14 @@ public Builder auth(AuthType auth) {
...
-        public Builder args(Iterable<ArgumentDefinition> args) {
+        public Builder args(Iterable<? extends ArgumentDefinition> args) {
             this.args.clear();
             ConjureCollections.addAll(
                     this.args, Objects.requireNonNull(args, "args cannot be null"));
             return this;
         }
 
-        public Builder addAllArgs(Iterable<ArgumentDefinition> args) {
+        public Builder addAllArgs(Iterable<? extends ArgumentDefinition> args) {
```
iamdanfox: I think these shouldn't have changed?

iamdanfox (approved):
I think this makes sense to me as it prevents the nasty footgun you encountered (putting `Optional.of(x)` into a builder method, resulting in an unwanted double-nested optional).

The code necessary to achieve this does seem pretty hard to read, but I think the outcome is probably worth it.

PR #11, opened 2018-05-31
use conjure core.

### LINKED ISSUE #10: Ensure `./gradlew test -Drecreate=true` can re-generate all test inputs
This repo was open-sourced before Conjure was open-sourced, so we had to check in some JSON IR files:

https://github.com/palantir/conjure-java/blob/1e5e14e08c65cba07b0c50281aa8839455f1a36d/conjure-java-core/src/test/java/com/palantir/conjure/java/types/ObjectGeneratorTests.java#L35

Now that conjure is published, we should generate these on demand rather than keeping them checked in.

ferozco:
Closes #10. Consume in-memory IR rather than checked in files


build.gradle
```diff
@@ -74,7 +74,6 @@ subprojects {
     apply plugin: 'com.palantir.baseline-eclipse'
     apply plugin: 'com.palantir.baseline-idea'
     apply plugin: 'com.palantir.baseline-error-prone'
-    apply plugin: 'com.palantir.baseline-class-uniqueness'
```
iamdanfox: I'm a little surprised [the build](https://circleci.com/gh/palantir/conjure-java/43) is failing because I thought we only had dropwizard/jetty on testCompile ?!?  Anyway, you can tell gradle that these two jars are actually the same thing using the following:

```gradle
    configurations.all {
        resolutionStrategy.eachDependency { DependencyResolveDetails details ->
            if (details.requested.group == 'org.glassfish.hk2.external' && details.requested.name == 'javax.inject') {
                details.useTarget group: 'javax.inject', name: 'javax.inject', version: '1'
                details.because "The glassfish shaded copy is only necessary for OSGI environments, which we do not endorse."
            }
        }
    }
```

I'd kinda prefer to keep the plugin switch on if possible! (We're already using this snippet internally)

ferozco: ok awesome! i just wasn't sure how to fix it

iamdanfox: yeah the README is pretty minimal atm https://github.com/palantir/gradle-baseline#class-uniqueness-plugin-compalantirbaseline-class-uniqueness

iamdanfox (approved):

PR #12, opened 2018-06-01
Use chunked transfer encoding for streaming endpoints

ellisjoe:


ellisjoe:
At the moment if a server implementing a method that returns `StreamingOuput` throws at any point after the first bytes have been sent across the wire, the client will treat the response as successful even though it has received a partial stream. [Chunked-Encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding#Rationale) handles these partial failures. Currently looking into if we can force the server to use chunked encoding whenever it see's an http1.1 connection.

ellisjoe:
Turns out this isn't what we want. Jetty actually does the correct thing and uses chunked encoding whenever the content-length isn't known. There's a bug where jersey doesn't tell jetty to abort the http response when there's been an exception mid stream.

PR #15, opened 2018-06-06
[Experiment] Try linting markdown on CI

iamdanfox:
To save code reviewers the job of leaving "nit" comments about indenting etc, I thought it might be worth setting up some kind of autoformatter / linter.

This one seems very extensible (almost too extensible)... you run it locally on one file by running `remark -o readme.md readme.md`.

https://github.com/remarkjs/remark-lint

If we find one that's convenient enough (especially if it has some VSCode integration), I vote we roll it out across more repos!


iamdanfox:
Closing out as non-urgent

PR #19, opened 2018-06-22
Tag (Alias).valueOf(String) with JsonCreator

### LINKED ISSUE #13: Run serialization tests with jackson 2.9.5
An internal product has reported conjure-related serialization tests failing when they updated jackson to 2.9.5 (via http-remoting: https://github.com/palantir/http-remoting/pull/713)

We could run tests against the same version instead of the current 2.7.4.

dansanduleac:
Otherwise the method won't be used by newer jackson-databind e.g. 2.9.6

Fixes #13 

FLUPs:
- [ ] ability to run all tests against multiple versions of jackson https://github.com/palantir/conjure-java/issues/20

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/BearerTokenAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/BearerTokenAliasExample.java
@@ -37,6 +37,7 @@ public final class BearerTokenAliasExample {
         return value.hashCode();
     }
 
+    @JsonCreator
     public static BearerTokenAliasExample valueOf(String value) {
         return new BearerTokenAliasExample(BearerToken.valueOf(value));
     }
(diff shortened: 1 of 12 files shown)


markelliot:
this shouldn't be required

dansanduleac:
@melliot tested with jackson 2.9.6 and it doesn't work. Not sure if it's a bug in that version of jackson or a deliberate change (couldn't find it in the changelog) but shouldn't we be extra clear with our intention of what is a JsonCreator?

dansanduleac:
This is broken by https://github.com/FasterXML/jackson-databind/commit/eb996aa0313bfc13bca5d5aa836dd224e179d8d1 as @robert3005 found.
I'll add a test that fails and the fix on top.

dansanduleac:
Acually, looking at https://github.com/palantir/conjure/blob/develop/wire.md, this maintains behaviour that used to work, but that we never committed to supporting.
Given that only one product internally seems to be relying on this, and the fix on their side would be easy, we can probably ask them to stop relying on unspecced behaviour.

PR #21, opened 2018-06-23
Add generation of back-compat methods for optional query parameters

markelliot:


markelliot:
@bavardage

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -236,4 +343,188 @@ private static ClassName httpMethodToClassName(String method) {
...
+            return CodeBlock.of("$T.emptyMap()", Collections.class);
+        }
+
+        @Override
+        public CodeBlock visitDefault() {
+            throw new IllegalArgumentException("Cannot backfill non-defaultable parameter type.");
+        }
+    };
+
+    private interface DefaultTypeVisitor<T> extends Type.Visitor<T> {
```
markelliot: for ergonomics, I'd love derive4j style visitors, so I could write:

```java
foo.visitor()
    .optional(x -> blargh())
    .elseThrow()
    .visit()
```

markelliot: #22 reduces a lot of the union code/might make it simpler to add friendlier visit patterns

conjure-java-core/src/test/resources/test/api/TestService.java.jersey
```diff
@@ -117,4 +121,42 @@ public interface TestService {
     @GET
     @Path("catalog/integer")
     int testInteger(@HeaderParam("Authorization") AuthHeader authHeader);
+
+    @Deprecated
+    default int testQueryParams(
```
bavardage, line 126: note to self: works fine in feign as described in https://github.com/OpenFeign/feign#static-and-default-methods

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -236,4 +343,188 @@ private static ClassName httpMethodToClassName(String method) {
...
+            return 1;
+        }
+
+        @Override
+        public Integer visitDefault() {
+            return 0;
+        }
+    };
+
+    private static final Type.Visitor<Boolean> TYPE_BACKFILL_PREDICATE = new DefaultTypeVisitor<Boolean>() {
```
bavardage: might be worth (minimal) javadoc here too, given this is a bit more opaque than the above two visitors - I have to scan up to understand what this is

iamdanfox:
Got some context behind this PR from slack:

Every time an API author adds something to their API endpoints (even if it's backwards compatible), all their consumers suffer a compilation break as they now have to supply `Optional.empty()` or something as the last parameter.

This PR tries to reduce these compilation failures by generating `default` interface methods which try to keep old signatures and just pass `Optional.empty()` or `Collection.emptySet()`.

conjure-java-core/src/test/resources/test/api/TestService.java.jersey
```diff
@@ -41,8 +42,8 @@ public interface TestService {
     @Path("catalog/datasets")
     Dataset createDataset(
             @HeaderParam("Authorization") AuthHeader authHeader,
-            CreateDatasetRequest request,
-            @HeaderParam("Test-Header") String testHeaderArg);
+            @HeaderParam("Test-Header") String testHeaderArg,
+            CreateDatasetRequest request);
```
iamdanfox, line 46: This will be a one-off compilation break, but I like that it'll be stable from now onwards :)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -236,4 +343,188 @@ private static ClassName httpMethodToClassName(String method) {
                 throw new IllegalArgumentException("Unrecognized HTTP method: " + method);
         }
     }
+
+    private static final ParameterType.Visitor<Boolean> QUERY_PARAM_PREDICATE = new ParameterType.Visitor<Boolean>() {
```
iamdanfox: Use `ParameterTypeVisitor.IS_QUERY` instead of this

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
iamdanfox: This visitor and the one above are kinda coupled (ie we'd never want them to every be out of sync).  Instead of a whole visitor here with DefaultTypeVisitor inheritance, can we just use an if statement with a fallback?

```java
private static boolean canBeCoercedFromAbsent(Type type) {
  return type.accept(TypeVisitor.IS_OPTIONAL) || type.accept(TypeVisitor.IS_LIST) || type.accept(TypeVisitor.IS_SET) || type.accept(TypeVisitor.IS_MAP);
}

private static int typeSortOrder(Type type) {
  return canBeCoercedFromAbsent(type) ? 1 : 0;
}
```

We might then be able to delete the DefaultTypeVisitor entirely (as I think we've got another of these in the `TypeVisitor` class already)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
markelliot: done

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
markelliot: done

iamdanfox (approved):
This PR only affects jersey services but not retrofit ones.  Even so, I think it's worth merging now as it's a strict improvement over the current state of the world (and is worth the one-off compilation break)!

markelliot:
One tricky thing about the sort order stuff is that a parameter that used to be required but is now optional will move in position. So this doesn’t fix all kinds of compile breaks due to Conjure changes, but it should fix a large class of practical changes we observe.

PR #22, opened 2018-06-24
simplify union code generator and generated union code

markelliot:


conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -194,10 +199,6 @@ public void testUnionType() throws Exception {
         assertThat(mapper.readValue(serializedUnionTypeSet, UnionTypeExample.class)).isEqualTo(unionTypeSet);
         assertThat(mapper.readValue(serializedUnionTypeInt, UnionTypeExample.class)).isEqualTo(unionTypeInt);
 
-        assertThat(unionTypeStringExample).isEqualTo(stringExample);
-        assertThat(unionTypeSet).isEqualTo(ImmutableSet.of("item"));
-        assertThat(unionTypeInt).isEqualTo(5);
```
markelliot, line 199: so it turns out that when we added the ability to have a union over the same type this test (and the behavior more generally) became invalid

conjure-java-core/src/integrationInput/java/com/palantir/product/UnionTypeExample.java
```diff
@@ -146,384 +120,360 @@ public String toString() {
...
-
-        @JsonCreator
-        private StringExampleWrapper(@JsonProperty("stringExample") StringExample value) {
-            Objects.requireNonNull(value, "stringExample cannot be null");
-            this.value = value;
-        }
+    @JsonDeserialize(builder = Union.Builder.class)
+    @Generated("com.palantir.conjure.java.types.BeanGenerator")
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private static final class Union {
```
markelliot: as a bit of an alternative for simplifying the generated code further, we could:
 * drop this to just be a mutable class
 * skip using the Builder class and handle everything as nullable instead of optional

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -102,22 +103,46 @@ public static JavaFile generateUnionType(TypeMapper typeMapper, UnionDefinition
...
-    private static MethodSpec generateConstructor(ClassName baseClass) {
-        return MethodSpec.constructorBuilder()
-                .addModifiers(Modifier.PRIVATE)
-                .addAnnotation(AnnotationSpec.builder(JsonCreator.class).build())
-                .addParameter(baseClass, VALUE_FIELD_NAME)
-                // no null check because this constructor is private and is only called by nice factory methods
-                .addStatement("this.$1L = $1L", VALUE_FIELD_NAME)
+    private static TypeSpec generateUnionBean(TypeMapper typeMapper, UnionDefinition typeDef, String typePackage,
+            ClassName unionClass) {
+        return BeanGenerator.generateBeanTypeSpec(typeMapper, ObjectDefinition.builder()
```
ferozco, line 108: could you fix the whitespace here? it makes it a little hard to read

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -149,50 +177,38 @@ private static MethodSpec generateAcceptVisitMethod(ClassName visitorClass, Set<
...
-                beginControlFlow = false;
-            } else {
-                codeBuilder.nextControlFlow("else $L", ifStatement);
+        if (!fieldDefs.isEmpty()) {
+            boolean first = true;
+            CodeBlock.Builder block = CodeBlock.builder();
+            for (FieldDefinition def : fieldDefs) {
+                String getterName = BeanGenerator.asGetterName(def.getFieldName().get());
+                if (first) {
+                    block.beginControlFlow("if ($1N.$2L().isPresent())", UNION_FIELD_NAME, getterName);
```
ferozco, line 186: Probably worth calling out that we now support/fix support for unions with the same type

markelliot:
happy to kill this/it's just a quick experiment, mostly pushed it as a thing to start a conversation around how we want to do unions. it's possible for us to cut more generated code than I've cut here at the cost of more generator code and maybe less immutable generated code.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
markelliot: behavior between this code and existing code is the same, the tests I killed were invalid

ferozco (approved):

markelliot:
@ferozco not actually sure we wanted to merge this/probably still worth a discussion...

iamdanfox:
TBH I'd prefer to only make changes like this when they are motivated by benchmarks or user requests. However, this doesn't change external facing APIs, so it's probably fine to just try it and see.  Slight concerns:

- Heap representation of a conjure union now contains `O(variants)` pointers (previously one pointer to polymorpic Base) because Union is full of optionals, where all but one is empty.
- only one JsonUnwrapped annotation is necessary (fixed: #26)
- keeping Optionals around internally seems unnecessary - why not just nullable fields?

markelliot:
cool yeah, i'm fine with throwing this away, it was a quick experiment to see if we could reduce overall amount of code, threw it up just as a discussion point

PR #23, opened 2018-06-28
Remove need to have ConjureCollections

markelliot:


markelliot:
I was looking at something else and noticed `ConjureCollections`, which we added to deal with iterables, but really shouldn't be required, so I removed it.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -1,41 +0,0 @@
...
-public final class ConjureCollections {
-
-    private ConjureCollections() {
-        // cannot instantiate
-    }
-
-    @SuppressWarnings("unchecked")
-    public static <T> void addAll(Collection<T> addTo, Iterable<? extends T> elementsToAdd) {
-        if (elementsToAdd instanceof Collection) {
-            addTo.addAll((Collection<T>) elementsToAdd);
```
iamdanfox, line 34: I think the point of having this method was that it's more performant to do `addAll` where possible...

markelliot: Depends on the data structure, I guess:

LinkedHashSet:
```java
   public boolean addAll(Collection<? extends E> c) {
        boolean modified = false;
        for (E e : c)
            if (add(e))
                modified = true;
        return modified;
    }
```

ArrayList:
```java
    public boolean addAll(Collection<? extends E> c) {
        Object[] a = c.toArray();
        int numNew = a.length;
        ensureCapacityInternal(size + numNew);  // Increments modCount
        System.arraycopy(a, 0, elementData, size, numNew);
        size += numNew;
        return numNew != 0;
    }
```

markelliot:
@iamdanfox one flag on this one, it probably ought to be documented in ConjureCollections why we do what we do.

iamdanfox:
Have opened a PR here: https://github.com/palantir/conjure-java/pull/24

PR #28, opened 2018-07-03
fix optional param filtering

### LINKED ISSUE #27: Undocumented breaking change between 1.0.0 and 1.1.0
There is an undocumented breaking change between 1.0.0 and 1.1.0 that causes any code that has query parameters appear before any required parameters to fail compilation. The easiest workaround is to reorder the parameters, however that can cause confusion to end users where the type safety may be weaker than in Java (aka TypeScript).

ferozco:
Closes #27.

## Before this PR
There was an issue where the generated services would not compile if they had optional path params and body params

## After this PR

The issue is resolved and we preserve ordering of parameters

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
@@ -58,6 +58,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import javax.lang.model.element.Modifier;
 import org.apache.commons.lang3.StringUtils;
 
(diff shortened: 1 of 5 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -178,10 +177,7 @@ private MethodSpec generateServiceMethod(
         List<MethodSpec> alternateMethods = Lists.newArrayList();
         for (int i = 0; i < queryArgs.size(); i++) {
             alternateMethods.add(createCompatibilityBackfillMethod(
-                    EndpointDefinition.builder()
-                            .from(endpointDef)
-                            .args(Iterables.concat(args, queryArgs.subList(0, i)))
-                    .build(),
+                    endpointDef,
```
markelliot, line 181: If you don’t filter the params, you need to at least pass `i`, so that the method can filter the defaulted params...

conjure-java-core/src/test/resources/test/api/TestService.java.jersey
```diff
@@ -124,29 +125,40 @@ public interface TestService {
...
 
     @Deprecated
     default int testQueryParams(
             AuthHeader authHeader,
             ResourceIdentifier something,
             ResourceIdentifier implicit,
-            Optional<ResourceIdentifier> optionalMiddle) {
+            Optional<ResourceIdentifier> optionalMiddle,
+            Set<String> setEnd,
+            Optional<ResourceIdentifier> optionalEnd,
```
markelliot: this shouldn’t be an argument

conjure-java-core/src/test/resources/test/api/TestService.java.jersey
```diff
@@ -155,8 +167,10 @@ public interface TestService {
             ResourceIdentifier something,
             ResourceIdentifier implicit,
             Optional<ResourceIdentifier> optionalMiddle,
-            Set<String> setEnd) {
+            Set<String> setEnd,
+            Optional<ResourceIdentifier> optionalEnd,
```
markelliot: shouldn’t be an argument

markelliot (changes requested):

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -195,33 +192,41 @@ private MethodSpec createCompatibilityBackfillMethod(
...
+                extraArgs.stream()
+                        .filter(arg -> arg.getArgName().get().equals(param.name))
+                        .findFirst())
+                .collect(Collectors.toList());
 
         MethodSpec.Builder methodBuilder = MethodSpec.methodBuilder(endpointDef.getEndpointName().get())
                 .addModifiers(Modifier.PUBLIC, Modifier.DEFAULT)
                 .addAnnotation(Deprecated.class)
-                .addParameters(params);
+                .addParameters(IntStream.range(0, sortedParams.size())
```
markelliot, line 207: Might be clearer to flatMap here?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -195,33 +192,41 @@ private MethodSpec createCompatibilityBackfillMethod(
             TypeMapper returnTypeMapper,
             TypeMapper methodTypeMapper,
             List<ArgumentDefinition> extraArgs) {
-        List<ParameterSpec> params = createServiceMethodParameters(endpointDef, methodTypeMapper, false);
+        List<ParameterSpec> sortedParams = createServiceMethodParameters(endpointDef, methodTypeMapper, false);
+        List<Optional<ArgumentDefinition>> sortedMaybeExtraArgs = sortedParams.stream().map(param ->
```
markelliot: Wondering if it’d be better to change the method signature/this is a little gross and probably at least deserves a comment explaining what’s going on.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -195,33 +192,41 @@ private MethodSpec createCompatibilityBackfillMethod(
...
-        for (ParameterSpec param : params) {
-            sb.append("$N, ");
-        }
-
-        List<CodeBlock> fillerValues = Lists.newArrayList();
-        for (ArgumentDefinition arg : extraArgs) {
-            sb.append("$L, ");
-            fillerValues.add(arg.getType().accept(TYPE_DEFAULT_VALUE));
-        }
+        List<Object> values = IntStream.range(0, sortedParams.size()).mapToObj(i -> {
```
markelliot, line 216: Can maybe map over the sortedMaybeExtraArgs list?

markelliot (approved):
Looks correct, now. Think it might be possible to make the code clearer/left some nits.

markelliot:
Thanks for cleaning up my error.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
ferozco: Unfortunately we can't since we need the index to check whether the arg exists in `sortedMaybeExtraArgs`

iamdanfox (approved):

PR #29, opened 2018-07-05
Test against golden master testcases.yml

iamdanfox:
[RFC 004](https://github.com/palantir/conjure/pull/35) defines some common test cases to prove that all languages conform to a common wire spec.

This PR loads up the test-cases.json file and a known-good verification-server (published by the [palantir/conjure-verification](https://github.palantir.build/foundry/conjure-verification) repo) and runs the tests.


conjure-java-verifier/src/test/java/com/palantir/conjure/java/compliance/ServerRule.java
```diff
@@ -0,0 +1,89 @@
...
+        Thread thread = new Thread(() -> {
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream(),
+                    StandardCharsets.UTF_8))) {
+                while (true) {
+                    String line = reader.readLine();
+                    if (line == null) {
+                        break;
+                    }
+                    // TODO(dsanduleac): super hacky
+                    // should have logic to derive port from the server's output in a structured way
```
iamdanfox: @dansanduleac could we just pass in a nice $PORT environment variable ?!?!

conjure-java-verifier/src/test/java/com/palantir/conjure/java/compliance/Cases.java
```diff
@@ -0,0 +1,78 @@
...
+                    .registerModule(new Jdk8Module())
+                    .readValue(file, IgnoredTestCases.class).getClient();
+        } catch (IOException e) {
+            throw new RuntimeException(
+                    String.format("Unable to read %s", file),
+                    e);
+        }
+    }
+
+    public static boolean shouldIgnore(EndpointName endpointName, String json) {
```
ferozco, line 64: can we include the test kind here? otherwise we are relying on endpointName uniqueness which I don't think we enforce anywhere

ferozco (approved):
just a nit. LGTM

conjure-java-verifier/src/test/java/com/palantir/conjure/java/compliance/Cases.java
iamdanfox: Nah, I'm actually happy with keeping all the endpoints distinct... it's just slightly fewer functions to have to throw araound

iamdanfox:
I've verified the local dev workflow is still tolerable here:

- `git clean -xdf && ./gradlew compileTestJava --parallel` works
- `./gradlew idea` doesn't re-generate the code unnecessarily (so it remains a <1sec task)
- IntelliJ understands the directories are generated:

screen shot 2018-07-12 at 15 31 52

PR #30, opened 2018-07-05
fix void methods

ferozco:
## Before this PR

There was an issue where generated code would not compile if they had void methods with many optional arguments
## After this PR

The issue is resolved and generated code should compile correctly


iamdanfox (approved):

iamdanfox:
Bug introduced in https://github.com/palantir/conjure-java/pull/21

PR #34, opened 2018-07-09
generate defensive name for inner union class

ferozco:
## Before this PR

If a conjure definition included a type with the name `Union` the generated code would fail to compile due to duplicate classes

## After this PR

We generate an inner class named `Inner_Union` which is guaranteed to not collide since conjure objects names must match `^[A-Z][a-z0-9]+([A-Z][a-z0-9]+)*$`

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/UnionTypeExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/UnionTypeExample.java
@@ -23,10 +23,10 @@ import javax.annotation.Generated;
 /** A type which can either be a StringExample, a set of strings, or an integer. */
 @Generated("com.palantir.conjure.java.types.UnionGenerator")
 public final class UnionTypeExample {
-    @JsonUnwrapped private final Union union;
+    @JsonUnwrapped private final Inner_Union union;
 
     @JsonCreator
-    private UnionTypeExample(Union union) {
+    private UnionTypeExample(Inner_Union union) {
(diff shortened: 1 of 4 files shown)


markelliot:
what can this nested class' name collide with?

ferozco:
Consider the following conjure type:
```
Union:
  union:
    foo: string
	bar: string
```

Both the inner and outer classes would have the same name

iamdanfox (commented):
Can we change one of the tests so that we don't get regressions on this?

conjure-java-core/src/integrationInput/java/com/palantir/product/UnionTypeExample.java
```diff
@@ -120,10 +124,10 @@ public static UnionTypeExample interface_(int value) {
         T visitUnknown(String unknownType);
     }
 
-    @JsonDeserialize(builder = Union.Builder.class)
+    @JsonDeserialize(builder = Inner_Union.Builder.class)
     @Generated("com.palantir.conjure.java.types.BeanGenerator")
     @JsonInclude(JsonInclude.Include.NON_ABSENT)
-    private static final class Union {
+    private static final class Inner_Union {
```
iamdanfox: Man this name just feels ugly to me - I wonder if we should just use the same pattern we've used elsewhere and do a single trailing underscore to ensure that it'll never clash with user code?

iamdanfox (approved):
Pattern we've used elsewhere is just a single trailing underscore for disambiguation, so approved if we use `Union_`!

PR #35, opened 2018-07-11
add retrofit default methods

bavardage:
## Before this PR

Upon adding an extra optional parameter to a service method, we break java backcompat.

## After this PR

Generate default methods for service endpoints, with subsets of parameters.

This is the retrofit equivalent of the work already done here for the jersey generator.


bavardage:
there are a number of other existing bugs I've found here - going to abandon this and try to carve out other fixes :/

iamdanfox:
OK closing out for now - feel free to re-open when you want to resume!

PR #36, opened 2018-07-11
[fix] Builders check primitives are initialized

### LINKED ISSUE #14: Make primitive fields required on builder.build()?
For non-primitive, non-optional, non-collection type fields, we require that the user explicitly set the field on the builder, or else throw an exception. I get that logic, cause not saying optional implies required, means they should be set. However, we do not do the same check for primitive fields. That is to say, the following type definition:

    Foo:
      fields:
        bar: integer

will allow the following builder invocation to build, successfully:

    Foo.builder().build();

this seems inconsistent with the idea that the server controls default parameters. Can we make primitive fields required to be called as well?

One implementation for this can be found in internal gerrit #306681 , can port to here if it's reasonable.

iamdanfox:
## Before this PR

Conjure POJOs containing primitives (e.g. int, boolean) would be coerced from JSON in which the fields were actually missing.  This was not compliant with the wire spec and could lead to nasty bugs, where a field was unintentionally initialized to `0` or `false`. For example, consider this type:

```yaml
UpdateSalaryRequest:
  salary: integer
  shares: integer
```

The following request would result in dfox's salary being set to 0, because the the 'salary' field was missing and is initialized to 0!!
```
curl -XPOST https://some-api/update-salary/dfox --data '{"shares":20}'
```

## After this PR

Trying to deserialize a type with booleans, integers or doubles from `{}` will now fail.

## Concerns

People may have been relying on this behaviour, so there's a chance this might break people's servers... In theory, we could put this behaviour behind a feature flag, and then phase it in slowly?? EDIT not gonna do this as it's not actually a wire-break and people's servers can be trivially fixed by updating their conjure defs:

```diff
 UpdateSalaryRequest:
-  salary: integer
+  salary: optional<integer>
   shares: integer
```

_Future work: unify this `validatePrimitivesHaveBeenInitialized` method with the other `validateFields` method_

fixes #14

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/BooleanExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/BooleanExample.java
@@ -77,7 +77,7 @@ public final class BooleanExample {
         @JsonSetter("coin")
         public Builder coin(boolean coin) {
             this.coin = coin;
-            _coinInitialized = true;
+            this._coinInitialized = true;
             return this;
         }
 
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/BooleanExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/BooleanExample.java
@@ -65,6 +65,8 @@ public final class BooleanExample {
     public static final class Builder {
         private boolean coin;
 
+        private boolean _coinInitialized;
+
         private Builder() {}
 
         public Builder from(BooleanExample other) {
@@ -75,6 +77,7 @@ public final class BooleanExample {
         @JsonSetter("coin")
         public Builder coin(boolean coin) {
             this.coin = coin;
+            _coinInitialized = true;
             return this;
         }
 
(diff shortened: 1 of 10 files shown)


qinfchen (approved):

j-baker (approved):

bkrieger:
It's a little late now, but I think this should have gone into Conjure 2.0.0. While it's technically not a wire break as you say, an unknowing server could accidentally make a wire break if it didn't notice this change.

iamdanfox:
@bkrieger yeah this was definitely a subtle change, but I'm not sure under what criteria we'd class this as a break?

- old JSON can still be deserialized by new servers
- JSON serialized by new servers can still be deserialized by old clients

The only case this wouldn't work is if a new client was trying to talk to an old server, but we don't claim to support that workflow anyway.

bkrieger:
If you upgrade the server to a version with this change, and you don't notice that you have to change `boolean` to `optional<boolean>`, then old JSON **cannot** be deserialized by new servers.

iamdanfox:
@bkrieger if the JSON came from a remoting client then you'll never actually hit that problem because ObjectMappers.newObjectMapper() will always serialize the value as a boolean:

```java
class Foo {
 boolean hello;
}

System.out.println(ObjectMappers.newClientObjectMapper().writeValueAsString(new Foo()));
// {"hello": false}
```

bkrieger:
Sure, but we have non-Java consumers. In addition, a client using an old
version on the Java API might not include the Boolean if it was previously
not included in the API.  We used the fact the booleans default to false as
a way to add fields to the a request object without a breaking change.

iamdanfox:
So the requirement now is to add `optional<boolean>` if you want to add a field to a request object in a backwards compatible way (still tolerating requests from older clients).

bkrieger:
Yep, and that's fine going forward. But if I used boolean with an old
conjure version, and then upgraded conjure without carefully checking the
release notes (because this was marked as a fix, not a break), I've lost
backcompat.

PR #37, opened 2018-07-12
[improvement] FeatureFlag to make Response the return type for binary Jersey Responses

ellisjoe:
## Before this PR
Currently binary response types are rendered as `StreamingOutput`. There's a bug in Jersey which causes a failure part way through streaming back that response to appear as a successful response to the client: https://github.com/eclipse-ee4j/jersey/issues/3850

## After this PR
By rendering `Response` as the return type we are still able to use `StreamingOutput` as the body of the response, but we are also able to include the `Content-Length` header when we know the size of the response. Including the content length allows clients to determine there was an error, even when the stream appeared to have been closed successfully.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/FeatureFlags.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/FeatureFlags.java
@@ -16,7 +16,9 @@
 
 package com.palantir.conjure.java;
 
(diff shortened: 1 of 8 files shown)


iamdanfox:
I think this is an acceptable short-term solution - we'll probably want to follow up with a more first-class type for java servers so that people can actually specify these content lengths without this hack, but happy to unblock things for now!

iamdanfox (approved):

PR #38, opened 2018-07-16
[improvement] Add feature flag to require NotNull AuthHeaders and BodyParams.

f-doot:
## Before this PR
In generated Jersey Services, AuthHeaders and BodyParams could be null, which causes NullPointerExceptions or requires defensive null checking.

## After this PR
If `jersey-bean-validation` is on the server class path, deserializing of `@NotNull` annotated method parameters generated by conjure should now fail with a 400/422, rather than inadvertantly causing 500s.

## Notes
- I've added a [new library](https://mvnrepository.com/artifact/javax.validation/validation-api/1.1.0.Final) to the dependencies for the annotation. According to mvnrepository this library is Apache 2 licensed.
- I wasn't sure if this warranted adding an ETE test, happy to add one if so.


conjure-java-core/src/test/resources/example-service-with-not-null.yml
```diff
@@ -0,0 +1,189 @@
+types:
```
ferozco: could we avoid adding a whole new yaml and instead re-use one of the existing ones?

iamdanfox (commented):
I'm keen to make this available for everyone, as a guiding principle of Conjure is that you should never have to worry about nulls anymore! Adding it as an opt-in feature flag initially sounds good though.

I think an ETE test to prove this works would actually be great, because otherwise it's just a magic annotation that we're adding (hopefully could be added to `JerseyServiceEteTest`??)

ferozco (commented):
Echo Dan's thought that an ETE test and controlled roll out is great. Also would be nice if you could add some documentation explaining the flag and what build configuration is required to get it to work as intended

conjure-java-core/src/test/resources/example-service-with-not-null.yml
f-doot: Wasn't sure what the policy of modifying those was, will do.

f-doot:
@iamdanfox @ferozco 
- I added the ETE tests. Feign doesn't allow you to post null request bodies and serializes null authheaders as the string `"null"` rather than a missing value so I had to get a little creative with implementing the relevant tests. Happy to adjust if there's a better way of doing this.
- I folded the new endpoint to test the service generator into the original service file as requested.
- About adding documentation on how to use this: Where do you think the correct place to add this is? Happy to add it to the Java enum but it feels a bit weird having specific instructions in the javadocs there.

uschi2000:
This is sort of going the wrong direction: I would expect every param to be not null **unless** explicitly tagged as `@Nullable`

iamdanfox:
@uschi2000 the intention here is to roll this behaviour out to everyone, we're just using a feature flag to do a 'soft launch'.

uschi2000:
sure, but code-style-wise I always expect a variable that's not annotated with `@Nullable` to be not null. that's we this is going the "wrong" direction.

iamdanfox:
@uschi2000 yeah I agree that would be nice, but it seems Jersey doesn't give you that behaviour. Instead, it seems to respect the `javax.validation.constraints.NotNull` annotation.

Is there an alternative approach you'd prefer to achieve the outcome of banning nulls?

uschi2000:
oh, this is for runtime enforcement? then it's probably fine. the not-null-by-default is mostly useful for static analysis purposes.

uschi2000:
is there a way to tell jersey "non of the parameters must ever be null?"

f-doot:
I had a cursory look around and couldn't find anything that immediately filled that, most of the stuff I found refers back to using `@NotNull` with the bean validation dependency on the classpath.

Should we keep moving forwards with this, given that we're talking about runtime enforcement here?

ferozco:
@fscreene I think its fine to continue with this. As for the documentation, I think it might make sense to have something in the readme. Perhaps we could add something under usage

f-doot:
@ferozco Added a small docs section on feature flags. I'm not sure how you do the manpage style stuff for something where the actual value following the `--` is what is variable. I think at the moment this can be a little bit confusing, happy to add an example if you'd like. I linked to the actual class, rather than writing a blurb about each feature, it felt a bit weird to have actual docs for individual featureflags in the readme.

f-doot:
@ferozco @iamdanfox Ping on the above

conjure-java-core/src/main/java/com/palantir/conjure/java/FeatureFlags.java
```diff
@@ -32,4 +33,10 @@
      * type.
      */
     JerseyBinaryAsResponse,
+
+    /**
+     * Instructs the {@link JerseyServiceGenerator} to add {@link NotNull}
+     * annotations to all auth parameters, as well as all non-optional body params on service endpoints.
+     */
+    RequireNotNullAuthAndBodyParams,
```
iamdanfox, line 41: I've opened https://github.com/palantir/conjure-java/issues/51 so we don't forget to promote this to an 'enabled-by-default' feature.

iamdanfox (approved):

conjure-java-core/src/main/java/com/palantir/conjure/java/FeatureFlags.java
f-doot: Sounds good!

PR #40, opened 2018-07-16
[WIP][break] remove support for Nan/Inf

qinfchen:
## Before this PR
Generated Double types allow `NaN` and `Infinity` as values

## After this PR
`NaN` and `Infinity` are now banned in generated Double types to be compliant with the [spec(to be updated)](https://github.com/palantir/conjure/blob/develop/docs/specification.md)

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
@@ -16,6 +16,7 @@
 
 package com.palantir.conjure.java.types;
 
+import static com.squareup.javapoet.TypeName.DOUBLE;
 import static java.util.stream.Collectors.toList;
 
 import com.fasterxml.jackson.annotation.JsonAnySetter;
@@ -53,6 +54,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import javax.lang.model.element.Modifier;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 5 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -60,15 +62,10 @@ public void testOptionalSerialization() throws Exception {
...
-    public void double_nan_fields_should_be_serialized_as_a_string() throws Exception {
-        assertThat(mapper.writeValueAsString(DoubleExample.of(Double.NaN)))
-                .isEqualTo("{\"doubleValue\":\"NaN\"}");
-    }
-
-    @Test
-    public void double_nan_fields_should_deserialize_from_a_string() throws Exception {
-        DoubleExample deserialized = mapper.readValue("{\"doubleValue\":\"NaN\"}", DoubleExample.class);
-        assertThat(deserialized.getDoubleValue()).isNaN();
+    public void deserializing_double_nan_fields_should_throw_an_error() {
```
ferozco, line 65: should also test that serialising a nan/inf fails

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -319,6 +320,19 @@ private static CodeBlock typeAwareAssignment(EnrichedField enriched, Type type,
...
                     .addStatement("this.$1N.rewind()", spec.name)
                     .build();
+        } else if (Objects.equals(spec.type, TypeName.DOUBLE)) {
+            // ban `NaN` and `Infinity` for double
+            CodeBlock.Builder codeBlockBuilder = CodeBlock.builder();
+            codeBlockBuilder
+                    .beginControlFlow(
+                            "if ($1T.isNaN($2N) || $1T.isInfinite($2N))",
+                            Double.class, spec.name)
+                    .addStatement("throw new $T(\"NaN or Infinity for double type, $N, is not allowed.\")",
```
ferozco, line 330: The error sounds a little awkward. maybe `Illegal value, doubles must have a concrete value`

ferozco (commented):
What about unions/optionals

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
qinfchen: serializing from DoubleExample? DoubleExample wouldn't be instantiated with nan/inf values

qinfchen:
Close this in favor of https://github.com/palantir/http-remoting/pull/761. There are too many places where we have to explicitly check for strict values which is error prone.  We could introduce a conjure new type for double, but that is pretty intrusive.

PR #44, opened 2018-07-18
Failure with union with single value.

jkozlowski:
I don't know the fix, but I thought I'd provide the failure case:

The code in this PR seems to fail to generate correct objects, @dansanduleac  could someone have a look?

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/BlaBla1.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/BlaBla1.java
@@ -1,0 +1,215 @@
+package com.palantir.product;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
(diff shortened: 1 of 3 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/BlaBla1.java
```diff
@@ -0,0 +1,215 @@
...
+                    .append("}")
+                    .toString();
+        }
+
+        public static Union_ of(
+                String type, BlaBla blaBla, Map<String, Object> __unknownProperties) {
+            return builder()
+                    .type(type)
+                    .blaBla(Optional.of(blaBla))
+                    .__unknownProperties(__unknownProperties)
```
jkozlowski, line 141: This is where it fails

dansanduleac: Seems like we're just not generating a correct builder method for unknown properties

jkozlowski:
Looks like the fix was merged?

PR #45, opened 2018-07-18
revert union changes

ferozco:
Revert #22. Since it was originally just an experiment and it caused #44, I hope its ok. I am keen to open up the discussion on how implement unions, especially if we have some numbers around the performance of one approach vs. another


markelliot (approved):
👍

dansanduleac (approved):

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -340,7 +301,9 @@ private static boolean isOptional(FieldSpec spec) {
 
         @Value.Derived
         default String getterName() {
-            return asGetterName(fieldName().get());
```
dansanduleac: With this refactor, asGetterName is now dead code

conjure-java-core/src/test/resources/example-types.yml
```diff
@@ -160,3 +160,6 @@ types:
         union:
```
dansanduleac, line 160: do we need this `Union` type at all, given that there is also a `UnionTypeExample` type that does a similar thing but with more variants?

dansanduleac: Note: it's not used anywhere

dansanduleac (changes requested):

dansanduleac (approved):

conjure-java-core/src/test/resources/example-types.yml
dansanduleac: Left it there so we can have more cases covered by tests

PR #50, opened 2018-07-25
[break] alias of binary return types is the same as binary return types

qinfchen:
## Before this PR
alias of binary return type was treated differently as  binary return type

## After this PR
alias of binary return type is now treated the same as binary return type

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/Retrofit2ReturnTypeClassNameVisitor.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/Retrofit2ReturnTypeClassNameVisitor.java
@@ -16,17 +16,16 @@
 
 package com.palantir.conjure.java.types;
 
-import com.palantir.conjure.spec.AliasDefinition;
+import static com.palantir.conjure.java.util.BinaryReturnTypeResolver.resolveReturnReferenceType;
+
 import com.palantir.conjure.spec.ExternalReference;
 import com.palantir.conjure.spec.ListType;
 import com.palantir.conjure.spec.MapType;
 import com.palantir.conjure.spec.OptionalType;
 import com.palantir.conjure.spec.PrimitiveType;
 import com.palantir.conjure.spec.SetType;
-import com.palantir.conjure.spec.Type;
 import com.palantir.conjure.spec.TypeDefinition;
 import com.palantir.conjure.visitor.TypeDefinitionVisitor;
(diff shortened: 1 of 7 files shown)


iamdanfox:
Cool, looks like this achieves the short-term goal of enabling client and server-side streaming when a conjure endpoint returns an alias of binary! Kinda sad that we lose the alias name though - for a longer term solution, could we set up some kind of [MessageBodyWriter](https://docs.oracle.com/javaee/7/api/javax/ws/rs/ext/MessageBodyWriter.html) somehow?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/Retrofit2ReturnTypeClassNameVisitor.java
```diff
@@ -75,4 +85,28 @@ public TypeName visitExternal(ExternalReference type) {
     public TypeName visitSet(SetType type) {
         return delegate.visitSet(type);
     }
+
+    private TypeName resolveReferenceType(com.palantir.conjure.spec.TypeName type) {
```
iamdanfox: Can we dedupe this with the jersey version?

qinfchen:
@iamdanfox, what does the alias name of return binary give us? Wouldn't that just introduce some ambiguity around what the return type of the generated interfaces should behave, both the client and the server side? In the end, we want alias of binary return type to be always considered as `StreamingOutput`, so is it desirable to expose alias name that wraps around `ByteBuffer` to the end users and then have a `MessageBodyWriter` to handle the response as `StreamingOutput`?

conjure-java-core/src/main/java/com/palantir/conjure/java/util/BinaryReturnTypeResolver.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param types - a map of name to conjure definition
+     * @param type - the given type to be resolved
+     * @param binaryTypeSupplier - supplier of the return binary type if the leaf node is of a binary type
+     * @return resolved return reference type
+     */
+    public static TypeName resolveReturnReferenceType(
+            Map<com.palantir.conjure.spec.TypeName, TypeDefinition> types,
+            com.palantir.conjure.spec.TypeName type,
+            Supplier<TypeName> binaryTypeSupplier) {
```
iamdanfox: Could this just be a plain TypeName instead of a supplier?

iamdanfox (approved):

PR #54, opened 2018-08-03
[improvement] Retrofit2 clients should also set hr-path-template

j-baker:
Feign does this, and Retrofit does not. Let's make our retrofit
interfaces contain this header, too!

I've traced this through with a debugger and verified that it does the
right thing.

## Before this PR

For tracing purposes (and soon, also concurrency limiting) the Feign
contract sets the hr-path-template header to the path template for
tracing purposes. An OkHttp interceptor removes this header, and
uses it to start a span.

## After this PR

Retrofit generated interfaces also contain the hr-path-template header.

Relevant to https://github.com/palantir/http-remoting/pull/786.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceRetrofit.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceRetrofit.java
@@ -4,9 +4,11 @@ import java.lang.Boolean;
 import javax.annotation.Generated;
 import retrofit2.Call;
 import retrofit2.http.GET;
+import retrofit2.http.Headers;
 
 @Generated("com.palantir.conjure.java.services.Retrofit2ServiceGenerator")
 public interface EmptyPathServiceRetrofit {
     @GET("./")
+    @Headers("hr-path-template: /")
     Call<Boolean> emptyPath();
 }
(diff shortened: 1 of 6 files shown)


iamdanfox:
@j-baker I didn't realise feign clients already set a `hr-path-template`, so looked up where this is implemented:

https://github.com/palantir/http-remoting/blob/3.41.1/tracing-okhttp3/src/main/java/com/palantir/remoting3/tracing/okhttp3/OkhttpTraceInterceptor.java#L33

```java
    /** The HTTP header used to communicate API endpoint names internally. Not considered public API. */
    public static final String PATH_TEMPLATE_HEADER = "hr-path-template";
```

This has also been copied to the new [palantir/tracing-java repo](https://github.com/palantir/tracing-java/blob/develop/tracing-okhttp3/src/main/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptor.java#L33).

If we're adding this to the generated code then I think we need to commit to this particular header as API...

j-baker:
depends what you mean by API, Dan. Anyone using http-remoting won't be able to get any useful information out of it - it doesn't go over the wire.

j-baker:
also, I'd argue we should commit to that header as API

iamdanfox (approved):
Looks good, this will enable the more granular throttling in https://github.com/palantir/conjure-java-runtime/pull/786.

PR #56, opened 2018-08-20
[break] conjure-lib shouldn't depend on errors

dansanduleac:


conjure-lib/build.gradle
```diff
@@ -20,7 +20,6 @@ dependencies {
     compile 'com.fasterxml.jackson.core:jackson-databind'
     compile 'com.palantir.ri:resource-identifier'
     compile 'com.palantir.tokens:auth-tokens'
-    compile 'com.palantir.remoting-api:errors'
```
dansanduleac, line 23: this dependency here had the "lucky" side effect that it magically brought in `javax.ws.rs-api` too

ferozco (approved):
Lets cut a release!

PR #59, opened 2018-08-30
[WIP] [fix] Use conjure executable in tests

dansanduleac:
Remaining work:
other tests need to use the generated IR that will be added to test resources


dansanduleac:
Decided against this for now.

PR #60, opened 2018-08-31
[fix] Use correct optional type for primitives

tom-s-powell:
Compatibility backfill methods do not currently work when the full implementation uses `optional<integer>` or `optional<double>` since these generate `OptionalInt` and `OptionalDouble` of which it is not valid to use `Optional.empty()`.


ferozco (approved):
This looks great!

Issue #61, opened 2018-08-31
Union generator generates a broken equals method


ellisjoe:
Example here: https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java#L55-L57

Equals is not symmetric with respect to the union wrapper and the thing it wraps. Eg:
```
FooWrapper fooWrapper = FooWrapper.of("foo");
Union union = Union.of(fooWrapper);

union.equals(fooWrapper) // isTrue
fooWrapper.equals(union) // isFalse
```

PR #62, opened 2018-08-31
[fix] Make the generated Union equals method symmetric

ellisjoe:
## Before this PR

https://github.com/palantir/conjure-java/issues/61
The generated union equals method is not symmetric with the wrapper equals methods as described in the ticket above.

## After this PR

The new generated method is only equal to other Union objects and not the wrapped types.

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
@@ -199,10 +199,6 @@ public final class WireFormatTests {
         assertThat(mapper.readValue(serializedUnionTypeSet, UnionTypeExample.class)).isEqualTo(unionTypeSet);
         assertThat(mapper.readValue(serializedUnionTypeInt, UnionTypeExample.class)).isEqualTo(unionTypeInt);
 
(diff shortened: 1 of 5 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -199,10 +199,6 @@ public void testUnionType() throws Exception {
         assertThat(mapper.readValue(serializedUnionTypeSet, UnionTypeExample.class)).isEqualTo(unionTypeSet);
         assertThat(mapper.readValue(serializedUnionTypeInt, UnionTypeExample.class)).isEqualTo(unionTypeInt);
 
-        assertThat(unionTypeStringExample).isEqualTo(stringExample);
-        assertThat(unionTypeSet).isEqualTo(ImmutableSet.of("item"));
-        assertThat(unionTypeInt).isEqualTo(5);
```
iamdanfox, line 204: Should we actually keep these tests by make them say `isNotEqualTo`?

iamdanfox (approved):
LGTM. @markelliot just to double-check we haven't missed an optimization here!

markelliot (approved):

PR #77, opened 2018-10-03
[fix] nested aliased query params delegate to aliased type's valueOf static method

### LINKED ISSUE #68: Conjure-generated service does not start if endpoint contains certain alias definitions
Example input definition:

```
types:
  definitions:
    default-package: com.palantir.product
    objects:
      AliasOne:
        alias: string
      AliasTwo:
        alias: optional<AliasOne>
services:
  EteService:
    name: Ete Service
    package: com.palantir.product
    default-auth: header
    base-path: /base

    endpoints:
      optionalAliasOne:
        http: GET /optionalAliasOne
        args:
          queryParamName:
            param-type: query
            type: optional<AliasOne>
        returns: optional<AliasOne>

      aliasTwo:
        http: GET /aliasTwo
        args:
          queryParamName:
            param-type: query
            type: AliasTwo
        returns: AliasTwo
```

Note that, form a serialization/semantic standpoint, the `optionalAliasOne` and `aliasTwo` are equivalent.

However, attempting to start a Conjure-generated server with these definitions fails with the following exception:

```
Caused by: org.glassfish.jersey.server.model.ModelValidationException: Validation of the application resource model has failed during application initialization.
[[FATAL] No injection source found for a parameter of type public com.palantir.product.AliasTwo com.palantir.conjure.java.EteResource.aliasTwo(com.palantir.tokens.auth.AuthHeader,com.palantir.product.AliasTwo) at index 1.; source='ResourceMethod{httpMethod=GET, consumedTypes=[application/json], producedTypes=[application/json], suspended=false, suspendTimeout=0, suspendTimeoutUnit=MILLISECONDS, invocable=Invocable{handler=ClassBasedMethodHandler{handlerClass=class com.palantir.conjure.java.EteResource, handlerConstructors=[org.glassfish.jersey.server.model.HandlerConstructor@6ac4c3f7]}, definitionMethod=public abstract com.palantir.product.AliasTwo com.palantir.product.EteService.aliasTwo(com.palantir.tokens.auth.AuthHeader,com.palantir.product.AliasTwo), parameters=[Parameter [type=class com.palantir.tokens.auth.AuthHeader, source=Authorization, defaultValue=null], Parameter [type=class com.palantir.product.AliasTwo, source=queryParamName, defaultValue=null]], responseType=class com.palantir.product.AliasTwo}, nameBindings=[]}']
```

If the `aliasTwo` endpoint is commented out, the server starts up as expected.

The `aliasTwo` types and definitions should be legal per the Conjure spec, so the Java generator creating a server that does not start based on these valid definitions is a bug.

qinfchen:
fixes https://github.com/palantir/conjure-java/issues/68
## Before this PR

Conjure generated endpoints with nested aliased query params do not meet any of the following conditions for `QueryParam` annotation type:
```
1. Be a primitive type
2. Have a constructor that accepts a single String argument
3. Have a static method named valueOf or fromString that accepts a single String argument (see, for example, Integer.valueOf(String))
4. Have a registered implementation of ParamConverterProvider JAX-RS extension SPI that returns a ParamConverter instance capable of a "from string" conversion for the type.
5. Be List<T>, Set<T> or SortedSet<T>, where T satisfies 2, 3 or 4 above. The resulting collection is read-only.
```
As a result, servers with those endpoints would fail with those exceptions
```
Caused by: org.glassfish.jersey.server.model.ModelValidationException: Validation of the application resource model has failed during application initialization.
[[FATAL] No injection source found for a parameter of type public com.palantir.product.AliasTwo com.palantir.conjure.java.EteResource.aliasTwo(com.palantir.tokens.auth.AuthHeader,com.palantir.product.AliasTwo) at index 1.; source='ResourceMethod{httpMethod=GET, consumedTypes=[application/json], producedTypes=[application/json], suspended=false, suspendTimeout=0, suspendTimeoutUnit=MILLISECONDS, invocable=Invocable{handler=ClassBasedMethodHandler{handlerClass=class com.palantir.conjure.java.EteResource, handlerConstructors=[org.glassfish.jersey.server.model.HandlerConstructor@6ac4c3f7]}, definitionMethod=public abstract com.palantir.product.AliasTwo com.palantir.product.EteService.aliasTwo(com.palantir.tokens.auth.AuthHeader,com.palantir.product.AliasTwo), parameters=[Parameter [type=class com.palantir.tokens.auth.AuthHeader, source=Authorization, defaultValue=null], Parameter [type=class com.palantir.product.AliasTwo, source=queryParamName, defaultValue=null]], responseType=class com.palantir.product.AliasTwo}, nameBindings=[]}']
```

## After this PR

Nested aliased query params are now generated with a static method named valueOf that accepts string.

TODO:
- [ ] alias of `binary` and `any` type on query param never worked. Here is a [PR](https://github.com/palantir/conjure/pull/114) created in conjure-core to invalidate them at compile time instead of generation time

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
@@ -133,7 +133,7 @@ public final class AliasGenerator {
                 && !conjureType.accept(TypeVisitor.IS_BINARY)) {
             return Optional.of(valueOfFactoryMethodForPrimitive(
                     conjureType.accept(TypeVisitor.PRIMITIVE), thisClass, aliasTypeName));
-        } else if (conjureType.accept(TypeVisitor.IS_REFERENCE)) {
+        } else if (conjureType.accept(TypeVisitor.IS_INTERNAL_REFERENCE)) {
             // delegate to aliased type's valueOf factory method
             Optional<AliasDefinition> aliasTypeDef = typeMapper.getType(conjureType.accept(TypeVisitor.REFERENCE))
                     .filter(type -> type.accept(TypeDefinitionVisitor.IS_ALIAS))
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
@@ -22,6 +22,7 @@ import com.palantir.conjure.java.ConjureAnnotations;
 import com.palantir.conjure.spec.AliasDefinition;
 import com.palantir.conjure.spec.PrimitiveType;
 import com.palantir.conjure.spec.Type;
+import com.palantir.conjure.visitor.TypeDefinitionVisitor;
 import com.palantir.conjure.visitor.TypeVisitor;
 import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.CodeBlock;
@@ -41,6 +42,7 @@ public final class AliasGenerator {
     public static JavaFile generateAliasType(
             TypeMapper typeMapper,
             AliasDefinition typeDef) {
+
         TypeName aliasTypeName = typeMapper.getClassName(typeDef.getAlias());
 
         String typePackage = typeDef.getTypeName().getPackage();
@@ -78,7 +80,7 @@ public final class AliasGenerator {
                         .build());
 
         Optional<CodeBlock> maybeValueOfFactoryMethod = valueOfFactoryMethod(
-                typeDef.getAlias(), thisClass, aliasTypeName);
+                typeDef.getAlias(), thisClass, aliasTypeName, typeMapper);
         if (maybeValueOfFactoryMethod.isPresent()) {
             spec.addMethod(MethodSpec.methodBuilder("valueOf")
                     .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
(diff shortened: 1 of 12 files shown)


uschi2000:
@nmiyake

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -122,15 +124,30 @@ public static JavaFile generateAliasType(
...
+
         // doesn't support valueOf factories for ANY and BINARY types
         if (conjureType.accept(TypeVisitor.IS_PRIMITIVE)
                 && !conjureType.accept(TypeVisitor.IS_ANY)
                 && !conjureType.accept(TypeVisitor.IS_BINARY)) {
             return Optional.of(valueOfFactoryMethodForPrimitive(
                     conjureType.accept(TypeVisitor.PRIMITIVE), thisClass, aliasTypeName));
+        } else if (conjureType.accept(TypeVisitor.IS_REFERENCE)) {
+            // delegate to aliased type's valueOf factory method
+            Optional<AliasDefinition> aliasTypeDef = typeMapper.getTypes().stream()
```
ferozco: could we pull this out and construct map of typeNames to definitions?

qinfchen: Instead of return type definitions, we could just move the map to TypeMapper and have a getter for it if people are okay with that.

dansanduleac (approved):

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -122,15 +124,28 @@ public static JavaFile generateAliasType(
...
                 && !conjureType.accept(TypeVisitor.IS_BINARY)) {
             return Optional.of(valueOfFactoryMethodForPrimitive(
                     conjureType.accept(TypeVisitor.PRIMITIVE), thisClass, aliasTypeName));
+        } else if (conjureType.accept(TypeVisitor.IS_REFERENCE)) {
+            // delegate to aliased type's valueOf factory method
+            Optional<AliasDefinition> aliasTypeDef = typeMapper.getType(conjureType.accept(TypeVisitor.REFERENCE))
+                    .filter(type -> type.accept(TypeDefinitionVisitor.IS_ALIAS))
+                    .map(type -> type.accept(TypeDefinitionVisitor.ALIAS));
+
+            return aliasTypeDef.map(type -> {
```
iamdanfox, line 142: I think there's a bug here because the following code-block assume that `type` is a conjure-defined type (because we got it from the `IS_REFERENCE` visitor).

HOWEVER - it seems the IS_REFERENCE visitor returns true for both external references and also internal references, so if it was actually pointing to an external class `Foo` then the `Foo.valueOf` method might not exist, leading to use generating broken code!

I think splitting the IS_REFERENCE visitor in two would be a reasonable way to solve this (it's bitten me in another project recently!)

iamdanfox (changes requested):
Let's fix the external reference bug pls!  When this merges I think we should also follow-up to add this edge case to the central test-cases.yml so that all generators are forced to handle this https://github.com/palantir/conjure-verification/blob/develop/test-cases.yml

iamdanfox (approved):

PR #85, opened 2018-10-12
[fix] Deserialize NaN and infinities into double aliases

mcopes73:
## Before this PR
Double aliases failed to deserialize when the double was NaN or infinity because no proper constructor was generated to deserialize a double from a string.

## After this PR
Double aliases now generate a constructor that takes a string and checks for the cases "NaN", "Infinity" and "-Infinity".

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
@@ -26,7 +26,7 @@ public final class DoubleAliasExample {
     public boolean equals(Object other) {
         return this == other
                 || (other instanceof DoubleAliasExample
-                        && this.value == ((DoubleAliasExample) other).value);
+                && this.value == ((DoubleAliasExample) other).value);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
```diff
@@ -47,4 +47,19 @@ public static DoubleAliasExample of(double value) {
...
     }
+
+    @JsonCreator
+    public static DoubleAliasExample of(String value) {
+        switch (value) {
+            case "NaN":
+                return DoubleAliasExample.of(Double.NaN);
+            case "Infinity":
+                return DoubleAliasExample.of(Double.POSITIVE_INFINITY);
+            case "-Infinity":
```
iamdanfox, line 58: Confirmed these match the wire spec at https://github.com/palantir/conjure/blob/develop/docs/spec/wire.md#json-format

iamdanfox: Ah I've just realised all of this logic is already encapsulated in Java's existing `Double.parseDouble(String)` method... since we already have a `valueOf` method that does this, I think we could just add the `@JsonCreator` annotation to that method, and then we won't need the whole new method?

mcopes73: This will also deserialize values like "123", ignore leading and trailing whitespace, and will even accept floats and round them, e.g. "1.60" deserializes to 2. It basically delegates to [Double#valueOf](https://docs.oracle.com/javase/6/docs/api/java/lang/Double.html#valueOf(java.lang.String)) - do we also want this behavior?

The Exception thrown is also different, we would have to catch it and rethrow IllegalArgument

mcopes73: I guess this behavior is fine. If so, I am in favor of just adding the `@JsonCreator` annotation to `valueOf`.

iamdanfox: No I think your point makes sense - I don't want to start tolerating quoted numbers.

iamdanfox:
Corresponding addition to test-cases.yml: https://github.com/palantir/conjure-verification/pull/46

EDIT merged: https://github.com/palantir/conjure-java/pull/86

dansanduleac (approved):

PR #86, opened 2018-10-12
Update conjure-verification server 0.6.0 -> 0.11.0

iamdanfox:
Many new test cases have been added to https://github.com/palantir/conjure-verification/blob/develop/test-cases.yml.  We should always be checking against latest!


conjure-java-verifier/src/test/resources/ignored-test-cases.yml
```diff
@@ -22,6 +22,9 @@ client:
       - "{\"value\":\"1523040070\"}" # jackson coerces things to other types
       - "{\"value\":1523040070}" # jackson coerces things to other types
 
+    receiveRawOptionalExample:
+      - "null" # https://github.com/palantir/conjure-java-runtime/issues/813
```
iamdanfox, line 26: Mentioning https://github.com/palantir/conjure-java-runtime/issues/813

dansanduleac (approved):

PR #88, opened 2018-10-16
[improvement] Annotate getters with required = true/false in addition to value = name

tjb9dc:
## Before this PR
We were trying to use [this library](https://github.com/mbknor/mbknor-jackson-jsonSchema) to generate a jsonSchema from our Conjure spec and ran into an issue where it couldn't handle Conjure optionals, as it expects required fields to be marked in the `jsonProperty` annotation. To be useful, we need the generator to properly designate things as required for validation of our Python-generated payloads. We are using Python because the Delta Sierras use Python to interact run their models.

## After this PR
This got me thinking, why doesn't Conjure designate required fields in the `jsonProperty` annotation ( it defaults to `required=false`)? This PR decorates the getters of optional fields as `required=false` and the other getters as `required=true`. It accomplishes this by checking for an `Optional` wrapper as well as `OptionalInt` and `OptionalDouble` types. The files located at `src/integrationInput/java` that tests are run against were updated with the newly generated classes and look good so far.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
@@ -185,6 +185,7 @@ public final class BeanGenerator {
                 .addModifiers(Modifier.PUBLIC)
                 .addAnnotation(AnnotationSpec.builder(JsonProperty.class)
                         .addMember("value", "$S", field.fieldName().get())
+                        .addMember("required", "$L", !isOptional(field.poetSpec()))
                         .build())
                 .returns(field.poetSpec().type);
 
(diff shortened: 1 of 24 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -291,6 +292,15 @@ private static boolean isOptional(FieldSpec spec) {
         return ((ParameterizedTypeName) spec.type).rawType.simpleName().equals("Optional");
     }
 
+    private static boolean isRequiredField(EnrichedField field) {
+        if (!(field.poetSpec().type instanceof ParameterizedTypeName)) {
+            // spec isn't a wrapper class, check for OptionalInt or OptionalDouble
+            return !field.poetSpec().type.toString().equals("java.util.OptionalInt")
```
TheBeruriahIncident, line 298: Can you compare types by comparing the `Class`s? E.g. rather than Strings?

TheBeruriahIncident: Something like `field.poetSpec().type.equals(OptionalInt.class)`

uschi2000:
Suppose we were to move off Jackson. How hard would it be to implement a conjure-to-jsonschema generator?

tjb9dc:
I'm going to take a look at https://github.com/palantir/conjure-python to see if it can do enough for us

tjb9dc:
@uschi2000 That's a good question...was hoping to get this for free but worth considering that.

PR #90, opened 2018-10-19
Fix #65: Publish conjure-java-core

### LINKED ISSUE #65: Publish jar file for core Java generator code
Publishing a jar for core Java generator code will allow other service targets (e.g. Undertow, Netty) to make use of the same infrastructure and type definitions while replacing the specific service outputs.

We might also consider splitting the type generators from the service generators -- but either way need things like the `TypeMapper` infrastructure to be available for services so they can refer to types.

carterkozak:
The `conjure-java-core` jar is published to allow us to prototype new generators without checking in library code.

### DIFF
--- a/conjure-java-core/build.gradle
+++ b/conjure-java-core/build.gradle
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+apply from: "$rootDir/gradle/publish-jar.gradle"
+
 sourceSets {
     integrationInput
     test {


carterkozak:
fysa @markelliot @uschi2000 @iamdanfox

uschi2000 (approved):

uschi2000:
also need this in other places. what's the deal, @iamdanfox ?

iamdanfox (changes requested):
I'm pretty torn about publishing this:

- We already have a clearly defined API-surface ([RFC002](https://github.com/palantir/conjure/blob/develop/docs/rfc/002-contract-for-conjure-generators.md)), but if we publish this jar then people will inevitably start relying on it. We know of your two use-cases already, but I anticipate there would be external users too.

- As maintainers of Conjure, we want to be able to refactor/rename the internals without impacting users.  If we publish the jar then it seems that we'll inevitably get complaints from users who have come to rely on some classes we considered internal.

Long-term, I think we really want one conjure-java executable that generates all the objects and services for the Java language. Crucially, I don't want there to be lots of slightly inter-dependent conjure-java generators which all make assumptions about the methods on a generated object or exactly how the builders work.

## Proposal

Assuming the goal is to get the code in your respective repos into this codebase eventually, how feasible would it be to push your work as a long-lived feature branch on a fork of this repo? This would allow you to set up your own publishing, but also frequently merge from develop to avoid diverging too hard?

carterkozak:
long-running feature branches tend to have similar support issues in my experience. Incubating sub-modules would be strictly superior.

Some features were removed from conjure-java prior to open sourcing, which you're suggesting that I re-implement in the OSS repository, is that correct?

uschi2000:
`we want to be able to refactor/rename the internals without impacting users`
I don't think you need to hold this as a goal.

conjure-dialogue was removed from the "main" conjure repo, so I think you'll have to pick a side: either allow external uses of the conjure library, or allow people to put experimental stuff in the core repo.

iamdanfox:
@uschi2000 I think it's pretty important that maintainers are free to refactor classes internally to get closer to better abstractions. Many of these classes were written a long time ago and were not intended to be used as Java API. If it's eventually a goal to commit to a larger API surface, then we need to be able to iterate towards that clean API rather than just commit to everything we have today.

That being said, I don't really want experimental stuff merged into the proper repo - anything that is merged to the repo needs to uphold the same back-compat guarantees that conjure-java already adheres to. I think having conjure-dialogue as a fork is quite appropriate until it reaches a critical mass of adoption and is ready to commit to those back-compat guarantees.  When that happens, I think merging it into conjure-java makes sense.

@cakofony agreed that long-running feature branches are pretty painful. I hope that we can upstream any widely applicable improvements ASAP to minimize the forkedness.

uschi2000:
`I think it's pretty important that maintainers are free to refactor classes internally to get closer to better abstractions`
I'm advocating for exactly that. You're exactly right, the code in this repo were written a long time ago, and most of them have in practice changed very little. Feel free to refactor whatever you want and to publish a new major release every week. 

There a (at least) three independent requests to share this space from @markelliot , @cakofony , and myself. It is not reasonable for this project to simultaneously reject experimental work *inside* the project and complicate uses of its infrastructure *outside* the project.

iamdanfox:
@uschi2000 at the end of the day everything is a tradeoff right?  Sure, we could just start publishing this project and get some immediate upside for yourself, Mark and Carter, but this must be weighed against the long-term concern that wider open-source users will misunderstand our intentions and build intricate tooling that depends on the published jar. If we then go ahead and refactor things internally then all of these external users would have a bad experience of Conjure.

I think sticking to the CLI interface we committed to in [RFC002](https://github.com/palantir/conjure/blob/develop/docs/rfc/002-contract-for-conjure-generators.md) will ensure that external users are not exposed to this risk.

For your use case, I guess my recommendations would be:

- fork this repo (accepting the danger of long-lived forks and pushing to getting things upstreamed as soon as they're stable)
- copy and paste the relevant code into your repo (which are the specific classes you are most interested in relying on?)
- run `./gradlew jar` to simulate a published version of this repo, then check this into your project and add something like:

```gradle
repositories {
   flatDir {
       dirs 'libs'
   }
}
```

What do you think of the alternatives above? Are any of these feasible/infeasible?

markelliot:
So far we went with copy code, solely to get the type generators. I think I'd propose we split the type and and service generators, and sign up to a very basic contract around invoking the type generator and some sort of Conjure -> Java class name resolver interface, at least so we don't have to fix bugs in type generation in multiple repos, forked or otherwise.

uschi2000:
See https://github.com/palantir/conjure-java/pull/108 for a little readme snippet.

ferozco:
Closing in favour of PRs to this repo

PR #99, opened 2018-11-06
[WIP] Consistently resolve binary aliases in requests and responses

pkoenig10:
Follow up to #92 to correctly implement the conjure spec:
> **Accept header** - Clients MUST send an Accept: application/json header for all requests UNLESS the endpoint returns binary, in which case the client MUST send Accept: application/octet-stream. This ensures changes can be made to the wire format in a non-breaking way.

This is consistent with the behavior of other generated clients:
- Jersey - [set by Feign](https://github.com/OpenFeign/feign/blob/v8.17.0/jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java#L68-L75) due to the class-level `Produces` annotation
- Typescript - [Always sets Accept header](https://github.com/palantir/conjure-typescript/blob/3.3.0/src/commands/generate/serviceGenerator.ts#L219-L224)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/Retrofit2ServiceGenerator.java
```diff
@@ -131,25 +131,26 @@ private MethodSpec generateServiceMethod(
             EndpointDefinition endpointDef,
             TypeMapper returnTypeMapper,
             TypeMapper methodTypeMapper) {
+        boolean returnsBinary = endpointDef.getReturns()
+                .map(type -> type.accept(TypeVisitor.IS_BINARY))
```
iamdanfox: can we make sure this still works if the endpoint returns an alias of binary? Might be able to use the DealiasingTypeVisitor

pkoenig10: This is the same behavior as the Jersey service generator:
https://github.com/palantir/conjure-java/blob/40cdc417c39db37b8916371a7637fe3f55ac5985/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java#L146-L150

Correctly resolving aliases when settings the response media type is separate from this PR. But I'm not sure it is the correct thing to do. Aliases are completely separate types that wrap the type they are aliasing. You can't stream an alias of a binary.

iamdanfox: As a general rule of thumb, we try to make sure that aliases are pretty transparent so an alias of X should behave as much as possible like X.

We rely on this assumption in the conjure backcompat checker, where introducing an alias is generally considered to be backwards compatible.

pkoenig10: We currently resolve aliases for binary return types, but not for binary body parameters. It seems to me that we would want be handle both cases identically. Should we also be resolving aliases for binary body parameters?

pkoenig10:
I've updated this PR to include the following fixes:
- Correctly set the `Accept` header for retrofit clients (see PR description)
- Add the `@Streaming` annotation to retrofit endpoints returning a binary alias
- Resolve binary alias body parameters for jersey and retrofit clients (do we want this?)

conjure-java-core/src/test/resources/test/api/TestService.java.jersey
```diff
@@ -87,6 +87,12 @@ public interface TestService {
             @HeaderParam("Authorization") @NotNull AuthHeader authHeader,
             @NotNull InputStream input);
 
+    @POST
+    @Path("catalog/datasets/upload-raw-aliased")
+    void uploadAliasedRawData(
+            @HeaderParam("Authorization") @NotNull AuthHeader authHeader,
+            @NotNull InputStream input);
```
pkoenig10, line 94: Before this change, this would be generated as `NestedAliasedBinary`

ferozco:
Historically, we have treated binary and alias of binary differently. Long term, we want to aliases be completely transparent but I don't think this PR is the place to make that change

pkoenig10:
Currently they are treated differently for input types but not for returned types. I made this change in response to @dfox's comment above.

I just want clarification on the desired outcome here:
- Resolve neither returned nor parameter binary types
- Resolve only returned binary types (the current behavior)
- Resolve both returned and parameter binary types (the current impl of this PR)

ferozco:
Aliases of binary are currently treated differently from raw binary, even for return types since it is not dealiased: [jersey](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java#L152), [retrofit](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/main/java/com/palantir/conjure/java/services/Retrofit2ServiceGenerator.java#L152).

I'm advocating for keeping the current behaviour with the intention of fixing the the inconsistency later

pkoenig10:
> Aliases of binary are currently treated differently from raw binary, even for return types

This is not entirely true:
https://github.com/palantir/conjure-java/blob/2.5.0/conjure-java-core/src/test/resources/test/api/TestService.java.jersey#L58-L69

The check you linked is only used to decide whether or not to add the `@Produces` annotation. This is a correctness issue. We claim to return an `application/json` type but actually return an `application/octet-stream`.

For aliased types we should either do both or neither of:
- Change the return type to `StreamingOutput`
- Add a `@Produces` annotation with `application/octet-stream`

pkoenig10:
This should be rebased on #127

pkoenig10:
Closing for now

PR #121, opened 2018-11-27
[fix] Remove JsonIgnoreProperties from beans

ferozco:
## Before this PR
Every bean builder was generated with `@JsonIgnoreProperties(ignoreUnknown = true)`. This would override the objectMappper's configuration and allow unknown properties to be passed to servers

## After this PR
We've removed the annotation to ensure that servers correctly ignore requests with additional properties.

Follow ups:
- [ ] Add tests to the verification-client to ensure that servers fail


ferozco:
Closing since we do not think that such a impactful break should be merged

PR #125, opened 2018-11-30
[improvement] Safe log missing parameters

ellisjoe:
## Before this PR
Errors during object construction would throw an `IllegalArgumentException` with the missing fields included in the message. This message is excluded when the logs are collected.

## After this PR
Throw a `SafeIllegalArgumentException` with the missing fields in a `SafeArg` so the missing fields are included during collection.


conjure-java-core/build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
     compile 'com.google.guava:guava'
     compile 'com.palantir.conjure.java.api:errors'
     compile 'com.palantir.ri:resource-identifier'
+    compile 'com.palantir.safe-logging:preconditions'
```
iamdanfox: I think we'd need to add this to the `:conjure-lib` project to make sure the class is available for all users.

conjure-java-core/build.gradle
```diff
@@ -51,15 +52,17 @@ dependencies {
...
     testCompile 'com.palantir.conjure.java.runtime:conjure-java-retrofit2-client'
+    testCompile 'com.palantir.safe-logging:preconditions-assertj'
     testCompile 'io.dropwizard:dropwizard-testing'
+    testCompile 'junit:junit'
+    testCompile 'org.assertj:assertj-core'
+    testCompile 'org.mockito:mockito-core'
 
     integrationInputCompile project(':conjure-lib')
     integrationInputCompile 'com.palantir.conjure.java.api:errors'
+    integrationInputCompile 'com.palantir.safe-logging:preconditions'
```
iamdanfox: I don't want people to have to manually add this dependency!

iamdanfox (commented):
This looks great! The unnecessary redaction was irritating before.

iamdanfox:
Only downside I can see is this does bloat all conjure-generated projects slightly by adding the `safe-logging:preconditions` jar (and its transitive deps).

conjure-java-core/src/test/java/com/palantir/conjure/java/types/MissingFieldWireFormatTests.java
```diff
@@ -33,58 +36,66 @@
...
     }
 
     @Test
     public void missing_binary_field_should_throw() throws Exception {
-        assertThatThrownBy(() -> mapper.readValue("{}", BinaryExample.class).getBinary())
-                .isInstanceOf(IllegalArgumentException.class)
-                .hasMessage("Some required fields have not been set: [binary]");
+        assertThatLoggableExceptionThrownBy(() -> mapper.readValue("{}", BinaryExample.class).getBinary())
+                .isInstanceOf(SafeIllegalArgumentException.class)
+                .hasMessage("Some required fields have not been set")
```
carterkozak: After safe-logging is update this assertion will fail because it checks the rendered throwable message rather than the safe log message.

carterkozak:
Love it

carterkozak:
I've updated this PR, should be good to go.

ferozco (approved):

PR #128, opened 2018-12-04
[break] Consistently resolve binary aliases in service endpoints

pkoenig10:
See https://github.com/palantir/conjure-java/pull/99#issuecomment-442928096

cc @iamdanfox @ferozco 

## Before this PR
Binary alias return types are resolved but binary alias parameters are not.

## After this PR
Both binary alias return types and parameters are resolved.


pkoenig10:
I think it is a bit strange to resolve binary types at all. Aliases in conjure provide type safety, but by resolving binary aliases we're explicitly removing the type safety.

Given that there is only a single type of a streaming request/response, I think it would make more sense to not resolve binary types meaning that only the primitive `binary` type with generate a streaming input/output.

ferozco:
You're right that its strange. I think that we should think through the implications of this change and what we want the end outcome to be. Its worth taking a step back and thinking through the wider usage of `binary` within the language and the peculiarities of how java currently handles it. Happy to spend some time and chat through different approaches.

dansanduleac:
Closing this as it's been stale for a while.
Additionally, we're probably going to go with adding logic around the generated binary aliases so they behave as we want, rather than resolving them to the type that already behaves the way we want.

PR #130, opened 2018-12-05
Example generated aliases fail to compile

carterkozak:


iamdanfox:
Woah this looks like a good bug!  Seems like we're failing to add the right imports for the `valueOf` method (added in https://github.com/palantir/conjure-java/pull/77 I think).
```
AliasThree.java:39: cannot find symbol
ERROR: cannot find symbol
        return new AliasThree(AliasTwo.valueOf(value));
                                      ^
  symbol:   method valueOf(String)
  location: class AliasTwo
File: conjure-java-core/src/integrationInput/java/com/palantir/product/AliasThree.java
Line: 39
```

PR #131, opened 2018-12-05
[fix] Add Produces annotation for binary alias return types

pkoenig10:
## Before this PR
For jersey service endpoints returning a binary alias type, the return type was resolved correctly (to  `StreamingOutput` or `Response`) but the correct `@Produces` annotation was not applied.

## After this PR
The `@Produces` annotation is applied to endpoints returning binary alias types.

See discussion about resolving binary types in #128.


conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -65,11 +67,18 @@
...
 import javax.lang.model.element.Modifier;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.StreamingOutput;
 import org.apache.commons.lang3.StringUtils;
 
 public final class JerseyServiceGenerator implements ServiceGenerator {
+
     private static final ClassName NOT_NULL = ClassName.get("javax.validation.constraints", "NotNull");
 
+    private static final ClassName BINARY_METHOD_TYPE = ClassName.get(InputStream.class);
```
dansanduleac: can we call this BINARY_ARGUMENT_TYPE instead, and likewise rename MethodTypeClassNameVisitor to ArgumentTypeClassNameVisitor?
Currently it seems a bit confusing since there's no such thing as a method type.

iamdanfox (approved):
LGTM, I've pushed a couple more commits to continue the refactor and kill the 'Factory' indirection which seemed a bit redundant now.  Hope it looks OK!

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ReturnTypeClassNameVisitor.java
```diff
@@ -63,15 +64,28 @@ public TypeName visitOptional(OptionalType type) {
...
     public TypeName visitReference(com.palantir.conjure.spec.TypeName type) {
-        return BinaryReturnTypeResolver.resolveReturnReferenceType(types, type, RESPONSE_BODY_TYPE);
+        if (!types.containsKey(type)) {
+            throw new IllegalStateException("Unknown LocalReferenceType type: " + type);
+        }
+
+        TypeDefinition def = types.get(type);
+        if (def.accept(TypeDefinitionVisitor.IS_ALIAS)) {
+            Type aliasType = def.accept(TypeDefinitionVisitor.ALIAS).getAlias();
+            TypeName aliasTypeName = aliasType.accept(this);
```
iamdanfox, line 82: Long term I'd like to completely get rid of this class - it's kinda doing some 'dealiasing' here, but it's pretty convoluted with the binary specialization we have going on above.

dansanduleac (approved):

PR #133, opened 2018-12-11
Publish a source jar alongside tgz

dansanduleac:
Seems we need to publish a source jar to be admissible into JCenter:

https://www.jfrog.com/confluence/display/BT/Central+Repositories#CentralRepositories-IncludingyourPackageinJCenter

Couldn't use `nebula.source-jar` directly because it adds the artifact to the `nebula` publication, but we use a `dist` publication (and have automation that expects that name). https://github.com/nebula-plugins/nebula-publishing-plugin/blob/master/src/main/groovy/nebula/plugin/publishing/publications/SourceJarPlugin.groovy

### DIFF
--- a/conjure-java/build.gradle
+++ b/conjure-java/build.gradle
@@ -28,3 +28,21 @@ dependencies {
 
     processor 'org.immutables:value'
 }
+
+// Need to publish a source jar to be accepted on JCenter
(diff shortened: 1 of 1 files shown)


dansanduleac:
de-prioritized

PR #135, opened 2018-12-12
[improvement] upgrade conjure-lib's jackson from 2.6.7 -> 2.9.7

Dgleish:
## Before this PR

The main repo had pinned `com.fasterxml.jackson.*:jackson-* = 2.9.6`, except for `conjure-lib` which had overridden to `com.fasterxml.jackson.*:jackson-* = 2.6.7`. Given `conjure-java-runtime-api` is on `2.9.7`, thought I'd try and bump everything to that. I don't understand why `conjure-lib` had the override in the first place, but potentially there's a good reason.
It's just currently annoying for consumers of conjure who don't have a way currently of pinning versions of transitive dependencies.
## After this PR
I removed the `conjure-lib` override and set everything to use `2.9.7`

### DIFF
--- a/conjure-lib/versions.props
+++ b/conjure-lib/versions.props
@@ -1,1 +1,0 @@
-com.fasterxml.jackson.*:jackson-* = 2.6.7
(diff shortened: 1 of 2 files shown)


iamdanfox (approved):
LGTM - we're using jackson 2.9.7 pretty much everywhere else anyway!

PR #139, opened 2018-12-20
Add an ImmutableByteArray implementation

markelliot:
Preparation for replacing ByteBuffer in generated classes since ByteBuffer is not
immutable and maintains mutable state.

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
@@ -22,7 +22,6 @@ import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
-import jdk.nashorn.internal.ir.annotations.Immutable;
 
 /** An immutable {@code byte[]} wrapper. */
 public final class ImmutableByteArray {
@@ -68,8 +67,8 @@ public final class ImmutableByteArray {
 
     @Override
     public boolean equals(Object obj) {
-        return this == obj ||
-                (obj instanceof ImmutableByteArray && Arrays.equals(safe, ((ImmutableByteArray) obj).safe));
+        return this == obj
+                || (obj instanceof ImmutableByteArray && Arrays.equals(safe, ((ImmutableByteArray) obj).safe));
     }
 
     /** Constructs a new {@link ImmutableByteArray} from the provided array. */
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
@@ -22,22 +22,15 @@ import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import jdk.nashorn.internal.ir.annotations.Immutable;
 
 /** An immutable {@code byte[]} wrapper. */
 public final class ImmutableByteArray {
     private final byte[] safe;
 
-    /** Constructs a new {@link ImmutableByteArray} from the provided array. */
-    @JsonCreator
-    public ImmutableByteArray(byte[] array) {
-        this(array, 0, array.length);
-    }
-
-    /** Constructs a new {@link ImmutableByteArray} read from provided offset for the provided length. */
-    public ImmutableByteArray(byte[] array, int offset, int length) {
-        safe = new byte[length];
-
-        System.arraycopy(array, offset, safe, 0, length);
+    /** Constructs a new {@link ImmutableByteArray} assuming the provided array is not held by any other class. */
+    private ImmutableByteArray(byte[] array) {
+        safe = array;
     }
 
     /** Returns a new read-only {@link ByteBuffer} backed by this byte array. */
@@ -77,5 +70,29 @@ public final class ImmutableByteArray {
     public boolean equals(Object obj) {
         return this == obj ||
                 (obj instanceof ImmutableByteArray && Arrays.equals(safe, ((ImmutableByteArray) obj).safe));
+    }
+
+    /** Constructs a new {@link ImmutableByteArray} from the provided array. */
+    @JsonCreator
+    public static ImmutableByteArray from(byte[] array) {
+        return from(array, 0, array.length);
+    }
+
+    /** Constructs a new {@link ImmutableByteArray} read from provided offset for the provided length. */
+    public static ImmutableByteArray from(byte[] array, int offset, int length) {
+        byte[] safe = new byte[length];
+        System.arraycopy(array, offset, safe, 0, length);
+
+        return new ImmutableByteArray(safe);
+    }
+
+    /** Constructs a new {@link ImmutableByteArray} read from the provided {@link ByteBuffer} */
+    public static ImmutableByteArray from(ByteBuffer buffer) {
+        ByteBuffer local = buffer.duplicate();
+
+        byte[] safe = new byte[local.remaining()];
+        local.get(safe);
+
+        return new ImmutableByteArray(safe);
     }
 }
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
@@ -1,0 +1,81 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.lib;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+
+/** An immutable {@code byte[]} wrapper. */
+public final class ImmutableByteArray {
+    private final byte[] safe;
+
+    /** Constructs a new {@link ImmutableByteArray} from the provided array. */
+    @JsonCreator
+    public ImmutableByteArray(byte[] array) {
+        this(array, 0, array.length);
+    }
+
+    /** Constructs a new {@link ImmutableByteArray} read from provided offset for the provided length. */
+    public ImmutableByteArray(byte[] array, int offset, int length) {
+        safe = new byte[length];
+
+        System.arraycopy(array, offset, safe, 0, length);
+    }
+
+    /** Returns a new read-only {@link ByteBuffer} backed by this byte array. */
+    public ByteBuffer getByteBuffer() {
+        return ByteBuffer.wrap(safe).asReadOnlyBuffer();
+    }
+
+    /** Returns a copy of this byte array. */
+    @JsonValue
+    public byte[] getBytes() {
+        byte[] unsafe = new byte[safe.length];
+        System.arraycopy(safe, 0, unsafe, 0, safe.length);
+        return unsafe;
+    }
+
+    /** Copies this byte array into the provided byte array beginning at offset and up to the provided length. */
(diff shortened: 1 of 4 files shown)


uschi2000:
@j-baker Didn't you have one of those somewhere already?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
```diff
@@ -0,0 +1,97 @@
...
+    /** Returns a copy of this byte array. */
+    @JsonValue
+    public byte[] getBytes() {
+        byte[] unsafe = new byte[safe.length];
+        System.arraycopy(safe, 0, unsafe, 0, safe.length);
+        return unsafe;
+    }
+
+    /** Copies this byte array into the provided byte array beginning at offset and up to the provided length. */
+    public void copy(byte[] array, int offset, int length) {
```
carterkozak: would prefer to rename the `array` parameter something along the lines of `destination`

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
```diff
@@ -0,0 +1,97 @@
...
+    public ByteBuffer getByteBuffer() {
+        return ByteBuffer.wrap(safe).asReadOnlyBuffer();
+    }
+
+    /** Returns a copy of this byte array. */
+    @JsonValue
+    public byte[] getBytes() {
+        byte[] unsafe = new byte[safe.length];
+        System.arraycopy(safe, 0, unsafe, 0, safe.length);
+        return unsafe;
```
carterkozak: `return safe.clone()` is simpler.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
```diff
@@ -0,0 +1,97 @@
...
+public final class ImmutableByteArray {
+    private final byte[] safe;
+
+    /** Constructs a new {@link ImmutableByteArray} assuming the provided array is not held by any other class. */
+    private ImmutableByteArray(byte[] array) {
+        safe = array;
+    }
+
+    /** Returns a new read-only {@link ByteBuffer} backed by this byte array. */
+    public ByteBuffer getByteBuffer() {
```
carterkozak: Thoughts on `as` or `to` over `get`?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
```diff
@@ -0,0 +1,97 @@
...
+        return ByteBuffer.wrap(safe).asReadOnlyBuffer();
+    }
+
+    /** Returns a copy of this byte array. */
+    @JsonValue
+    public byte[] getBytes() {
```
carterkozak: It might be helpful to include some indication in the method name that we're making a copy, which is potentially expensive.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+
+/** An immutable {@code byte[]} wrapper. */
+public final class ImmutableByteArray {
```
carterkozak: Thoughts on naming this `Binary` to couple this with the cojure primitive? Byte array is an implementation detail.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public int hashCode() {
+        return Arrays.hashCode(safe);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return this == obj
+                || (obj instanceof ImmutableByteArray && Arrays.equals(safe, ((ImmutableByteArray) obj).safe));
+    }
```
carterkozak: Missing toString, though I'm not sure what I'd expect a binary toString to do. Perhaps class name and length?

carterkozak (commented):
Strongly in favor of this feature, thanks Mark!
I recall @schlosna was interested in this feature as well.

Old PR from @j-baker:`<internal-github>/conjure/pull/816`

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
```diff
@@ -0,0 +1,97 @@
...
+        safe = array;
+    }
+
+    /** Returns a new read-only {@link ByteBuffer} backed by this byte array. */
+    public ByteBuffer getByteBuffer() {
+        return ByteBuffer.wrap(safe).asReadOnlyBuffer();
+    }
+
+    /** Returns a copy of this byte array. */
+    @JsonValue
```
carterkozak: Should we move `@JsonValue` to `getByteBuffer` and avoid duplicating contents for serialization? Annotation may apply to the `safe` field, but I've not tested that type of usage.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
schlosna: bikeshed: method name `copyTo`?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: Maybe ImmutableBytes? “binary” translates a few different ways — sometimes as a stream — so would avoid the direct naming here. Also open to “ByteString” or just “Bytes”.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: I guess “asNewByteBuffer”?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: “asNewByteArray”?

schlosna (commented):
https://github.com/palantir/atlasdb/issues/2020 was one place where we didn't want to require a 3rd party dependency in public API that this could be useful.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
carterkozak: `Bytes` is reasonable, though I prefer `Binary`. Other types have different java bindings based on context, `int`, `OptionalInt`, `List<Integer>` for example.
Bytes sounds a bit like a utility rather than a data object (see `com.google.common.primitives.Bytes`).

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
carterkozak: Not sure `new` is necessary here, since the byte buffer isn't mutable. It happens to be new, but that makes no difference to api consumers.
Our data could just as easily be backed by a heap `ByteBuffer` rather than `byte[]`, in which case we wouldn't create new objects.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
uschi2000: why not just ByteArray? why don't prefix other classes with `Immutable` even if they are, e.g. ResourceIdentifier.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
```diff
@@ -0,0 +1,97 @@
...
+    public static ImmutableByteArray from(byte[] array, int offset, int length) {
+        byte[] safe = new byte[length];
+        System.arraycopy(array, offset, safe, 0, length);
+
+        return new ImmutableByteArray(safe);
+    }
+
+    /** Constructs a new {@link ImmutableByteArray} read from the provided {@link ByteBuffer}. */
+    public static ImmutableByteArray from(ByteBuffer buffer) {
+        ByteBuffer local = buffer.duplicate();
```
uschi2000: can you add a comment saying why this `duplicate` is necessary?

conjure-lib/src/test/java/com/palantir/conjure/java/lib/ImmutableByteArrayTests.java
```diff
@@ -0,0 +1,155 @@
...
+public final class ImmutableByteArrayTests {
+    @Test
+    public void testConstructionCopiesInputArray() {
+        byte[] input = new byte[]{0};
+        ImmutableByteArray immutable = ImmutableByteArray.from(input);
+
+        assertThat(immutable.getBytes()[0]).isEqualTo((byte) 0);
+
+        input[0] = 1;
+
```
uschi2000: nit: could cut this line (34) and the one above (30)

conjure-lib/src/test/java/com/palantir/conjure/java/lib/ImmutableByteArrayTests.java
```diff
@@ -0,0 +1,155 @@
...
+        assertThat(immutable.getBytes()[0]).isEqualTo((byte) 0);
+    }
+
+    @Test
+    public void testConstructionCopiesArrayWithRange() {
+        byte[] input = new byte[]{0, 1, 2};
```
uschi2000: also test end by doing `{0,1,2,3} --> {1,2}`?

uschi2000 (approved):
looks good modulo the naming discussion and one docs suggestion.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
carterkozak: Since this is a constant time/memory operation unlike `getBytes` the original`getByteBuffer` is probably best here.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: going to go with "Bytes"

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: added toString that includes the size but nothing else

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: yep, will rename to copyTo

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: ok, for symmetry this will be called asByteBuffer (no "new").

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: sure, I'll move it to the other method. It's more appropriate to label a method than a private member, so we'll stick with methods.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: I'd prefer to know we're doing this using the System copy primitives, and so going to keep this relatively small amount of code, confidently given test coverage.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/ImmutableByteArray.java
markelliot: done

conjure-lib/src/test/java/com/palantir/conjure/java/lib/ImmutableByteArrayTests.java
markelliot: added a comment on what line 33 is doing, believe the whitespace improves readability

conjure-lib/src/test/java/com/palantir/conjure/java/lib/ImmutableByteArrayTests.java
markelliot: done

carterkozak (approved):
thanks

PR #140, opened 2018-12-21
[feature] Undertow Server Conjure Generator

carterkozak:
Using only a dependency on `conjure-java-undertow-runtime` a service may be bootstrapped using the following:

```java
public static void main(String[] args) {
    RoutingHandler handler = Handlers.routing();
    RecipeBookServiceEndpoint.of(new RecipeBookResource())
        .create(HandlerContext.builder()
            .serializerRegistry(new SerializerRegistry(Serializers.json()))
            .build())
        .register(new ConjureRoutingRegistry(handler));
    
    Undertow server = Undertow.builder()
            .addHttpListener(8080, "0.0.0.0")
            .setHandler(handler)
            .build();
    server.start();
}
```


carterkozak:
I've got a couple API changes I'd like to make, but would prefer to avoid additional changes in this large PR.

* It should not be necessary to create a RoutingHandler, the ConjureRoutingRegistry should be renamed to ConjureHandler and implement HttpHandler.
* We should provide `HandlerContext.default()` to avoid the builder and serializer registry creation in the standard case.

ferozco (approved):
Carter walked me through the changes offline and it all seems sensible. Lets get this in so that we can continue to iterate and improve the generator in standalone PRs. Its worth noting that currently the runtime is currently checked in along side the generator for faster iteration. Once things have stabilised we factor this out into `conjure-java-runtime`

PR #141, opened 2018-12-21
[improvement] Refactor ConjureRoutingRegistry to ConjureHandler

carterkozak:
Implementing both HttpHandler and RoutingRegistry provides more
leverage to make changes in the future without breaking API
or requiring additional code changes.

### DIFF
--- a/conjure-java-server-verifier/src/test/java/com/palantir/conjure/java/verification/server/undertest/UndertowServerUnderTestRule.java
+++ b/conjure-java-server-verifier/src/test/java/com/palantir/conjure/java/verification/server/undertest/UndertowServerUnderTestRule.java
@@ -19,15 +19,13 @@ package com.palantir.conjure.java.verification.server.undertest;
 import com.google.common.reflect.Reflection;
 import com.palantir.conjure.java.undertow.lib.Endpoint;
 import com.palantir.conjure.java.undertow.lib.HandlerContext;
-import com.palantir.conjure.java.undertow.lib.RoutingRegistry;
 import com.palantir.conjure.java.undertow.lib.SerializerRegistry;
-import com.palantir.conjure.java.undertow.runtime.ConjureRoutingRegistry;
+import com.palantir.conjure.java.undertow.runtime.ConjureHandler;
 import com.palantir.conjure.java.undertow.runtime.Serializers;
 import com.palantir.conjure.verification.client.AutoDeserializeServiceEndpoint;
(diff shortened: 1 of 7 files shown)


rubenfiszel:
That's clever but now we have N fallbacks instead of a flat RoutingHandler. It's probably benign in term of performance but in the unlucky case, you will have N-1 more indirections than before.

carterkozak:
I'm not sure that I understand, what do you mean by N fallbacks?

All of our apis (the `Endpoint` impls) will be registered with a single `ConjureHandler` at the `/api` path, which may allow fallback to another handler if no paths are matched.

rubenfiszel:
You're right ignore the above.

conjure-java-server-verifier/src/test/java/com/palantir/conjure/java/verification/server/undertest/UndertowServerUnderTestRule.java
```diff
@@ -42,19 +40,17 @@ protected void before() throws Throwable {
...
-
-        RoutingRegistry routingRegistry = new ConjureRoutingRegistry(routingHandler);
+        ConjureHandler handler = new ConjureHandler();
         SerializerRegistry serializers = new SerializerRegistry(Serializers.json(), Serializers.cbor());
         HandlerContext context = HandlerContext.builder()
                 .serializerRegistry(serializers)
                 .build();
 
-        endpoint.create(context).register(routingRegistry);
+        endpoint.create(context).register(handler);
```
rubenfiszel, line 49: To enable inverting the flow, maybe we could add a method in ConjureHandler such that this could be declared as 
`handler.register(endpoint.create(context))`. 

That would enable to write directly, `.setHandler(Handlers.path().addPrefixPath("/test/api", handler.register(endpoint.create(context)))` (and even inline everything if we wanted sine no statement above break the flow.

carterkozak: We could, however in most cases I expect that the webserver owns (and does not expose) the HandlerContext, and must iterate through all `Endpoint` objects to build mappings, so I'm not sure how helpful it will be outside of the simple demo/test case.
I'd like to keep the surface area minimal for the time being, if we find that such a method makes usage easier we can revisit.

rubenfiszel (approved):

ferozco (approved):

Issue #134, opened 2018-12-11
Generated enum valueOf method is incorrectly case-insensitive


sfackler:
The valueOf method generated for enums (which is also used for deserialization) upper-cases input, in violation of the conjure spec.

https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java#L60-L71

sfackler:
https://github.com/palantir/conjure/issues/184

markelliot:
I think the case-insensitivity here is reasonably desirable and likely hard to remove without a major break (and I'm not sure it's worth a major break).

sfackler:
Should this be a spec update instead then?

I don't really see where this behavior is providing all that much value - nothing else in the code we generate is case insensitive (field names etc).

markelliot:
I'd prefer a spec update -- we added this leniency at the request of early adopters.

More broadly, we might want to consider allowing unions to contain 'null' object values and dropping enums, as they're a subset of the functionality of unions.

carterkozak:
Could you expand upon the rationale for supporting mixed case strings? Are we certain the concerns raised by early adopters are still relevant? I'd appreciate if you could link me to that discussion.

iamdanfox:
Discussion moved to https://github.com/palantir/conjure/issues/193

PR #142, opened 2018-12-25
Generate visitor interfaces for conjure enums

carterkozak:
Most uses of enums involve a switch or if-else block across values
along the lines of "boolean isTerminalState(enum)" or
"String getDescription(enum)". It's easy to add new enum values
without updating all usages, by using a visitor we catch these
bugs at compile time rather than runtime.


markelliot:
Mentioned on https://github.com/palantir/conjure-java/issues/134, but I think we should strongly consider dropping enums in favor of unions and/or switching enum code-gen to use the union code-gen, as enums are really a subset of the union functionality.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
```diff
@@ -133,6 +148,57 @@ private static TypeSpec createEnum(ClassName enumClass, Iterable<EnumValueDefini
...
+                .addStatement("return visitor.$1L(string)", VISIT_UNKNOWN_METHOD_NAME)
+                .endControlFlow();
+        ParameterizedTypeName parameterizedVisitorClass = ParameterizedTypeName.get(visitorClass, TYPE_VARIABLE);
+        return MethodSpec.methodBuilder("accept")
+                .addModifiers(Modifier.PUBLIC)
+                .addParameter(ParameterSpec.builder(parameterizedVisitorClass, "visitor").build())
+                .addTypeVariable(TYPE_VARIABLE)
+                .returns(TYPE_VARIABLE)
+                .addCode(switchBlock.build()
+                ).build();
```
rubenfiszel: weird formatting

rubenfiszel (approved):

rubenfiszel:
@markelliot One possible migration path is to keep the conjure frontend the same but compile both enum and union to the same internal representation by adding a `literal` type.

```
EnumFoo:
  values:
    - a 
    - b
    - c
```

would be equivalent to:

```
UnionFoo:
  union:
    - a: literal
    - b: literal
    - c: literal
```

(can shorten to lit)

and an instance of UnionFoo.a would generate the actual type in the ir: `literal<pkg.UnionFoo.a>`

Examples of literal types in other languages: 
- Typescript: https://basarat.gitbooks.io/typescript/docs/types/literal-types.html + https://basarat.gitbooks.io/typescript/docs/types/discriminated-unions.html
- Scala https://docs.scala-lang.org/sips/42.type.html

markelliot:
I think a special “none” type would be cleaner and avoid opening up enums to also be used as constants. It’d also satisfy a common union pain point. 

After collapsing the codegen, the union codegen probably needs some much closer inspection for clarity and perf.

carterkozak:
This proposal will cause issues with header and query parameter enum values, we would have to validate that all enum types are lit, and cannot ever receive non-lit values. fwiw I think the use cases for union and enums are fairly distinct, if anything I'd prefer to remove the specifier key from the union type which encroaches on the enums territory.

iamdanfox (approved):
:+1: Lots of the arguments for why we recommend visitors for unions also apply to enums, so I think adding visitor codegen here makes a bunch of sense. Whether we want to eventually unify them or not, I think this is a step in the right direction.

I think discussions about whether we should drop enums completely warrants a dedicated ticket on palantir/conjure as this would affect all languages.

carterkozak:
Planning to merge tomorrow unless there are strong reservations.

PR #143, opened 2018-12-25
Enum values are case sensitive, values are validated

carterkozak:
The Conjure specification requires upper case string values for
enum constants.

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
@@ -164,15 +164,15 @@ public final class WireFormatTests {
         assertThat(mapper.readValue("\"ONE_HUNDRED\"", EnumExample.class)).isEqualTo(EnumExample.ONE_HUNDRED);
         assertThatExceptionThrownRootCause(() -> mapper.readValue("\"one\"", EnumExample.class))
                 .isInstanceOf(SafeIllegalArgumentException.class)
-                .hasLogMessage("Enum values must use SCREAMING_SNAKE_CASE")
+                .hasLogMessage("Enum values must use UPPER_SNAKE_CASE")
                 .hasExactlyArgs(UnsafeArg.of("value", "one"));
         assertThatExceptionThrownRootCause(() -> mapper.readValue("\"onE\"", EnumExample.class))
                 .isInstanceOf(SafeIllegalArgumentException.class)
-                .hasLogMessage("Enum values must use SCREAMING_SNAKE_CASE")
+                .hasLogMessage("Enum values must use UPPER_SNAKE_CASE")
                 .hasExactlyArgs(UnsafeArg.of("value", "onE"));
         assertThatExceptionThrownRootCause(() -> mapper.readValue("\"oNE\"", EnumExample.class))
                 .isInstanceOf(SafeIllegalArgumentException.class)
-                .hasLogMessage("Enum values must use SCREAMING_SNAKE_CASE")
+                .hasLogMessage("Enum values must use UPPER_SNAKE_CASE")
                 .hasExactlyArgs(UnsafeArg.of("value", "oNE"));
     }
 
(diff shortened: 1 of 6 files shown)


uschi2000 (approved):

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -152,11 +152,11 @@ public void testEmptyObjectsDeserialize() throws Exception {
     }
 
     @Test
-    public void testEnumCasingDeserializationInvariantToInputCase() throws Exception {
+    public void testEnumCasingDeserialization() throws Exception {
```
uschi2000: is there no corresponding test in conjure-verifier? (I think there should be)

markelliot:
I think it'd be preferable to finish the discussion on https://github.com/palantir/conjure-java/issues/134 prior to merging this. As noted, this might be a significant break since early adopters requested this leniency during decoding.

carterkozak:
ack

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -152,11 +152,11 @@ public void testEmptyObjectsDeserialize() throws Exception {
...
         assertThat(mapper.readValue("\"ONE\"", EnumExample.class)).isEqualTo(EnumExample.ONE);
-        assertThat(mapper.readValue("\"one\"", EnumExample.class)).isEqualTo(EnumExample.ONE);
```
iamdanfox, line 157: I think this is a pretty dangerous behaviour change in the case where users might have already persisted lowercase enum values to a database. Would prefer this to throw.

iamdanfox (changes requested):
Spec discussion ongoing in https://github.com/palantir/conjure/issues/193

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
carterkozak: That's fair, I will add validation matching our schema regex to the valueOf(String) statement

carterkozak: For now I'll just validate chars `A-Z_` without dealing with leading/trailing/duplicate underscores, we can add that later.

iamdanfox (approved):

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -154,9 +161,30 @@ public void testEmptyObjectsDeserialize() throws Exception {
...
+                .isInstanceOf(SafeIllegalArgumentException.class)
+                .hasLogMessage("Enum values must use SCREAMING_SNAKE_CASE")
+                .hasExactlyArgs(UnsafeArg.of("value", "one"));
+        assertThatExceptionThrownRootCause(() -> mapper.readValue("\"onE\"", EnumExample.class))
+                .isInstanceOf(SafeIllegalArgumentException.class)
+                .hasLogMessage("Enum values must use SCREAMING_SNAKE_CASE")
+                .hasExactlyArgs(UnsafeArg.of("value", "onE"));
+        assertThatExceptionThrownRootCause(() -> mapper.readValue("\"oNE\"", EnumExample.class))
+                .isInstanceOf(SafeIllegalArgumentException.class)
+                .hasLogMessage("Enum values must use SCREAMING_SNAKE_CASE")
```
iamdanfox: Could we tweak this error message to just say `UPPER_SNAKE_CASE`? I think the 'screaming' bit might be a bit more humour than java programmers can handle.

Edit I'm just gonna push this fix and merge I think.

carterkozak: Sounds good, thanks Dan!

iamdanfox: ah whoops, can I have some perms?

PR #145, opened 2018-12-25
[Undertow] Move the Serializer api and SerializerRegistery impl to "-runtime"

carterkozak:
While we don't necessarily want to allow other SerializerRegistry
implementations, the logic belongs in our runtime module, not
the library.


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/SerializerRegistry.java
```diff
@@ -16,77 +16,16 @@
...
-                    if (serializer.supportsContentType(acceptValue)) {
-                        return serializer;
-                    }
-                }
-            }
-        }
-        // Fall back to the default
-        return defaultSerializer;
-    }
+public interface SerializerRegistry {
```
rubenfiszel, line 24: In the current form as an interface, the methods make it more akin to a "HttpExchangeSerializer" than a registry.
Since the Serializer has been moved to the runtime library, one could argue that implementing it as a SerializerRegistry is an implementation detail of the runtime.

carterkozak: I opted to avoid changing the generated code in this PR.

There are some changes I'd like to make to the `Serializer` and `SerializerRegistry` APIs to avoid blocking ourselves in a corner:

Serializer -> Encoding
SerializerRegistry -> SerializationContext

Update serializer and serializerregistry(context?) to create typed Serializer and Deserializer objects to avoid type info lookups per request

rubenfiszel: Sounds good. If you're looking for inspiration for typesafe encoding/decoding, this library could be insightful, even just for the API: https://circe.github.io/circe/
Unfortunately, it relies on implicits which don't exist in Java.

iamdanfox (approved):

PR #146, opened 2018-12-25
[Undertow] Runtime module supports non-service response status

carterkozak:
Fixes existing test for unprocessable media.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/FrameworkException.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/FrameworkException.java
@@ -1,0 +1,80 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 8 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/FrameworkException.java
```diff
@@ -0,0 +1,80 @@
...
+package com.palantir.conjure.java.undertow.runtime;
+
+import com.google.common.collect.ImmutableList;
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeLoggable;
+import io.undertow.util.StatusCodes;
+import java.util.List;
+
+/** Internal type to signal a conjure protocol-level failure with a specific response code. */
+final class FrameworkException extends RuntimeException implements SafeLoggable {
```
iamdanfox, line 26: This entire class exists just because conjure doesn't support Unprocessable entity / unsupported media type right?

https://github.com/palantir/conjure-java-runtime-api/blob/develop/errors/src/main/java/com/palantir/conjure/java/api/errors/ErrorType.java#L39-L48

iamdanfox (approved):

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/FrameworkException.java
carterkozak: That is correct, because unprocessable entity is an rpc concern that we would not want to expose for services to throw.

Sorry I missed the notification on this last week!

PR #148, opened 2018-12-27
[fix] [Undertow] handler sets MDC state for both header and cookie auth

carterkozak:
Previously all requests were checked for an Authorization header,
which was parsed into a jwt. Now we take advantage of codegen to
check the correct place, and only for authenticated APIs.

Cookie auth token data has never been propagated to thread state
for the jersey handlers.


uschi2000:
looks like a lot of this implementation assumes that requests are bound to a single thread uninterruptedly. is that the case even if/when we move more async models of operation?

carterkozak:
When we support async operations I believe the plan is to use ListenableFuture results, rather than provide an executor to the app. If we do eventually need to wrap other invocations with logging context data, we will still need the initial resource method to be wrapped in this way.

In the future we may want to add a pt-executors executor inheritable threadlocal to propagate values to additional threads, but this commit gives what we need in order to implement that.

carterkozak:
Asynchronously reading request data prior to the auth.parse call could get weird, but we can code-gen handlers extract jwt data on an io thread and add it to the attachment proposed in https://github.com/palantir/conjure-java/pull/149

I don't want to prematurely optimize on that until we start looking into async request parsing.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/internal/Auth.java
```diff
@@ -19,26 +19,67 @@
...
+        return setState(StringDeserializers.deserializeBearerToken(
+                exchange.getRequestCookies().get(cookieName).getValue()));
+    }
+
+    /**
+     * Attempts to extract a {@link UnverifiedJsonWebToken JSON Web Token} from the
+     * {@link BearerToken} value, and populates the SLF4J {@link MDC} with
+     * user id, session id, and token id extracted from the JWT. This is
+     * best-effort and does not throw an exception in case any of these steps fail.
+     */
```
iamdanfox, line 68: :+1: appreciate the thorough javadoc!

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/BearerTokenLoggingTest.java
```diff
@@ -69,7 +71,10 @@ public void before() {
         delegateRunnable.set(null);
         exchange = HttpServerExchanges.createStub();
         exchange.setRequestMethod(Methods.GET);
-        handler = new BearerTokenLoggingHandler(delegate);
+        handler = new LoggingContextHandler(httpServerExchange -> {
+            Auth.header(httpServerExchange);
```
iamdanfox, line 75: From the method signature, it's a tiny bit surprising that this has the side effect of writing to the MDC too... do you think it's worth trying to get the acronym 'MDC' in there somewhere? Maybe something like `MdcAuth.header()` or `Auth.headerToMdc`?

iamdanfox (approved):

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/BearerTokenLoggingTest.java
carterkozak: I'd prefer to keep the current method names -- in the future we may want to code-gen a handler to extract auth tokens on an IO thread, and update this method to do a simple attachment lookup. This way we avoid expanding the API based on implementation details.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/LoggingContextHandler.java
```diff
@@ -0,0 +1,45 @@
...
+
+    LoggingContextHandler(HttpHandler delegate) {
+        this.delegate = Preconditions.checkNotNull(delegate, "Delegate handler is required");
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        // Jersey filter implementations often fail to clean up MDC state after each request.
+        // In environments using both Jersey and generated Undertow handlers, the MDC may
+        // contain stale values.
```
dansanduleac, line 37: This might save us a lot of pain! I remember stale uids in the logs confusing people in the past.

PR #150, opened 2018-12-29
[fix] [Undertow] Streaming responses cause client exceptions after partial write

### LINKED ISSUE #16: Errors not handled properly when streaming responses
Currently, when an endpoint returns `binary` we render a `StreamingOutput` in the jersey interface. Because the Content-Length is unknown, Jetty properly responds using [chunked transfer encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding) which allows the client to determine whether the stream has been fully received or failed part way through. There's an issue in Jersey, though, which actually closes the stream when an exception is thrown after some bytes have been written to the client which in turn causes Jetty to close the stream as if it were completed successfully: https://en.wikipedia.org/wiki/Chunked_transfer_encoding

There's not much we can do about streaming back things of unknown size, but we should be able to support streams where the size is known. I'd propose that we add a `BinaryStream` or `Binary` or something which takes the place of `StreamingOutput` in conjure generated interfaces and has two methods:
```
of(StreamingOutput)
of(StreamingOutput, long contentLength)
```
think it may be reasonable to stick this in remoting, but sounds like these two projects are converging and we're going to need to change the conjure code gen at some point.

This change also doesn't affect clients of these endpoints because they must use the Retrofit client when consuming streaming endpoints.

carterkozak:
This fixes #16 for anyone using the undertow handlers, not for
jersey.


dansanduleac (approved):

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -351,6 +351,23 @@ public void testBinaryOptionalPresentResponse() throws Exception {
...
     }
 
+    @Test
+    public void testBinaryServerSideFailureAfterManyBytesSent() throws Exception {
+        Response<ResponseBody> response = binaryClient.getBinaryFailure(AuthHeader.valueOf("authHeader"),
+                // Write more bytes than one buffer
+                20000).execute();
+        try (ResponseBody body = response.body()) {
+            assertThatThrownBy(() -> ByteStreams.copy(body.byteStream(), ByteStreams.nullOutputStream()))
+                    .isInstanceOf(IOException.class);
```
ellisjoe, line 361: out of curiosity, what's the exact IOException here?

this seems to handle exceptions properly, which would mean the undertow-jersey impl is different than the jetty-jersey impl in the way it handles exceptions. original issue with jersey here: https://github.com/eclipse-ee4j/jersey/issues/3850

carterkozak: EOFException.

undertow+jersey fails the same way as jetty+jersey, this implementation uses generated undertow handlers, bypassing servlets entirely.

ellisjoe: oh I see. so there's no jersey here at all?

carterkozak: That is correct.

PR #151, opened 2018-12-30
[Undertow] Remove unnecessary Undertow binary string deserializers

carterkozak:
Conjure does not allow binary data outside of request and response
bodies.


rubenfiszel (approved):

iamdanfox:
Hang on - fields of the Conjure `binary` type can appear within JSON objects, but they're just transported as base64...  this is particularly useful when someone wants to send some cryptographic binary data around.

carterkozak:
StringDeserializers is only used for query and header parameters, the conjure parser does not allow the binary datatype in query param or header fields.

iamdanfox (approved):
Ah OK makes sense!

PR #152, opened 2018-12-31
[fix] Undertow handler sets 204 status for methods with no return

carterkozak:
Previously they resulted in a 200 status with no response body.


conjure-java-core/src/integrationInput/java/com/palantir/product/EteBinaryServiceEndpoint.java
```diff
@@ -74,7 +75,7 @@ public void handleRequest(HttpServerExchange exchange) throws IOException {
                 if (result.isPresent()) {
                     BinarySerializers.serialize(result.get(), exchange);
                 } else {
-                    exchange.setStatusCode(204);
+                    exchange.setStatusCode(StatusCodes.NO_CONTENT);
```
carterkozak, line 77: This is easier to read, nobody will have to look up 204. Compiled bytecode is identical.

ferozco (approved):

PR #154, opened 2019-01-03
[fix] Undertow endpoints set Content-Length for small responses.

carterkozak:
Responses below the undertow buffer size (generally 16k) set
a Content-Length response header, additionally allowing response
compression to be avoided in environments which compress chunked
responses.


uschi2000 (approved):
Where does the content length actually get written to the response? does undertow do it internally?

carterkozak:
Undertow sets Content-Length internally when we close the stream prior to filling a buffer or flushing/committing data:
https://github.com/undertow-io/undertow/blob/ff4c9cf37872cb96070ba6a2fcbbaa6df291e390/core/src/main/java/io/undertow/io/UndertowOutputStream.java#L333

PR #155, opened 2019-01-04
[Undertow] Add web security and cache-control headers

carterkozak:


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -88,11 +94,35 @@ public ConjureHandler delete(String template, HttpHandler handler) {
...
+
+        private final HttpString method;
+        private final String template;
+
+        EndpointDetails(HttpString method, String template) {
+            this.method = method;
+            this.template = template;
+        }
+
+        HttpString getMethod() {
```
uschi2000: IMO, since this is a private class it's fine to omit the getters and just access the fields directly

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/NoCachingResponseHandler.java
```diff
@@ -0,0 +1,63 @@
...
+ * desirable and that such reuse is the default behavior when no requirement or local configuration prevents it."
+ */
+final class NoCachingResponseHandler implements HttpHandler {
+
+    private static final String DO_NOT_CACHE = "no-cache, no-store, must-revalidate";
+
+    private final HttpHandler next;
+
+    NoCachingResponseHandler(HttpHandler next) {
+        this.next = Preconditions.checkNotNull(next, "Delegate HttpHandler is required");
```
uschi2000: IMO fine to avoid this check. We generally assume that everything is non-null unless explicitly documented

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/NoCachingResponseHandler.java
```diff
@@ -0,0 +1,63 @@
...
+        next.handleRequest(exchange);
+    }
+
+    private enum Listener implements ConduitWrapper<StreamSinkConduit> {
+        INSTANCE;
+
+        @Override
+        public StreamSinkConduit wrap(ConduitFactory<StreamSinkConduit> factory, HttpServerExchange exchange) {
+            HeaderMap responseHeaders = exchange.getResponseHeaders();
+            if (!responseHeaders.contains(Headers.CACHE_CONTROL)) {
```
uschi2000: should we instead forcefully set the header? under what conditions would it be set already and why would we want to support that configurability?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java
```diff
@@ -0,0 +1,80 @@
...
+import io.undertow.server.ConduitWrapper;
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.util.ConduitFactory;
+import io.undertow.util.HeaderMap;
+import io.undertow.util.Headers;
+import io.undertow.util.HttpString;
+import org.xnio.conduits.StreamSinkConduit;
+
+final class WebSecurityHandler implements HttpHandler {
```
uschi2000, line 44: add high-level docs, maybe a link to the authoritative source of information of how-to-secure-the-web?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java
```diff
@@ -0,0 +1,80 @@
...
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        exchange.addResponseWrapper(Listener.INSTANCE);
+        next.handleRequest(exchange);
+    }
+
+    private enum Listener implements ConduitWrapper<StreamSinkConduit> {
+        INSTANCE;
+
+        @Override
+        public StreamSinkConduit wrap(ConduitFactory<StreamSinkConduit> factory, HttpServerExchange exchange) {
```
uschi2000: @walkerburgin could you take a look at this, please?
@cakofony let's also get someone from infosec to take a look?

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandlerTest.java
```diff
@@ -0,0 +1,92 @@
...
+    public void testWebSecurityHeaders() throws IOException {
+        try (Response response = client.newCall(new Request.Builder()
+                .get().url("http://localhost:12345").build()).execute()) {
+            validateCommonResponseHeaders(response);
+            assertThat(response.header("X-Content-Security-Policy")).isNull();
+        }
+    }
+
+    @Test
+    public void testInternetExploder10() throws IOException {
```
uschi2000: let's avoid throwing punches in public repos

uschi2000:
lgtm modulo some nits, but would appreciate some infosec eyes

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/NoCachingResponseHandler.java
carterkozak: This could be configured in a manually built endpoint, or if we add a feature similar to the jersey response-as-binary flag. We have tests elsewhere that validate the default can be overridden, but we could ignore those for undertow endpoints if we think that's a bad idea. I don't have a strong preference, but biased toward providing the same functionality.

uschi2000: I think my preference is for conjure endpoints to be zero config(urability)

rubenfiszel:
We should prefix those PRs with [Undertow] or equivalent

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -45,7 +51,7 @@
             // Bearer token and trace handler must execute prior to the exception
             // to provide user and trace information on exceptions.
             (endpoint, handler) -> new BearerTokenLoggingHandler(handler),
-            TraceHandler::new,
+            (endpoint, handler) -> new TraceHandler(endpoint.method + " " + endpoint.template, handler),
```
rubenfiszel, line 59: nit: Are there other places where the TraceHandler is instantiated ? Otherwise I would prefer to pass the new EndpointDetails and let the TraceHandler decides how it wants to format it.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -88,11 +94,27 @@ public ConjureHandler delete(String template, HttpHandler handler) {
...
-        for (BiFunction<String, HttpHandler, HttpHandler> wrapper : WRAPPERS) {
+        EndpointDetails endpoint = new EndpointDetails(method, template);
+        for (BiFunction<EndpointDetails, HttpHandler, HttpHandler> wrapper : WRAPPERS) {
             current = wrapper.apply(endpoint, current);
         }
         routingHandler.add(method, template, current);
         return this;
     }
+
+    private static final class EndpointDetails {
```
rubenfiszel, line 110: Seems like the only thing you can do with an EndpointDetails is to display it. That seems like a weird abstraction. Either add getters (or make it immutable) or remove the class in favor of a static method "String displayEndpoint".

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandlerTest.java
```diff
@@ -0,0 +1,92 @@
...
+    private static Undertow server;
+
+    private static final OkHttpClient client = new OkHttpClient.Builder().build();
+
+    @BeforeClass
+    public static void beforeClass() {
+        server = Undertow.builder()
+                .addHttpListener(12345, "localhost")
+                .setHandler(new WebSecurityHandler(exchange -> {
+                }))
```
rubenfiszel: inline

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
carterkozak: EndpointDetails is a private internal structure, the TraceHandler just needs an operation name, I'd prefer not to leak the EndpintDetails abstraction here

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
carterkozak: We use it for the cache-control handler to avoid handler registration for non-GET requests. @uschi2000 requested that I remove the getters in an earlier review.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
rubenfiszel: Could we just pass in the method and template as separated args ? Not strongly opinionated about it but it feel a bit strange in terms of separation of concerns.

carterkozak: I'd like to avoid that, by design I don't want the trace handler to be smart, it should be impossible to make it calculate operation names per request because that can leak unsafe data.

rubenfiszel: 👍

ghost:
+1 from AppSec

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java
```diff
@@ -0,0 +1,80 @@
...
+            HeaderMap headers = exchange.getResponseHeaders();
+            headers.put(Headers.CONTENT_SECURITY_POLICY, CONTENT_SECURITY_POLICY);
+            headers.put(Headers.REFERRER_POLICY, REFERRER_POLICY);
+
+            String userAgent = exchange.getRequestHeaders().getFirst(Headers.USER_AGENT);
+            if (userAgent != null) {
```
walkerburgin: [This](https://content-security-policy.com/) page suggests that we may run into trouble setting both `content-security-policy` and `x-content-security-policy`: 

> Note: It is known that having both Content-Security-Policy and X-Content-Security-Policy or X-Webkit-CSP causes unexpected behaviours on certain versions of browsers. Please avoid using deprecated X-* headers.

(But if we haven't seen any issues in practice it may not be worth changing here)

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java
```diff
@@ -0,0 +1,80 @@
...
+import io.undertow.util.ConduitFactory;
+import io.undertow.util.HeaderMap;
+import io.undertow.util.Headers;
+import io.undertow.util.HttpString;
+import org.xnio.conduits.StreamSinkConduit;
+
+final class WebSecurityHandler implements HttpHandler {
+
+    private static final String CONTENT_SECURITY_POLICY =
+            "default-src 'self'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; frame-ancestors 'self';";
```
walkerburgin, line 47: `img-src data:` and `style-src 'unsafe-inline'` might be a little too permissive / surprising for defaults. But I guess it may not matter if these headers are being applied predominantly to JSON API endpoints and not page endpoints in practice.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java
uschi2000: was this resolved?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java
uschi2000: was this resolved?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java
carterkozak: based on `if we haven't seen any issues in practice it may not be worth changing here` I'd say so. I prefer not to make changes here unless things are currently broken, since this is just applying existing logic in a new format.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java
carterkozak: I'm happy leaving this as is unless there are serious concerns. If we'd like to lock this down, we should coordinate between all the projects which define this value.

iamdanfox: :+1: this matches the internal content security policy we have in WC already

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java
```diff
@@ -0,0 +1,82 @@
...
+        headers.put(Headers.X_FRAME_OPTIONS, FRAME_OPTIONS);
+        headers.put(Headers.X_XSS_PROTECTION, XSS_PROTECTION);
+        String userAgent = exchange.getRequestHeaders().getFirst(Headers.USER_AGENT);
+        if (userAgent != null) {
+            // send the CSP header so that IE10 and IE11 recognise it
+            if (userAgent.contains(USER_AGENT_IE_10) || userAgent.contains(USER_AGENT_IE_11)) {
+                headers.put(HEADER_IE_X_CONTENT_SECURITY_POLICY, CONTENT_SECURITY_POLICY);
+            }
+        }
+        next.handleRequest(exchange);
```
iamdanfox, line 80: :+1: this block looks exactly the same as WC one we have internally.

iamdanfox (approved):

PR #158, opened 2019-01-07
[break] Convert from ByteBuffer to Bytes

markelliot:


markelliot:
I can't figure out how to resolve the one failing test from the verifier. Any tips?

conjure-java-core/src/integrationInput/java/com/palantir/product/BinaryAliasExample.java
```diff
@@ -2,21 +2,21 @@
...
 
-    private BinaryAliasExample(ByteBuffer value) {
+    private BinaryAliasExample(Bytes value) {
         Objects.requireNonNull(value, "value cannot be null");
         this.value = value;
     }
 
     @JsonValue
-    public ByteBuffer get() {
+    public Bytes get() {
```
iamdanfox, line 19: Just flagging that this is a major break and any conjure-java users who have `binary` in their API will impose a compilation break on all their downstream consumers.

Can we prefix the PR with `[break]` so that it's clear this should be a major rev? (Alternatively, perhaps there should be some kind of preview flag?)

carterkozak:
Link to the description of bug highlighted by the verifier: https://github.com/palantir/conjure-java/pull/162#discussion_r245789858

carterkozak:
You should be able to get tests passing by moving the `JsonCreator` annotation to `from(ByteBuffer)` from `from(byte[])`

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
```diff
@@ -96,6 +95,7 @@ public static Bytes from(byte[] array, int offset, int length) {
     }
 
     /** Constructs a new {@link Bytes} read from the provided {@link ByteBuffer}. */
+    @JsonCreator
```
carterkozak, line 98: Would you mind adding a serialized form tests including a negative test for `[]` so it's easier for the next dev to debug this type of failure?

iamdanfox:
A quick bit of searching shows 17 Conjure definitions using binary, including some of our most core Foundry services. 

This change looks good, but I'm uncomfortable causing a compilation break in all these services' APIs, especially when the ergonomics our new 'Bytes' class hasn't been proven yet.  Could we phase this in using a feature flag instead? (Or provide a feature flag to preserve the old behaviour?)

I really don't want to couple this change to other unrelated conjure-java changes (e.g. the undertow generator) as I'm concerned people will just not upgrade.

carterkozak:
> A quick bit of searching shows 17 Conjure definitions using binary

Is this excluding streamed request/response bodies, which are handled differently?

Fwiw another very large internal project is nearly done migrating bean binary fields to this class.

carterkozak:
Based on what it took to move said large internal project from ByteBuffer->Bytes, I don't imagine it will take a great deal of effort to migrate those 17 definitions either. I wouldn't be opposed to this merging without a feature flag, but I defer to your judgement.

👍

carterkozak:
Has there been offline movement on this?

markelliot:
Nope, haven't had time to build this against a feature flag.

carterkozak:
I've added a feature flag in #226

carterkozak:
Closing in favor of #226

PR #159, opened 2019-01-07
[Undertow] Support for enum values as query parameters

carterkozak:


conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -487,27 +491,89 @@ private CodeBlock decodePlainParameterCodeBlock(Type type, TypeMapper typeMapper
...
-                    typeMapper.getClassName(type),
-                    resultVarName,
-                    typeMapper.getClassName(type.accept(TypeVisitor.OPTIONAL).getItemType()),
-                    StringDeserializers.class,
-                    paramsVarName,
-                    paramId
-            );
-        }
-        return CodeBlocks.statement(
+        Optional<CodeBlock> complexDeserializer = getComplexTypeDeserializer(
```
rubenfiszel: For once, I would argue that it might be more readable using an if-else statement:

`if (isPrimitive) ... else ...`

carterkozak: ha, updated

rubenfiszel (approved):

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -510,6 +507,78 @@ private CodeBlock decodePlainParameterCodeBlock(Type type, TypeMapper typeMapper
...
+                typeMapper.getClassName(type),
+                resultVarName,
+                StringDeserializers.class,
+                functionName,
+                paramsVarName,
+                paramId,
+                typeMapper.getClassName(getComplexType(type))));
+    }
+
+    private Type getComplexType(Type type) {
```
iamdanfox, line 571: In conjure documentation, we've been using the term 'container' types to refer to list, set, map and optional.

https://palantir.github.io/conjure/#/docs/concepts?id=container-types

EDIT- actually where will 'map<K,V>' fit in to all this? Is this method just for generics with a single param?

carterkozak: Map is not allowed in query/path/header parameters

carterkozak: This will never return a container type, just the contained type in the presence of a container, though an imported type is considered complex, and we use the valueOf function on it for parity with jersey.
I should clearly write some javadoc :-)

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/internal/StringDeserializers.java
```diff
@@ -425,4 +426,38 @@ public static UUID deserializeUuid(Iterable<String> in) {
...
+        }
+        ImmutableList.Builder<T> builder = ImmutableList.builder();
+        for (String item : in) {
+            builder.add(deserializeComplex(item, factory));
+        }
+        return builder.build();
+    }
+
+    public static <T> Set<T> deserializeComplexSet(Iterable<String> in, Function<String, T> factory) {
+        if (in == null) {
```
uschi2000, line 465: should add `@Nullable` (same above)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
carterkozak: Updated

conjure-java-core/src/test/resources/ete-service.yml
```diff
@@ -132,3 +135,36 @@ services:
...
+            param-type: query
+            type: SimpleEnum
+        returns: SimpleEnum
+
+      enumListQuery:
+        http: GET /enum/list/query
+        args:
+          queryParamName:
+            param-type: query
+            type: list<SimpleEnum>
```
uschi2000, line 152: interestingly, the wire format spec doesn't allow this: `If the param-type is path, header or query then the de-aliased type MUST be an enum, a primitive or an optional of one of these. `

https://palantir.github.io/conjure/#/docs/spec/conjure_definitions

conjure-java-core/src/integrationInput/java/com/palantir/product/EteServiceRetrofit.java
```diff
@@ -112,4 +113,28 @@
...
+    @Headers({"hr-path-template: /base/enum/query", "Accept: application/json"})
+    Call<SimpleEnum> enumQuery(
+            @Header("Authorization") AuthHeader authHeader,
+            @Query("queryParamName") SimpleEnum queryParamName);
+
+    @GET("./base/enum/list/query")
+    @Headers({"hr-path-template: /base/enum/list/query", "Accept: application/json"})
+    Call<List<SimpleEnum>> enumListQuery(
+            @Header("Authorization") AuthHeader authHeader,
+            @Query("queryParamName") List<SimpleEnum> queryParamName);
```
iamdanfox, line 127: I don't think we were intending to support Lists or Sets in query / header paramters?

> If the param-type is path, header or query then the de-aliased type MUST be an enum, a primitive or an optional of one of these. Additionally, the type MUST NOT be bearertoken when the param-type is header or query

https://palantir.github.io/conjure/#/docs/spec/conjure_definitions

uschi2000:
code looks fine, and yet: we need to either update the language in the spec (https://palantir.github.io/conjure/#/docs/spec/conjure_definitions) to allow list<enum> and add conjure-verification tests, or ban these constructs in the conjure yaml.

conjure-java-core/src/integrationInput/java/com/palantir/product/EteServiceRetrofit.java
carterkozak: We already support both lists and sets in query parameters, we use them heavily across the board, this provides parity with the jersey implementation.

Happy to update the spec, but I'd prefer not to block this change on it.

conjure-java-core/src/integrationInput/java/com/palantir/product/EteService.java
```diff
@@ -130,4 +156,14 @@ default StringAliasExample optionalAliasOne(AuthHeader authHeader) {
...
+
+    @Deprecated
+    default List<SimpleEnum> enumListQuery(AuthHeader authHeader) {
+        return enumListQuery(authHeader, Collections.emptyList());
+    }
+
+    @Deprecated
+    default Optional<SimpleEnum> optionalEnumQuery(AuthHeader authHeader) {
+        return optionalEnumQuery(authHeader, Optional.empty());
+    }
```
iamdanfox, line 168: Rather than having to build up test cases in this EteService.java (which only benefits this one particular project) I'd prefer to improve the test cases in the conjure-verification project: https://github.com/palantir/conjure-verification/pull/109

This ensures that all language implementations will be forced to handle enums as path, query, header params!

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/internal/StringDeserializers.java
carterkozak: Good call, fixed!

carterkozak:
PR: https://github.com/palantir/conjure/pull/191

conjure-java-core/src/test/resources/ete-service.yml
carterkozak: Opened palantir/conjure#191

iamdanfox (commented):
I'd like to get the new conjure-verification 0.16.5 tests passing here.  Don't think we should block this PR if it turns out to be a much larger effort to make conjure-verification support lists of enums as query parameters.

carterkozak:
+1 for merging this to unblock verifier fixes -- I follow up once the verifier has been upgraded to deduplicate cases from the undertow ete test service.

iamdanfox:
@cakofony can you cherry-pick 14a477f? I can't push to your branch.

Separately, I think it would be great to the conjure-verification tests actually running against some of the generated undertow handlers!!

carterkozak:
> can you cherry-pick 14a477f? I can't push to your branch.

Currently all java implementations will fail the negative test on "!!!" because it's not rejected, but used as a valid UNKNOWN string value

carterkozak:
I was planning to fix enum validation on a separate branch because it's not related to undertow

iamdanfox:
OK fair enough - I tried to upgrade to 0.16.5 independently from this PR but the undertow generator threw an error as part of the pre-test codegen!

iamdanfox (approved):

PR #160, opened 2019-01-07
Simplify Bytes.toString implementation

carterkozak:


markelliot:
the original implementation matches the style of the beans we generate

carterkozak:
I don't believe they use `.class.getSimpleName()`. They also memoize hashcode values, but I wouldn't recommend doing that here.

markelliot:
would be fine to remove the .getSimpleName() stuff in favor of hard-coding `"Bytes"` but prefer we match the style of the codebase.

carterkozak:
Updated

markelliot (approved):

PR #162, opened 2019-01-07
fix #161: Annotate Bytes with custom serializers to avoid double-allocation

### LINKED ISSUE #161: Bytes type should avoid re-allocating input contents from jackson
This operation is twice as expensive as it needs to be, we can provide a custom deserializer rather than annotations to avoid this multiplier.

carterkozak:
fixes #161

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
@@ -16,14 +16,23 @@
 
 package com.palantir.conjure.java.lib;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 
 /** An immutable {@code byte[]} wrapper. */
+@JsonSerialize(using = Bytes.Serializer.class)
+@JsonDeserialize(using = Bytes.Deserializer.class)
 public final class Bytes {
     private final byte[] safe;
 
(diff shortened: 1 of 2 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
```diff
@@ -105,4 +112,19 @@ public static Bytes from(ByteBuffer buffer) {
...
+        public void serialize(Bytes value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
+            gen.writeBinary(value.safe);
+        }
+    }
+
+    static final class Deserializer extends JsonDeserializer<Bytes> {
+        @Override
+        public Bytes deserialize(JsonParser parser, DeserializationContext ctxt) throws IOException {
+            // Avoid making a copy of the value from jackson
+            return new Bytes(parser.getBinaryValue());
```
markelliot, line 127: how do you guarantee that the underlying byte array is never modified?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
```diff
@@ -105,4 +112,19 @@ public static Bytes from(ByteBuffer buffer) {
 
         return new Bytes(safe);
     }
+
+    static final class Serializer extends JsonSerializer<Bytes> {
+        @Override
+        public void serialize(Bytes value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
+            gen.writeBinary(value.safe);
```
markelliot, line 119: how do you guarantee that writeBinary does not mutate `safe`?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
carterkozak: We must trust jackson.
This change should fix https://github.com/palantir/conjure-java/pull/158

markelliot: why does this fix #158? (as noted on #158, it's not clear what exactly is failing or how to debug it)

carterkozak: It looks like the verifier is testing that `[]` is not valid binary data because our spec requires base-64 encoded strings. Jacksons reflective deserialization allows a json array of bytes to deserialize into `byte[]`. By using JsonParser.getBinaryValue we disallow that codepath.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
iamdanfox: The `JsonGenerator#writeBinary(bytes[])` delegates to an abstract method, so technically someone could supply an implementation which does modify bytes. If we're going for zero-copy perf though I think we might just have to trust Jackson!

carterkozak: Given that we fully control the jackson object mappers used by conjure implementations, I think this is reasonable.

iamdanfox (approved):

PR #163, opened 2019-01-08
[fix][Undertow] Allow path parameters with encoded slashes to be decoded

carterkozak:
Servers may configure UndertowOptions.DECODE_URL to false in order
to support path parameters with query parameters.

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
@@ -53,6 +53,7 @@ import com.palantir.ri.ResourceIdentifier;
 import com.palantir.tokens.auth.AuthHeader;
 import io.undertow.Handlers;
 import io.undertow.Undertow;
+import io.undertow.UndertowOptions;
 import java.io.DataOutputStream;
 import java.io.File;
 import java.io.IOException;
@@ -127,6 +128,7 @@ public final class UndertowServiceEteTest extends TestBase {
                 EteBinaryServiceEndpoint.of(new UndertowBinaryResource()));
         endpoints.forEach(endpoint -> endpoint.create(context).register(handler));
         server = Undertow.builder()
+                .setServerOption(UndertowOptions.DECODE_URL, false)
                 .addHttpListener(8080, "0.0.0.0")
                 .setHandler(Handlers.path().addPrefixPath("/test-example/api", handler))
                 .build();
@@ -327,6 +329,12 @@ public final class UndertowServiceEteTest extends TestBase {
                 RequestBody.create(MediaType.parse("application/unsupported"), new byte[] {1, 2, 3})).execute())
                 .isInstanceOf(RemoteException.class)
                 .hasMessageContaining("INVALID_ARGUMENT");
+    }
+
+    @Test
+    public void testSlashesInPathParam() {
+        String expected = "foo/bar/baz/%2F";
+        assertThat(client.path(AuthHeader.valueOf("bearer"), expected)).isEqualTo(expected);
     }
 
     @Test
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
@@ -37,6 +37,8 @@ public final class ConjureHandler implements HttpHandler, RoutingRegistry {
 
     private static final ImmutableList<BiFunction<String, HttpHandler, HttpHandler>> WRAPPERS =
             ImmutableList.<BiFunction<String, HttpHandler, HttpHandler>>of(
+            // Allow the server to configure UndertowOptions.DECODE_URL = false to allow slashes in parameters.
+            // Servers which do not configure DECODE_URL will still work properly except for encoded slash values.
             (endpoint, handler) -> new URLDecodingHandler(handler, "UTF-8"),
             (endpoint, handler) -> new PathParamDecodingHandler(handler),
             // It is vitally important to never run blocking operations on the initial IO thread otherwise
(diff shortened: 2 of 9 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -127,6 +128,7 @@ public static void before() {
                 EteBinaryServiceEndpoint.of(new UndertowBinaryResource()));
         endpoints.forEach(endpoint -> endpoint.create(context).register(handler));
         server = Undertow.builder()
+                .setServerOption(UndertowOptions.DECODE_URL, false)
```
uschi2000, line 131: can we make it so that we get the correct behavior by default?

carterkozak: No, undertow decodes URLs internally prior to routing, but excludes slashes (`%2F`) to avoid request smuggling vulnerabilities. We have to ask undertow not to decode until after we've routed requests.

uschi2000: Let me try again :) What I would like is the a conjure-undertow-based Witchcraft service does the expected thing, i.e., a request to `GET /foo%2Fbar` for a template `/{var}` gives me `var=="foo/bar"`. Surely we can do this without asking Undertow to change?

carterkozak: Of course, we'll make a slight change to how we configure undertow (Add `.setServerOption(UndertowOptions.DECODE_URL, false)` and a URLDecodingHandler), and everything will work fine. Is that what you're asking?

This conjure-undertow-runtime change allows us to make that witchcraft change, resulting in the correct behavior.

carterkozak: I have an idea for an alternative implementation of this change which will likely be less invasive

uschi2000: ack

carterkozak: My quick attempt at an alternative fell through, it will be a fair amount of work, I think this PR provides what we need.

carterkozak: @uschi2000 Any concerns about this PR in its current state?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -36,6 +37,10 @@
 
     private static final ImmutableList<BiFunction<String, HttpHandler, HttpHandler>> WRAPPERS =
             ImmutableList.<BiFunction<String, HttpHandler, HttpHandler>>of(
+            // Allow the server to configure UndertowOptions.DECODE_URL = false to allow slashes in parameters.
+            // Servers which do not configure DECODE_URL will still work properly except for encoded slash values.
+            (endpoint, handler) -> new URLDecodingHandler(handler, "UTF-8"),
```
uschi2000, line 42: the javadoc of URLDecodingHandler says:

```
 * If you are using this handler you must set the {@link io.undertow.UndertowOptions#DECODE_URL} parameter to false.
```
, we'll do that in the server, ja?

carterkozak: That is correct, however the comment isn't quite right because this handler will safely no-op to the next handler in the chain if DECODE_URL is not configured.

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -329,6 +331,12 @@ public void testUnknownContentType() {
                 .hasMessageContaining("INVALID_ARGUMENT");
     }
 
+    @Test
+    public void testSlashesInPathParam() {
+        String expected = "foo/bar/baz/%2F";
+        assertThat(client.path(AuthHeader.valueOf("bearer"), expected)).isEqualTo(expected);
```
uschi2000, line 337: to test my understanding: prior to this change, this request would not have been matched by the router because undertow would first urldecode the `foo%2Fbar` portion to `foo/bar` and feed that string into the router?

carterkozak: Almost, Undertow would decode the URI, but explicitly avoid decoding `%2F` to `/`, and pass `foo%2Fbar` to our handler

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
uschi2000: looks ok. the undertow/jersey behavior already treats path parameters correctly, ja?

uschi2000 (approved):

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
carterkozak: That is correct

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
uschi2000: ok. so both of these two handlers would be no-op in that case. can you clarify this in the comment?

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
uschi2000: ok

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
carterkozak: ack

PR #164, opened 2019-01-08
[undertow] Wrappable RoutingRegistry and Endpoint

rubenfiszel:
## Before this PR

One could not simply add an HandlerWrapper to an endpoint.

## After this PR

One can use a `WrappedEndpoint` to easily wrap all handlers of an endpoint.

### DIFF
--- a/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointDetails.java
+++ b/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointDetails.java
@@ -28,6 +28,6 @@ public interface EndpointDetails {
     String template();
 
     static EndpointDetails of(HttpString method, String template) {
-        return ImmutableEndpointDetails.builder().build();
+        return ImmutableEndpointDetails.builder().method(method).template(template).build();
     }
 }
--- a/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointDetails.java
+++ b/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointDetails.java
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointDetails.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.undertow.lib;
+
+import com.palantir.tokens.auth.ImmutablesStyle;
```
carterkozak, line 19: Why are we using style from `com.palantir.tokens.auth` here?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointDetails.java
```diff
@@ -0,0 +1,33 @@
...
+import io.undertow.util.HttpString;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@ImmutablesStyle
+public interface EndpointDetails {
```
carterkozak, line 25: I'd prefer a concrete class here, no reason to allow folks to subclass

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/WrappedEndpoint.java
```diff
@@ -0,0 +1,52 @@
...
+import io.undertow.server.HandlerWrapper;
+import io.undertow.server.HttpHandler;
+import java.util.function.BiFunction;
+
+/**
+ * Wraps all the {@link HttpHandler} of an {@link Endpoint} with an HandlerWrapper defined as a function
+ * that takes an {@link EndpointDetails} and a {@link HttpHandler} and return the wrapped (or not) {@link HttpHandler}.
+ * Useful when you need to add a custom wrapper to an individual endpoint.
+ */
+public final class WrappedEndpoint implements Endpoint {
```
carterkozak, line 28: I'd prefer to hold off making this sort of utility public until the system is a bit more stable

rubenfiszel: If we don't make this public, then there isn't much sense to this PR since it's not used internally either. Would it be enough to make it `@Beta` ?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointDetails.java
rubenfiszel: Sure, but what is the risk of people subclassing it ? This is easier to mock this way.

PR #167, opened 2019-01-09
Unpin com.palantir.conjure.java.runtime:conjure-java-jaxrs-client

uschi2000:
The bug/regression that required this pin was fixed in
https://github.com/palantir/conjure-java-runtime/pull/865


uschi2000:
nvm, I misread.

PR #168, opened 2019-01-09
 Proposal: ParametersAreNonnullByDefault 

carterkozak:
Annotate generated types with `@ParametersAreNonnullByDefault` for more helpful inspections.

Intellij Idea will get upset if a null value is passed to a service annotated with `ParametersAreNonnullByDefault`, which isn't the case using our opt-in `javax.validation.constraints.NotNull` annotations.

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceGeneratorTests.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceGeneratorTests.java
@@ -63,7 +63,8 @@ public final class JerseyServiceGeneratorTests extends TestBase {
 
         for (Path file : files) {
             if (Boolean.valueOf(System.getProperty("recreate", "false"))) {
-                Path output = Paths.get("src/test/resources/test/api/" + file.getFileName() + ".jersey");
+                Path output = Paths.get("src/test/resources/test/api/"
+                        + file.getFileName() + ".jersey_require_not_null");
                 Files.delete(output);
                 Files.copy(file, output);
             }
--- a/conjure-java-core/src/integrationInput/java/com/palantir/another/ConjureErrors.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/another/ConjureErrors.java
@@ -3,8 +3,10 @@ package com.palantir.another;
 import com.palantir.conjure.java.api.errors.ErrorType;
 import com.palantir.conjure.java.api.errors.ServiceException;
 import javax.annotation.Generated;
+import javax.annotation.ParametersAreNonnullByDefault;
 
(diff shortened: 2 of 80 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceGeneratorTests.java
```diff
@@ -63,7 +63,8 @@ public void testServiceGeneration_exampleService_requireNotNullAuthHeadersAndReq
 
         for (Path file : files) {
             if (Boolean.valueOf(System.getProperty("recreate", "false"))) {
-                Path output = Paths.get("src/test/resources/test/api/" + file.getFileName() + ".jersey");
+                Path output = Paths.get("src/test/resources/test/api/"
+                        + file.getFileName() + ".jersey_require_not_null");
```
dansanduleac, line 67: well spotted inconsistency!

carterkozak: I'll pull this out into a separate PR, it's not related to this feature, it just allowed `-Drecreate=true` not cause tests to pass

dansanduleac:
Instead of annotating every single class from this repo with this, why don't we annotate the package instead in `package-info.java`?

carterkozak:
Our generated code will end up in packages that may be used by other non-generated code. We shouldn't apply this check to things we haven't generated.

conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceGeneratorTests.java
carterkozak: PR is up: https://github.com/palantir/conjure-java/pull/169

dansanduleac:
I meant that only for packages belonging to conjure-java-core, which we do control.
Edit: nvm, actually all of the touched files were generated, missed that the first time.

carterkozak:
I didn't annotate anything in conjure-java-core, only classes output from our generators (which take up most of the diff).

dansanduleac:
👍

iamdanfox:
Given that the `@ParametersAreNonnullByDefault` won't actually fail any builds, is this just for the benefit of people who diligently resolve all their IntelliJ warnings?

From a purely cosmetic perspective it just looks a bit gross (e.g. on plain Java Beans where there are no parameters at all).  I wonder if we should be pursuing stricter static analysis like NullAway again?

carterkozak:
That works for me given error-prone doesn't support this annotation. We found this helpful on another project, but it didn't have much in the way of static analysis at the time.

PR #169, opened 2019-01-09
Fix inconsistency in JerseyServiceGeneratorTests

carterkozak:
recreate=true will recreate a correctly suffixed output file.


dansanduleac:
better yet, why don't we pull this out into a variable so someone else can't make the same mistake in the future?

carterkozak:
> better yet, why don't we pull this out into a variable so someone else can't make the same mistake in the future?

That's fair, we had a ton of duplicated code. I factored out the validation.

dansanduleac:
👍

PR #177, opened 2019-01-11
[fix] Undertow Endpoints do not use form encoding in url paths

carterkozak:
This fixes a compatibility issue with retrofit clients, but feign was never affected


uschi2000 (approved):

uschi2000:
does retrofit send form-encoded data? that seems wrong.

carterkozak:
Retrofit allows literal `+` characters in the request path (which is allowed, they only need to be encoded in query parameters) but feign encodes them in both sections which is also allowed afaict.

PR #178, opened 2019-01-11
[improvement][undertow] safe-logging npe when a null BinaryResponseBody is returned

carterkozak:


iamdanfox (approved):
Always a fan of banning more nulls

PR #181, opened 2019-01-12
[Undertow][WIP] Per-endpoint metrics

rubenfiszel:
@cakofony This should be enough to add tracing and per-endpoint metrics in wc. What do you think?


conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoint.java
```diff
@@ -46,7 +47,12 @@ public void register(RoutingRegistry routingRegistry) {
         private class EmptyPathHandler implements HttpHandler {
             @Override
             public void handleRequest(HttpServerExchange exchange) throws IOException {
+                Long start = System.currentTimeMillis();
                 boolean result = delegate.emptyPath();
+                Long duration = System.currentTimeMillis() - start;
```
carterkozak: should be unboxed `long`.

I think I may have accidentally mislead you earlier, timing for per-endpoint metrics should probably be the full request duration, not time spent inside the resource method, so we should be able to safely get rid of timing code here.

rubenfiszel:
Currently you can get both and I think it would be valuable to get both
(look at the new handler)

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoint.java
```diff
@@ -46,7 +47,12 @@ public void register(RoutingRegistry routingRegistry) {
...
+                exchange.putAttachment(Metrics.DELEGATE_DURATION_KEY, duration);
+                exchange.putAttachment(Metrics.SERVICE_NAME_KEY, "EmptyPathService");
+                exchange.putAttachment(Metrics.RESOURCE_METHOD_NAME_KEY, "emptyPath");
```
carterkozak: At the moment our generated <endpoint_name>Handler classes are only responsible for the bare minimum to read inputs, delegate to the service, and respond. Can we move these attachments to a separate handler like we do with the tracing handler?

carterkozak: If we could provide service name and resource method to the RoutingRegistry we would be free to implement the handler in the runtime module

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WorkerStartTimeHandler.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private final HttpHandler next;
+
+    WorkerStartTimeHandler(HttpHandler next) {
+        this.next = next;
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        exchange.putAttachment(WORKER_START_TIME_KEY, System.currentTimeMillis());
```
carterkozak: The exchange already provides the start time. Since we're not consuming this value, I think this functionality is a better fit for individual webservers which produce full metric data. We already have a MetricsHandler in WC which duplicates this functionality.

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoint.java
rubenfiszel: Yes, we could do that and add that to the RoutingRegistry. Will adapt the PR.
I still think it would be valuable to get the duration of the delegate method in addition to the full request duration. For profiling of your resource code, you're probably interested in that value and not the rest. So I would be in favor to keep the duration attachment.

carterkozak: I think we already get that data if the resource itself has been instrumented with Tritium, I'd prefer to focus on providing a replacement for the jersey metrics functionality here, and add additional features separately if that's alright with you.

rubenfiszel: But tritium does it using reflection and proxying. It is also a pretty big dependency we could remove on the conjure part. Imho, It's a bit sad to not take advantage that using codegen we can avoid that, but we can always add it later.

carterkozak: Sure, I don't think the service method invocations should necessarily be tied to undertow handler, we don't get leverage out of attaching invocation data to the HttpServerExchange. We could instead generate `InstrumentedFooService` as a concrete implementation which delegates to the service interface implementation. This way when services are called internally from code (as opposed to from top level http requests) we can still record metrics.

rubenfiszel: That's an interesting idea:

Is this what you have in mind ?

```
//all generated from conjure service definition
interface FooService {
    Bar bar(...);
}

abstract class InstrumentedFooService implements FooService {
  abstract public Bar instrumentedBar(...);

  public final bar(...) {
    instrument(instrumentedBar(...));
  }
}
```

rubenfiszel: That's an interesting idea:

Is this what you have in mind ?

```
//all below is generated from conjure service definition
interface FooService {
    Bar bar(...);
}

public final class InstrumentedFooService implements FooService {
  FooService fooService;
  ...

  of(FooService fooService) {...} 

  public final Bar bar(...) {
     Instruments instruments = Instruments.start("FooService", "bar", ...)
     Bar bar = fooService.bar(...);
     instruments.complete();
     return bar;
  }
}
```

carterkozak: Yep

rubenfiszel: That makes a lot of sense. I guess we should do that in a separate PR once and if we want to remove the need for tritium.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RoutingRegistry.java
```diff
@@ -27,11 +27,11 @@
...
+    RoutingRegistry get(String template, String serviceName, String endpointName, HttpHandler handler);
 
-    RoutingRegistry post(String template, HttpHandler handler);
+    RoutingRegistry post(String template, String serviceName, String endpointName, HttpHandler handler);
 
-    RoutingRegistry put(String template, HttpHandler handler);
+    RoutingRegistry put(String template, String serviceName, String endpointName, HttpHandler handler);
 
-    RoutingRegistry delete(String template, HttpHandler handler);
+    RoutingRegistry delete(String template, String serviceName, String endpointName, HttpHandler handler);
```
carterkozak, line 36: This breaks the audit library -- is that okay?

If we're going to break API I'd prefer to simplify RoutingRegistry to something along these lines to allow extensibility in the future:

```java
interface RoutingRegistry {
  RoutingRegistry register(EndpointRegistration);
}

final class EndpointRegistration {
   HttpHandler handler;
   pathTemplate; // e.g. /foo/{bar}
   requestMethod // e.g. GET
   optional<endpointName>
   optional<serviceName>
  ..etc..
}
```

carterkozak:
Alternatively the easier solution is to implement a ServiceNameHandler in lib.internal with sets the attachments before delegating to our handler, and we wrap the MethodNameHandler with that in generated code.

carterkozak:
@rubenfiszel Can this be closed?

PR #184, opened 2019-01-14
Add experimental DialogueClientGenerator

uschi2000:


uschi2000:
This is copy&pasted from the internal repo. I will port over the conjure-verifier tests in a separate PR.

uschi2000:
@iamdanfox @rubenfiszel

uschi2000:
ready for review

conjure-java-core/build.gradle
```diff
@@ -38,13 +38,15 @@ dependencies {
     compile 'com.google.googlejavaformat:google-java-format'
     compile 'com.google.guava:guava'
     compile 'com.palantir.conjure.java.api:errors'
+    compile 'com.palantir.dialogue:dialogue-client'
```
carterkozak, line 41: Do we want to make this repo public first?

carterkozak:
Needs a section in the readme

conjure-java-core/src/test/resources/test/api/DialogueDialogueSampleService.java.dialogue
```diff
@@ -0,0 +1,522 @@
...
+                                ImmutableList.of(
+                                        PathTemplate.Segment.fixed("base"),
+                                        PathTemplate.Segment.fixed("binaryEcho")));
+
+                private final Serializer<String> serializer =
+                        Serializers.jackson(
+                                "binaryEcho",
+                                com.palantir.product.DialogueDialogueSampleService.mapper);;
+
+                private final Deserializer<InputStream> deserializer = Deserializers.passthrough();;
```
carterkozak: double `;` on this line and above

conjure-java-core/build.gradle
```diff
@@ -38,13 +38,15 @@ dependencies {
...
     compile 'com.palantir.ri:resource-identifier'
     compile 'com.palantir.syntactic-paths:syntactic-paths'
     compile 'com.palantir.tokens:auth-tokens'
     compile 'com.squareup:javapoet'
     compile 'javax.annotation:javax.annotation-api'
     compile 'javax.validation:validation-api'
     compile 'org.apache.commons:commons-lang3'
+    compile 'org.glassfish.jersey.core:jersey-common' // TODO(rfink): only here for UriTemplateParser. Re-implement?
```
carterkozak, line 49: would prefer to do this before merging

conjure-java-core/build.gradle
uschi2000: done

conjure-java-core/src/test/resources/test/api/DialogueDialogueSampleService.java.dialogue
uschi2000: done (and in other places)

conjure-java-core/build.gradle
uschi2000: will do

conjure-java-core/src/integrationInput/java/com/palantir/product/DialogueDialogueSampleService.java
```diff
@@ -0,0 +1,522 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.conjure.java.serialization.ObjectMappers;
+import com.palantir.dialogue.Call;
+import com.palantir.dialogue.Calls;
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Deserializer;
+import com.palantir.dialogue.Deserializers;
+import com.palantir.dialogue.DialogueOkHttpErrorDecoder;
+import com.palantir.dialogue.Endpoint;
```
rubenfiszel, line 15: off-topic: I see `Endpoint`  to designate a single route. Same for conjure (under `EndpointName`). Whereas in undertow-lib, we use it to designate something that is gonna register many different routes. I think this is a bit confusing. Could we potentially rename `Endpoint` to `Endpoints` in undertow-lib @cakofony ?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/ClassVisitor.java
```diff
@@ -0,0 +1,89 @@
...
+import com.palantir.conjure.spec.MapType;
+import com.palantir.conjure.spec.OptionalType;
+import com.palantir.conjure.spec.PrimitiveType;
+import com.palantir.conjure.spec.SetType;
+import com.palantir.conjure.spec.TypeDefinition;
+import com.squareup.javapoet.TypeName;
+import java.io.InputStream;
+import java.util.List;
+
+final class ClassVisitor implements ClassNameVisitor {
```
rubenfiszel: name seems a bit off + potentially a short line of docs ?

Also @iamdanfox I find it confusing that ClassNameVisitor is a default interface and DefaultClassNameVisitor is a final class. Could we not merge ClassNameVisitor and DefaultClassNameVisitor into a single default interface and have this extends/implents it ?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/ClassVisitor.java
```diff
@@ -0,0 +1,89 @@
...
+
+    enum Mode {
```
rubenfiszel: Might be worth considering using a boolean instead of an enum is there are only two values and only one we actually use. The other one exists only as the complementary.
`isReturnValue` ?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueClientGenerator.java
```diff
@@ -0,0 +1,586 @@
...
+                .collect(Collectors.toSet());
+    }
+
+    // TODO(rfink): Split into separate classes: endpoint, interface, impl.
+    private static final class TypeAwareGenerator {
+        private static final String AUTH_HEADER_NAME = "Authorization";
+        private static final String AUTH_HEADER_PARAM_NAME = "authHeader";
+        private static final String MAPPER_FIELD_NAME = "mapper";
+        private static final String REQUEST_VAR_NAME = "_request";
+        private static final String REQUEST_VAR_NAME_BUILDER = "_request_builder";
```
rubenfiszel: curious to know what is the convention in such case. Does the prefix `_` forces us to avoid caml case or is it also acceptable to use "_requestBuilder"

conjure-java-core/build.gradle
uschi2000: Second thought: if I touch this then I would want to implement the "Unique Conjure API endpoint paths" proposal (will send internal link). That's a somewhat bigger project so I'd prefer to merge this as is and then re-visit the path/routing implementation more holistically later.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueClientGenerator.java
```diff
@@ -0,0 +1,586 @@
...
+                new ClassVisitor(conjureDefinition.getTypes(), ClassVisitor.Mode.PARAMETER));
+        TypeMapper returnTypes = new TypeMapper(conjureDefinition.getTypes(),
+                new ClassVisitor(conjureDefinition.getTypes(), ClassVisitor.Mode.RETURN_VALUE));
+        TypeAwareGenerator generator = new TypeAwareGenerator(parameterTypes, returnTypes);
+        return conjureDefinition.getServices().stream()
+                .flatMap(serviceDef -> generator.service(serviceDef).stream())
+                .collect(Collectors.toSet());
+    }
+
+    // TODO(rfink): Split into separate classes: endpoint, interface, impl.
```
rubenfiszel, line 114: Are we still interested to do that ? later PR ?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueClientGenerator.java
```diff
@@ -0,0 +1,586 @@
...
+            ClassName className = serviceClassName("Dialogue", def);
+            TypeSpec.Builder serviceBuilder = TypeSpec.classBuilder(className)
+                    .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
+                    .addAnnotation(ConjureAnnotations.getConjureGeneratedAnnotation(DialogueClientGenerator.class));
+
+            serviceBuilder.addMethod(MethodSpec.constructorBuilder().addModifiers(Modifier.PRIVATE).build());
+            serviceBuilder.addMethod(blockingClient(def));
+            serviceBuilder.addField(FieldSpec.builder(ObjectMapper.class, "mapper")
+                    .initializer("$T.newClientObjectMapper()",
+                            // TODO(rfink): Stop relying on http-remoting.
```
rubenfiszel: This can be removed I believe?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/ClassVisitor.java
uschi2000: done

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/ClassVisitor.java
uschi2000: I think it's easier to read as is

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueClientGenerator.java
uschi2000: should be _requestBuilder, thanks

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueClientGenerator.java
uschi2000: yep, later pr

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueClientGenerator.java
uschi2000: done, thanks

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueClientGenerator.java
```diff
@@ -0,0 +1,586 @@
...
+
+            return JavaFile
+                    .builder(def.getServiceName().getPackage(), serviceBuilder.build())
+                    .skipJavaLangImports(true)
+                    .build();
+        }
+
+        private MethodSpec blockingApiMethod(EndpointDefinition endpointDef) {
+            MethodSpec.Builder methodBuilder = MethodSpec.methodBuilder(endpointDef.getEndpointName().get())
+                    .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
```
rubenfiszel, line 154: curious: are `PUBLIC` `ABSTRACT` necessary since they are the "methods" of an interface (and obv javapoet knows this since it doesn't render those modifiers). What happens if you remove those modifiers ?

uschi2000: `java.lang.IllegalArgumentException: modifiers [abstract] must contain one of [public, private]`

rubenfiszel: Found that: https://github.com/square/javapoet#interfaces

Makes sense. Thanks!

carterkozak:
I don't want to block this merge on nits, we can make smaller changes in follow-ups.
👍

uschi2000:
@iamdanfox do you want to take a look?

iamdanfox:
@uschi2000 I have a couple of concerns:

1) I don't think dialogue's design is aligned with the best practises in RFC003: https://palantir.github.io/conjure/#/docs/rfc/003-loosely-coupled-conjure-clients.

Specifically, the generated `DialogueDialogueSampleService.java` is deeply coupled to the `dialogue-client` classes:

```java
import com.palantir.dialogue.Call;
import com.palantir.dialogue.Calls;
import com.palantir.dialogue.Channel;
import com.palantir.dialogue.Deserializer;
import com.palantir.dialogue.Deserializers;
import com.palantir.dialogue.DialogueOkHttpErrorDecoder;
import com.palantir.dialogue.Endpoint;
import com.palantir.dialogue.Exceptions;
import com.palantir.dialogue.HttpMethod;
import com.palantir.dialogue.OkHttpErrorDecoder;
import com.palantir.dialogue.PathTemplate;
import com.palantir.dialogue.Request;
import com.palantir.dialogue.Serializer;
import com.palantir.dialogue.Serializers;
```

(This is a lot more than a regular jax-rs service which essentially just imports a few jax-rs annotations to mark the `@Path`, `@GET`, `@POST` etc).

This structure is alarming to me because if a service wants to talk to 5 different Dialogue Conjure APIs, then there will need to be a version of the [dialogue-client jar](https://github.com/palantir/dialogue/tree/develop/dialogue-client/src/main/java/com/palantir/dialogue) which is compatible with 5 different generated dialogue clients, even if they were generated by different versions of the dialogue generator! This wasn't too scary when the shared 'contract' was jaxrs (because this is an incredibly stable open-source thing), but it's scarier to me when we've invented dialogue ourselves and not proven it anywhere.

2) My other concern is that I'm still not convinced that the conjure-java repo is the right home for lots of experimental generators.  Every experimental thing we add to conjure-java dilutes releases and makes it less clear to consumers when a minor bump to 'conjure-java' will actually affect them.  It also increases the temptation to mis-label breaking things in non-breaking conjure-java releases to avoid impacting lots of unrelated users.

3) We have the conjure-verification harness, why not use it?

4) this is a 2000 line PR without any description of motivation, design principles etc and the README only has a couple of sentences. I think PR descriptions are especially valuable for posterity when we want to see why decisions were made at the time.

Possible solutions here would be to experiment more on a fork of conjure-java (which users could still consume if you set up publishing for it), or push more aggressively for a stable conjure-java API.

uschi2000:
ACK on the README and rationale docs, will work on something. conjure-verifier tests exist in the internal repo, but I left them out to keep the PR manageable; will add them as a follow-up.

Re library coupling: dialogue is the generator target and as such has binary coupling with consumers via the generator. Like in http-remoting and other shared libraries, this is mitigated by careful back compat consideration plus package and artifact versioning. I can also try to split out a dialogue-generator-api from dialogue-clients, but there will always be some binary coupling.

Re experimental/fork: I believe you committed to host experimental generators in this repo when we discussed whether or not to publish conjure-java-core in order to facilitate external generators. We have plenty of organizational evidence that forks are not an effective collaboration mechanism.

uschi2000:
double backflip, going with separate repo after all.

PR #186, opened 2019-01-14
Remove TraceHandler iothread check

carterkozak:
We do not allow these handlers to run on io threads, there's no
reason to push this validation to runtime. Furthermore there's
no reason not to allow tracing on handlers which run asynchronously.


uschi2000:
Defer to you. I included this as a safety guard because I didn't fully understand the threading model and wanted to be safe rather than sorry.

uschi2000 (approved):

carterkozak:
I think it was the right thing to do when we implemented this because TraceHandler was public at the time. Now that we protect configuration behind ConjureHandler we're safer against misuse.

PR #190, opened 2019-01-16
[improvement] add more efficient Bytes constructors

berler:
Adds methods to construct Bytes from InputStreams and
ByteArrayOutputStreams to avoid extra copying.

## Before this PR
Code that constructs Bytes as part of return values often results in a lot of extra copying, when the underlying data already came from an InputStream or a ByteArrayOutputStream.

## After this PR
Using the new Bytes#from methods with an InputStream or ByteArrayOutputStream results in one less copy compared to constructing your own byte[] and using from(byte[])

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
@@ -115,7 +115,7 @@ public final class Bytes {
         return new Bytes(safe);
     }
 
-    /** Constructs a new {@link Bytes} from the provided {@link ByteArrayOutputStream} */
+    /** Constructs a new {@link Bytes} from the provided {@link ByteArrayOutputStream}. */
     public static Bytes from(ByteArrayOutputStream byteStream) {
         // this returns a new copy already
         byte[] safe = byteStream.toByteArray();
@@ -123,7 +123,7 @@ public final class Bytes {
         return new Bytes(safe);
     }
 
-    /** Constructs a new {@link Bytes} by reading the provided {@link InputStream} fully */
+    /** Constructs a new {@link Bytes} by reading the provided {@link InputStream} fully. */
     public static Bytes from(InputStream inputStream) throws IOException {
         ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
         copy(inputStream, outputStream);
@@ -131,13 +131,13 @@ public final class Bytes {
         return from(outputStream);
     }
 
-    // more or less what IOUtils would do
+    // more or less what IOUtils would do. In Java9, you could do inputStream.transferTo(outputStream)
     private static void copy(InputStream inputStream, OutputStream outputStream) throws IOException {
         byte[] buf = new byte[4096];
-        int r;
+        int read;
 
-        while ((r = inputStream.read(buf)) != -1) {
-            outputStream.write(buf, 0, r);
+        while ((read = inputStream.read(buf)) != -1) {
+            outputStream.write(buf, 0, read);
         }
     }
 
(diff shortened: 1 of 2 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
```diff
@@ -24,7 +24,9 @@
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.google.common.io.ByteStreams;
```
carterkozak: I'm surprised that guava is on the classpath for conjure-lib

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
```diff
@@ -113,6 +115,21 @@ public static Bytes from(ByteBuffer buffer) {
         return new Bytes(safe);
     }
 
+    /** Constructs a new {@link Bytes} from the provided {@link ByteArrayOutputStream} */
+    public static Bytes from(ByteArrayOutputStream byteStream) {
```
carterkozak, line 119: How common is this operation?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
berler: ah yeah, fixed. i'm not sure why my IDE thought it was

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
berler: in our internal product we have quite a few places that would use this, and I think this is a lot more common than ByteBuffer

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
```diff
@@ -113,6 +115,32 @@ public static Bytes from(ByteBuffer buffer) {
...
+        // this returns a new copy already
+        byte[] safe = byteStream.toByteArray();
+
+        return new Bytes(safe);
+    }
+
+    /** Constructs a new {@link Bytes} by reading the provided {@link InputStream} fully */
+    public static Bytes from(InputStream inputStream) throws IOException {
```
carterkozak, line 127: I don't think this method is a good idea, we don't want to encourage fully buffering InputStreams on heap. I realize sometimes we end up with code that does this, and the implementation you have provided does it better, but I'm worried that by adding this method, people will use it instead of plumbing more efficient streams end to end.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
carterkozak: I can extend `ByteArrayOutputStream` overriding `toByteArray` in a way that breaks the immutability of this class.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
berler: I guess that is fair. If we keep the ByteArrayOutputStream version then I imagine others would duplicate this though.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
berler: hmm. I didn't think about that. I suppose if you want to guarantee safety you could make an additional copy if the runtime class is a subclass of ByteArrayOutputStream. this might be too complicated now though

berler:
abandoning in light of the issues pointed out by @cakofony

PR #196, opened 2019-01-21
[improvement][undertow] Remove unnecessary serializer memoized suppliers 

carterkozak:
We create a single SerializerRegistry per application, there is no reason to memoize our serializers.


rubenfiszel (commented):
No more singleton?

carterkozak:
Nope, we don't create these frequently enough for it to matter. Originally I think we passed a serializer to each generated handler, but now we use the SerializerRegistry per application.

rubenfiszel (approved):

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Serializers.java
```diff
@@ -69,54 +67,48 @@ public void serialize(Object value, OutputStream output) throws IOException {
         }
     }
 
-    private static final Supplier<Serializer> jsonInstance = Suppliers.memoize(() ->
```
carterkozak, line 72: No code change aside from removing the memoized supplier.

iamdanfox (approved):

PR #201, opened 2019-01-22
[fix] optional<binary> equivalent to binary for retrofit + Optional<StreamingOutput> for jersey

### LINKED ISSUE #182: Retrofit should handle optional<binary> responses
## What happened?

Retrofit fails to decode optional<binary> because it expects JSON and receives `application/octet-stream` data.

Retrofit should allow the consumer to stream a response if present.

rubenfiszel:
Fix #182 

For jersey and retrofit, the current behavior of `binary` enables an implicit tri-state with different status code:
- non-empty stream: 200
- empty stream: 200
- null: 204 + null stream on retrofit client

The way forward to achieve a standard behavior with all clients and server is to consider that service that take advantage of this behavior should declare a return type of `optional<binary>` instead. The jersey resource implementation will need to stay the same and keep encoding empty as null.

In this PR, we are making the return type `optional<binary>` being equivalent to `binary` for the retrofit generator and change the jersey generator to return `Optional<StreamingOutput>`.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
@@ -53,6 +53,7 @@ import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.JavaFile;
 import com.squareup.javapoet.MethodSpec;
 import com.squareup.javapoet.ParameterSpec;
+import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
 import java.io.InputStream;
@@ -97,9 +98,16 @@ public final class JerseyServiceGenerator implements ServiceGenerator {
                 ? BINARY_RETURN_TYPE_RESPONSE
                 : BINARY_RETURN_TYPE_OUTPUT;
 
+        TypeName optionalBinaryReturnType = experimentalFeatures.contains(FeatureFlags.JerseyBinaryAsResponse)
+                ? BINARY_RETURN_TYPE_RESPONSE
+                : ParameterizedTypeName.get(ClassName.get(Optional.class), binaryReturnType);
+
         TypeMapper returnTypeMapper = new TypeMapper(
                 conjureDefinition.getTypes(),
-                new ReturnTypeClassNameVisitor(conjureDefinition.getTypes(), binaryReturnType));
+                new ReturnTypeClassNameVisitor(
+                        conjureDefinition.getTypes(),
+                        binaryReturnType,
+                        optionalBinaryReturnType));
 
         TypeMapper argumentTypeMapper = new TypeMapper(
                 conjureDefinition.getTypes(),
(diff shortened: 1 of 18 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/ReturnTypeClassNameVisitor.java
```diff
@@ -58,6 +59,10 @@ public TypeName visitMap(MapType type) {
 
     @Override
     public TypeName visitOptional(OptionalType type) {
+        if (type.getItemType().accept(TypeVisitor.IS_PRIMITIVE)
+                && type.getItemType().accept(TypeVisitor.PRIMITIVE).equals(PrimitiveType.BINARY)) {
+            return delegate.visitPrimitive(PrimitiveType.BINARY);
```
carterkozak: I think we want this.visit, not delegate, since this implementation provides return-type specific binary classes.

rubenfiszel: fixed

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -169,9 +170,13 @@ private MethodSpec generateServiceMethod(
         }
 
         if (returnType.equals(BINARY_RETURN_TYPE_OUTPUT) || returnType.equals(BINARY_RETURN_TYPE_RESPONSE)) {
-            methodBuilder.addAnnotation(AnnotationSpec.builder(ClassName.get("javax.ws.rs", "Produces"))
-                    .addMember("value", "$T.APPLICATION_OCTET_STREAM", ClassName.get("javax.ws.rs.core", "MediaType"))
-                    .build());
+            methodBuilder
```
carterkozak: Nit: Would be nice to comment that jersey cannot support `Optional<StreamingOutput>`

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -169,9 +170,13 @@ private MethodSpec generateServiceMethod(
...
+                    .addAnnotation(Nullable.class)
```
carterkozak: Can we avoid adding `Nullable` to non-optional returns?

rubenfiszel: They are nullable technically tho. But being an annotation I guess it's fine.

carterkozak: Returning null for those endpoints violates the conjure defined API

carterkozak (dismissed):

carterkozak (changes requested):
Based on https://github.com/palantir/conjure-java-runtime/pull/931 I think we can return optional<StreamingOutput>

carterkozak (approved):

PR #204, opened 2019-01-23
Set the correct content-type for jersey optional binary responses

carterkozak:
Updates the dropwizard EteTestServer to use the correct objectmapper
and register ConjureJerseyFeature in order to add tests.


rubenfiszel (approved):
Good catch

carterkozak:
Thanks 👍 debugging the dropwizard nondeterministic-404-on-empty while trying to add tests for my fix wasn't fun!

iamdanfox (approved):

PR #213, opened 2019-01-26
Add the 'Undertow: ' prefix to conjure-undertow tracing spans

carterkozak:
For additional context see:
https://github.com/palantir/tracing-java/pull/58


rubenfiszel:
Typo in PR name.
If we go that route, we might me interested to also insert the Server library version

carterkozak:
Woops, fixed the typo. I want to keep parity with tracing-java for the time being

markelliot:
It would be slightly better decomp if this tracing handler lived with all the other tracing handlers so that we could keep the tracing concern in sync.

carterkozak:
Sure, now that this generator is stable we can split out the TraceHandler into the tracing project.

carterkozak:
https://github.com/palantir/tracing-java/pull/60

PR #219, opened 2019-02-01
[fix] ConjureExceptionHandler understands QosExceptions

iamdanfox:
## Before this PR

There are three kinds of QosException in conjure-java-runtime-api, but the ConjureExceptionHandler currently sends them all over the wire as 500 errors.

(See the full list of available [exceptions here](https://github.com/palantir/conjure-java-runtime-api/tree/develop/errors/src/main/java/com/palantir/conjure/java/api/errors))

## After this PR

The Undertow behaviour matches the old Jersey behaviour ([QosExceptionMapper](https://github.com/palantir/conjure-java-runtime/blob/develop/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/QosExceptionMapper.java))

### DIFF
--- a/conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
+++ b/conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
@@ -18,7 +18,9 @@ package com.palantir.conjure.java.undertow.runtime;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.google.common.collect.ImmutableList;
 import com.palantir.conjure.java.api.errors.ErrorType;
+import com.palantir.conjure.java.api.errors.QosException;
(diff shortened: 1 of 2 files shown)


conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
```diff
@@ -90,6 +96,48 @@ public void handlesRemoteException() throws IOException {
         assertThat(response.code()).isEqualTo(ErrorType.INTERNAL.httpErrorCode());
     }
 
+    @Test
+    public void handlesQosExceptionThrottleWithoutDuration() throws IOException {
```
iamdanfox, line 100: These methods match the old ones in https://github.com/palantir/conjure-java-runtime/blob/develop/conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/QosExceptionMapperTest.java as much as possible

carterkozak (approved):
Thanks!

PR #222, opened 2019-02-04
Switch to gradle-consistent-versions

iamdanfox:
No longer using nebula for our dependencies, instead using: https://github.com/palantir/gradle-consistent-versions

**Use `./gradlew --write-locks` to regenerate**


dansanduleac:
👍

PR #226, opened 2019-02-06
[improvement] 'UseImmutableBytes' feature flag allows opting-in to Bytes instead of ByteBuffer

carterkozak:


rubenfiszel:
What are the pro and cons of Bytes vs ByteBuffer?

carterkozak:
Bytes is immutable (like protobuf ByteString), where ByteBuffer is stateful, the cursor position changes as the value is read.
The Bytes type was added in #139, this commit cleans up #158 so that we can roll the feature out without necessarily forcing a break.

carterkozak:
~~I missed updating the CLI to expose the new feature flag, commit incoming.~~ updated

iamdanfox:
@cakofony thanks for pushing this through - any chance we can flip the default so this isn't a break initially? The flag would be an opt-in, something like `UseImmutableBytes`, that way candidates like our evolution of Atlas can opt-in at their convenience?

carterkozak:
At what point would we change the default? Are we confident that anyone will opt into the change?
I suppose this could be excavated out?

iamdanfox:
Yeah we'd just print log lines saying you're using deprecated `ByteBuffer` codegen, please enable this feature flag to opt in to the preferred codepath - each product can then choose how and when they want to release and communicate a compilation break in their API jars.

(We'd also have it on by default in all newly generated repos)

carterkozak:
Updated, would you like me to add deprecation logging to the CLI in this PR?

carterkozak:
@iamdanfox Anything missing here?

iamdanfox (approved):
Looks great, thanks @cakofony! I've just put a quick note on the README so we can point people here

dansanduleac (approved):

iamdanfox:
Policy bot says this is approved but the light hasn't gone green... just gonna use admin perms in a sec:

screen shot 2019-02-08 at 14 03 25

PR #230, opened 2019-02-10
Fix bintray publishing

iamdanfox:
@cakofony noticed our 2.9.0 publish failed.  I SSH'd into the build and realised it was trying to create a bintray pkg called 'palantir-conjure-java', but we only have perms to write to 'conjure-java'.

This is an unintended side effect of https://github.com/palantir/conjure-java/pull/222

### DIFF
--- a/gradle/publish-jar.gradle
+++ b/gradle/publish-jar.gradle
@@ -17,7 +17,7 @@ bintray {
     publish = true
     pkg {
         repo = 'releases'
-        name = rootProject.name
+        name = 'conjure-java'
         userOrg = 'palantir'
         licenses = ['Apache-2.0']
         publications = ['nebula']
--- a/gradle/publish-dist.gradle
+++ b/gradle/publish-dist.gradle
@@ -12,7 +12,7 @@ bintray {
     publish = true
     pkg {
         repo = 'releases'
-        name = rootProject.name
+        name = 'conjure-java'
         userOrg = 'palantir'
         licenses = ['Apache-2.0']
         publications = ['dist']


gradle/publish-jar.gradle
```diff
@@ -17,7 +17,7 @@ bintray {
     publish = true
     pkg {
         repo = 'releases'
-        name = rootProject.name
+        name = 'conjure-java'
```
carterkozak, line 20: would it be cleaner to update the root project name in settings.gradle?

dansanduleac: We can't because of the [consistent-versions limitation](https://github.com/palantir/gradle-consistent-versions#known-limitation-root-project-must-have-a-unique-name) :(

dansanduleac:
@iamdanfox this is not ideal indeed as that file is templated.
Maybe we should change the tempplate to source that information from either
* github repo name (env variable)
* predefined env variable that our bintray automation could populate in circle

robert3005:
Change root project to have correct name and sub project to have a
different one?

dansanduleac:
Not so easy, since we rely on the subproject name too to determine the artifact name.

iamdanfox:
Given that internal bintray-automation gives us perms based on the repository name, I think it's safest to just have excavator embed the actual REPOSITORY_NAME in the repo.  That way whatever the rootProject.name is it should "just work".  (I've put up a PR to make excavator do this).

carterkozak:
I think this can be closed since #231 has merged

PR #232, opened 2019-02-11
Add the '--useInsensitiveEnums' flag to allow case-insensitive enum values

carterkozak:


markelliot:
should this be the default so that there's no surprise breaks?

j-baker:
why would we not want this to just be the only one?

j-baker:
thankfully it's not a backcompat break to allow more input

carterkozak:
> should this be the default so that there's no surprise breaks?

@markelliot This flag enables a feature which does not align with the specification, nor should it, the goal of conjure is to provide a strict set of standards across our ecosystem. Based on the features that have been removed despite existing usage in favor of a cleaner standard, I do not think it's reasonable to update the spec based on this detail either.

> why would we not want this to just be the only one?

@j-baker I'm sorry but I'm not sure what you're trying to say.

iamdanfox:
@j-baker you might have missed all the back and forth on this, more context on: https://github.com/palantir/conjure/issues/193

markelliot:
idk, feels like we're having a circular argument at this point. This unbreaks the thing I'd like to unbreak, and likely unbreaks a lot of services. I think the spec should be changed along with other implementations because case-insensitive interpretation and single-case production is as reasonable a semantic as case-sensitive interpretation and no less precise, while helping a number of existing clients.

When can this be merged and released?

iamdanfox (approved):
OK let's get this merged as a short term fix.  I'm still going to discourage anyone from using this flag because it reduces the guarantees we can provide about interop with other languages - would prefer repos to make a decent effort to switch to uppercase enums.

PR #233, opened 2019-02-12
[undertow] Allow instrumentation of service implementations

carterkozak:
This allows server-specific metrics to be applied to service
implementations without exposing complexity to authors of
individual services.

### DIFF
--- a/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServiceContext.java
+++ b/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServiceContext.java
@@ -25,10 +25,13 @@ import com.palantir.logsafe.Preconditions;
 public final class ServiceContext {
 
     private final SerializerRegistry serializerRegistry;
+    private final ServiceInstrumenter serviceInstrumenter;
 
     private ServiceContext(Builder builder) {
         this.serializerRegistry = Preconditions.checkNotNull(builder.serializerRegistry,
                 "Missing required SerializerRegistry");
+        this.serviceInstrumenter = Preconditions.checkNotNull(builder.serviceInstrumenter,
+                "Missing required ServiceInstrumenter");
     }
 
     /**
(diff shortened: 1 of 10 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServiceContext.java
```diff
@@ -38,13 +41,26 @@ public SerializerRegistry serializerRegistry() {
         return serializerRegistry;
     }
 
+    /**
+     * {@link ServiceInstrumenter} to apply metric instrumentation to exposed services.
+     */
+    public ServiceInstrumenter serviceInstrumenter() {
```
rubenfiszel, line 47: Can we make this optional such that not instrumenting is the empty case.

iamdanfox: It's already a no-op by default?

iamdanfox:
This looks very neat, and is clearly super pluggable... 

Just playing devils-advocate: wasn't one of the big draws of the pure undertow generator that there is no reflection anymore, and adding a proxy kinda brings that back? Could we just codegen the instrumentation code we want?

carterkozak:
> wasn't one of the big draws of the pure undertow generator that there is no reflection anymore, and adding a proxy kinda brings that back?

Avoiding reflection definitely makes the code a great deal easier to understand and debug, and we can avoid generalizations by taking advantage of compile-time knowledge here. I would prefer to avoid reflection if possible, but jackson is already entirely reflective. The big draw of the undertow generator is correctness.

> Could we just codegen the instrumentation code we want?

We could, but it would significantly increase the maintenance burden for this project, and keeping parity with Tritium will be difficult over time. It may be worth doing eventually, but I don't think the time is right. We can deprecate the ServiceInstrumenter later on if we change our minds.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServiceContext.java
rubenfiszel: Yes, but my point was that the Optional pattern is imho more readable than the NO_OP one. But to be fair, that's just a matter of preference.

iamdanfox (approved):
LGTM, feel free to add 'merge when ready' when you want to pull the trigger

markelliot:
who fills in the plugins for this impl?

carterkozak:
> who fills in the plugins for this impl?

The context is configured by the server implementation, it cannot be modified by wc consumers.

PR #239, opened 2019-02-18
[fix] 'useInsensitiveEnums' feature flag fully restores old behaviour

iamdanfox:
## Before this PR

@jcvitanovic reported problems on an internal product where empty strings were persisted to a database, but trying to opt in to the old behaviour using this feature flag would still break.

```gradle
conjure {
    java {
        useInsensitiveEnums = true
    }
}
```

## After this PR

The feature flag brings back exactly the old pre-2.8.0 behavour, where strings are leniently deserialized.

_A full revert is still being considered here: https://github.com/palantir/conjure-java/pull/238_

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/insensitive/InsensitiveEnum.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/insensitive/InsensitiveEnum.java
@@ -2,7 +2,6 @@ package com.palantir.insensitive;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
-import com.palantir.conjure.java.lib.internal.ConjureEnums;
 import com.palantir.logsafe.Preconditions;
 import java.util.Locale;
 import javax.annotation.Generated;
@@ -75,7 +74,6 @@ public final class InsensitiveEnum {
             case "ONE_HUNDRED":
                 return ONE_HUNDRED;
             default:
-                ConjureEnums.validate(value);
                 return new InsensitiveEnum(Value.UNKNOWN, value);
         }
     }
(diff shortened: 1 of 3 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/insensitive/InsensitiveEnum.java
```diff
@@ -75,7 +74,6 @@ public static InsensitiveEnum valueOf(String value) {
             case "ONE_HUNDRED":
                 return ONE_HUNDRED;
             default:
-                ConjureEnums.validate(value);
                 return new InsensitiveEnum(Value.UNKNOWN, value);
```
iamdanfox, line 77: You can see this matches the pre 2.8.0 behaviour, where no validation is performed in the UNKNOWN case.

https://github.com/palantir/conjure-java/blob/2.7.1/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java#L62-L76

dansanduleac (approved):

conjure-java-core/src/test/java/com/palantir/conjure/java/types/EnumTests.java
```diff
@@ -60,6 +60,18 @@ public void testInsensitiveEnum_lowerCaseUnknown() {
         assertThat(value.get()).isEqualTo(InsensitiveEnum.Value.UNKNOWN);
     }
 
+    @Test
+    public void testInsensitiveEnum_unknownToleratesEmptyStringsForHistoricalReasons() {
+        InsensitiveEnum value = InsensitiveEnum.valueOf("");
+        assertThat(value.get()).isEqualTo(InsensitiveEnum.Value.UNKNOWN);
```
jcvitanovic, line 66: 👍

PR #245, opened 2019-02-21
[fix] Suppress deprecation warnings in generated undertow services

pkoenig10:
## Before this PR
Undertow service handlers for deprecated endpoints fail to compile when using `-Xlint:deprecation`. This wasn't a problem for Jersey services because the deprecated methods were not called in the generated code.

## After this PR
Undertow service handlers suppress deprecation warnings when calling deprecated endpoints.


carterkozak:
:+1: thanks!

PR #246, opened 2019-02-21
Reuse conjure-generator-common classes

dansanduleac:
## Before this PR

We had copied over the conjure spec implementation, and some type visitors too.

## After this PR

Re-use the classes made available by conjure's conjure-generator-common, and the published conjure-api for the IR spec.

@jamesthomp this will make you happy


dansanduleac:
@cakofony you're right it seems, just added base type and now it works

ferozco (approved):

PR #247, opened 2019-02-22
Suppress unchecked warnings in generated builders

pkoenig10:
## Before this PR
Conjure generated projects with covariant optional setters can not be compile with `-Xlint:unchecked`. This is because the improvements made in #9 requires casting optionals in certain cases.

## After this PR
Conjure generated builders now suppress unchecked warnings. This seems reasonable since we control the code generation here.

We could attempt to be more targeted and only add this annotation to covariant optional setters but:
- The builder generator code has relatively pure functions so passing this information back to the method level is not easy.
- Adding the annotation to the builder level prevents regression if we require casting in other setters in the future.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
@@ -106,6 +106,10 @@ public final class BeanBuilderGenerator {
                 .addAnnotation(
                         AnnotationSpec.builder(JsonIgnoreProperties.class)
                                 .addMember("ignoreUnknown", "$L", true)
+                                .build())
+                .addAnnotation(
+                        AnnotationSpec.builder(SuppressWarnings.class)
+                                .addMember("value", "$S", "unchecked")
                                 .build());
 
         return builder.build();
--- a/conjure-java-core/src/integrationInput/java/com/palantir/binary/BinaryExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/binary/BinaryExample.java
@@ -90,6 +90,7 @@ public final class BinaryExample {
 
     @Generated("com.palantir.conjure.java.types.BeanBuilderGenerator")
     @JsonIgnoreProperties(ignoreUnknown = true)
+    @SuppressWarnings("unchecked")
     public static final class Builder {
         private ByteBuffer binary;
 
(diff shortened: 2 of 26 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -106,6 +106,10 @@ private TypeSpec generate(ObjectDefinition typeDef) {
                 .addAnnotation(
                         AnnotationSpec.builder(JsonIgnoreProperties.class)
                                 .addMember("ignoreUnknown", "$L", true)
+                                .build())
+                .addAnnotation(
+                        AnnotationSpec.builder(SuppressWarnings.class)
+                                .addMember("value", "$S", "unchecked")
```
iamdanfox, line 112: At a minimum I think we need a comment here justifying why we add this to _all_ builders... I just ran it locally and got:

```
> Task :conjure-java-core:compileIntegrationInputJava FAILED
/Users/dfox/conjure-java/conjure-java-core/src/integrationInput/java/com/palantir/product/CovariantOptionalExample.java:107: warning: [unchecked] unchecked cast
            this.item = (Optional<Object>) Preconditions.checkNotNull(item, "item cannot be null");
                                                                     ^
  required: Optional<Object>
  found:    Optional<CAP#1>
  where CAP#1 is a fresh type-variable:
    CAP#1 extends Object from capture of ?
error: warnings found and -Werror specified
1 error
1 warning
```

iamdanfox:
It feels pretty gross to be adding the `@SuppressWarnings("unchecked")` to the whole builder.

Could we have a stab at adding it to exactly the right methods?

```diff

        @JsonSetter("item")
+       @SuppressWarnings("unchecked")
        public Builder item(Optional<?> item) {
            this.item = (Optional<Object>) Preconditions.checkNotNull(item, "item cannot be null");
            return this;
        }
```

Also, if we want conjure-java to pass without warnings then I think this PR needs to also enable `-Xlint:unchecked` in the top level build.gradle to ensure we don't regress somewhere else!!

carterkozak:
Rather than adding suppressions to generated code, should we provide an `Optional<Object> narrow(Optional<?>)` utility function in the library, so we have a single suppression?

iamdanfox:
TBH I'd really like to keep conjure-lib as minimal as possible. If we could target these annotations accurately then most generated code wouldn't be affected, it would only be for people using `any`, which seems not too bad?

iamdanfox:
Closing out as I don't think it's really OK to suppress warnings on all builders - lmk if you want to have another stab @pkoenig10!

PR #255, opened 2019-02-27
[undertow] add functions to apply to every marked params in the runtime

rubenfiszel:
## Before this PR

It is not possible to do any actions on marked params.

## After this PR

The runtime include functions that will be called on every marked params. This allow the possibility to to add safe logging among others.


conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -487,12 +491,50 @@ private CodeBlock generateParameterCodeBlock(Stream<ArgumentDefinition> params,
...
+    private Optional<String> getStoreParamMethodName(
+            ArgumentDefinition arg,
+            ParameterType.Visitor<Boolean> visitor) {
+        boolean isSafe = arg.getMarkers().stream().anyMatch(type ->
+                type.accept(
+                        new DefaultTypeVisitor<Boolean>() {
+                            @Override
+                            public Boolean visitExternal(ExternalReference value) {
+                                return value.getExternalReference().getName().equals("Safe")
+                                        && value.getExternalReference().getPackage().equals("com.palantir.redaction");
```
uschi2000: isn't the right package `com.palantir.logsafe`?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,70 @@
...
+    public static void putUnsafeQueryParam(HttpServerExchange exchange, String key, String value) {
+        putParam(exchange, UNSAFE_QUERY_PARAMS_ATTACH_KEY, key, value);
+    }
+
+    private static void putParam(
+            HttpServerExchange exchange,
+            AttachmentKey<Multimap<String, String>> attachmentKey,
+            String key,
+            String value) {
+        Optional<Multimap<String, String>> multimap = Optional.ofNullable(exchange.getAttachment(attachmentKey));
```
uschi2000: no need to use Optional in private/local context

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParametersHandler.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final HttpHandler delegate;
+
+    ParametersHandler(HttpHandler delegate) {
+        this.delegate = Preconditions.checkNotNull(delegate, "Delegate handler is required");
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        Multimap<String, String> safePathParams = HashMultimap.create();
```
uschi2000: I suspect that @cakofony would prefer to avoid object allocations on this code path where possible. Do we have to differentiate between query and path params?

uschi2000: what about body and header parameters, btw?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParametersHandler.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+import com.palantir.conjure.java.undertow.lib.Parameters;
+import com.palantir.logsafe.Preconditions;
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class ParametersHandler implements HttpHandler {
```
uschi2000: needs docs

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,70 @@
...
+import io.undertow.util.AttachmentKey;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class Parameters {
+
+    private static final Logger log = LoggerFactory.getLogger(Parameters.class);
+
+    public static final AttachmentKey<Multimap<String, String>> SAFE_PATH_PARAMS_ATTACH_KEY =
```
uschi2000: I think these should live in the handler

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,70 @@
...
+    public static final AttachmentKey<Multimap<String, String>> UNSAFE_PATH_PARAMS_ATTACH_KEY =
+            AttachmentKey.create(Multimap.class);
+    public static final AttachmentKey<Multimap<String, String>> SAFE_QUERY_PARAMS_ATTACH_KEY =
+            AttachmentKey.create(Multimap.class);
+    public static final AttachmentKey<Multimap<String, String>> UNSAFE_QUERY_PARAMS_ATTACH_KEY =
+            AttachmentKey.create(Multimap.class);
+
+    private Parameters() {}
+
+    public static void putSafePathParam(HttpServerExchange exchange, String key, String value) {
```
uschi2000: maybe these should also live in the handler?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
rubenfiszel: It appears it is not, else the generator tests would not pass.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: But isn't it more readable/style preference ?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParametersHandler.java
rubenfiszel: 👍

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: Currently, the generator depend on the lib and not on the runtime so we need them here unless we want to change that dependency.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: same as above

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParametersHandler.java
rubenfiszel: My reasoning for differentiating everything is that if not then we lose that information and it should be up to the consumer (which might need to do the differentiation) to decide if we should ignore it or not. 
But if the allocation cost is prohibitive we can decide that consumers should not be able to differentiate.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
uschi2000: in the undertow case we now need to encode the class that would *actually* be used, not the one used in isolation in tests. https://github.com/palantir/safe-logging/blob/develop/safe-logging/src/main/java/com/palantir/logsafe/Safe.java

rubenfiszel:
@uschi2000 WC does not support safe body or header parameters. For header parameters it uses a whitelist instead.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
uschi2000: depend on who you ask :) I personally think that nullable variables are ok *within the context of a single function*, and might even be ok *across private functions*, but never in a public API.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
uschi2000: ah, my bad.

uschi2000:
re header & body: what does Conjure support?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
rubenfiszel: Ah that makes sense, I thought Safe was an hardcoded/pre-imported marker and that it was always com.palantir.redaction.
There are a few alternatives:
- filter on getName == Safe exclusively (my preferred option, I don't think we will have many cases of conflict on the marker for Safe)
- filter on union of com.palantir.redaction.Safe and com.palantir.logsafe.Safe (pretty sad)
- rewrite the tests so that they depend on logsafe

What do you prefer?

rubenfiszel:
@uschi2000 The conjure spec only mentions that they are represented by types in the IR without any requirement for supporting particular ones. I could not find a spec for conjure-java.
So I am assuming the question is: what does the conjure-java-jersey-generator currently support? 
Safe markers for header and body are not tested but I modified the tests, and the effect is to write those annotations even for header and body params. But those annotations are without effect when not with our internal webserver library RequestLoggingAnnotationFilter so it is debatable that reaching feature-parity with Jersey is equivalent to also store header and body params.

There is a stronger argument imho which is the one of consistency: All params should be treated the same, as Argument of the endpoints, regardless of their actual wire reification. In that regard, I would agree that we would need to support Header and Body Param (although here we are storing safe **and unsafe params**, I am not sure we actually want to store heavy body params like binary stream as string attachment key of the exchange).

Nevertheless, my opinion is that it is an under-specified hack that we implement currently to unblock ourselves but without any requirement to do more than what is necessary. 
I'm not opposed to storing header params but body params do not make a lot of sense to me. Would that be satisfying? Also the differentiation question now extends to header param as well.

uschi2000:
I think filtering on “Safe” (case-insensitive) is my preferred option.

uschi2000:
Let’s add header parameters, but not body ones. Could you document the behavior, please?

rubenfiszel:
@uschi2000 Latest commit now only use 2 multimaps, one for safe and one for unsafe params. It supports header params and some docs has been added. It now filters on "Safe" name.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParametersHandler.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+import com.palantir.conjure.java.undertow.lib.Parameters;
+import com.palantir.logsafe.Preconditions;
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+
+/**
+ * {@link HttpHandler} that attach to the exchange a safe and an unsafe parameters MultiMap such that a subsequent
```
uschi2000: attaches

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ParametersHandlerTest.java
```diff
@@ -0,0 +1,96 @@
...
+                    Parameters.putSafeHeaderParam(exchange, "safeHeaderName", "safeHeaderValue");
+                    cExchange = exchange;
+                }))
+                .build();
+        server.start();
+    }
+
+    @AfterClass
+    public static void afterClass() {
+        if (server != null) {
```
uschi2000: can't be null, remove

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,79 @@
...
+import com.palantir.logsafe.SafeArg;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.util.AttachmentKey;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Utility class to put different types of safe or unsafe parameters in a multimap attached the exchange.
+ * It requires the exchanged o have been previously through a ParametersHandler which would have created the attached
```
uschi2000: typos

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public static void putSafePathParam(HttpServerExchange exchange, String key, Object value) {
+        putParam(exchange, SAFE_PARAMS_ATTACH_KEY, key, value);
+    }
+
+    public static void putUnsafePathParam(HttpServerExchange exchange, String key, Object value) {
+        putParam(exchange, UNSAFE_PARAMS_ATTACH_KEY, key, value);
+    }
+
+    public static void putSafeQueryParam(HttpServerExchange exchange, String key, Object value) {
```
uschi2000: collapse to putSafeParam and putUnsafeParam

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,79 @@
...
+    public static void putUnsafeHeaderParam(HttpServerExchange exchange, String key, Object value) {
+        putParam(exchange, UNSAFE_PARAMS_ATTACH_KEY, key, value);
+    }
+
+    private static void putParam(
+            HttpServerExchange exchange,
+            AttachmentKey<Multimap<String, Object>> attachmentKey,
+            String key,
+            Object value) {
+        Optional<Multimap<String, Object>> multimap = Optional.ofNullable(exchange.getAttachment(attachmentKey));
```
uschi2000: remove Optional

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,79 @@
...
+        if (multimap.isPresent()) {
+            multimap.get().put(key, value);
+        } else {
+            log.warn("Attachment key not found. Impossible to put param {}.", SafeArg.of("param-name", key));
```
uschi2000: nit:
```

log.warn("Failed to attached endpoint parameter because the the required exchange attachment does not exist. Did this request get handled by the ParametersHandler?", SafeArg.of("attachmentKey", attachmentKey), SafeArg.of("parameterName", key));
```

uschi2000 (approved):
looks good modulo a few nitpicks

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParametersHandler.java
rubenfiszel: ty

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ParametersHandlerTest.java
rubenfiszel: 👍

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: ty

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: On that one I was less certain cause I felt that this is an internal detail. If we ever decide to revert that decision and split the different types, the generated code is still valid and we only need to change the lib. What do you think?

rubenfiszel: This is the last point to resolve. Do you feel strongly about this?

carterkozak:
I’d like to take a quick pass over this PR this evening if you don’t mind tainting a bit!

-ck

rubenfiszel:
@cakofony ofc!

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,81 @@
+/*
```
carterkozak: I think this belongs in `lib.internal` since it's not meant to be used by code written by humans

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,81 @@
...
+ * MultiMaps.
+ */
+public final class Parameters {
+
+    private static final Logger log = LoggerFactory.getLogger(Parameters.class);
+
+    public static final AttachmentKey<Multimap<String, Object>> SAFE_PARAMS_ATTACH_KEY =
+            AttachmentKey.create(Multimap.class);
+    public static final AttachmentKey<Multimap<String, Object>> UNSAFE_PARAMS_ATTACH_KEY =
+            AttachmentKey.create(Multimap.class);
```
carterkozak: We can move these to `Attachments.java` since they are consumed by custom request logging code

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,81 @@
...
+ * It requires the exchange to have been priorly through a ParametersHandler which would have created the attached
+ * MultiMaps.
+ */
+public final class Parameters {
+
+    private static final Logger log = LoggerFactory.getLogger(Parameters.class);
+
+    public static final AttachmentKey<Multimap<String, Object>> SAFE_PARAMS_ATTACH_KEY =
+            AttachmentKey.create(Multimap.class);
+    public static final AttachmentKey<Multimap<String, Object>> UNSAFE_PARAMS_ATTACH_KEY =
```
carterkozak: Is there any value in logging unsafe components? Why not only add safe ones?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Utility class to put different types of safe or unsafe parameters in a MultiMap attached to the exchange.
+ * It requires the exchange to have been priorly through a ParametersHandler which would have created the attached
+ * MultiMaps.
+ */
+public final class Parameters {
+
+    private static final Logger log = LoggerFactory.getLogger(Parameters.class);
+
+    public static final AttachmentKey<Multimap<String, Object>> SAFE_PARAMS_ATTACH_KEY =
```
carterkozak: Do these need to be multimaps? If we key by the variable name we're guaranteed uniqueness because a method may not have multiple parameters with the same name.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParametersHandler.java
```diff
@@ -0,0 +1,46 @@
...
+import com.palantir.conjure.java.undertow.lib.Parameters;
+import com.palantir.logsafe.Preconditions;
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+
+/**
+ * {@link HttpHandler} that attaches to the exchange a safe and an unsafe parameters MultiMap such that a subsequent
+ * handler can attach parameters to the exchange using {@link Parameters} utility methods.
+ */
+public final class ParametersHandler implements HttpHandler {
```
carterkozak: If we keep this, it should be package private

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParametersHandler.java
```diff
@@ -0,0 +1,46 @@
...
+        this.delegate = Preconditions.checkNotNull(delegate, "Delegate handler is required");
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        Multimap<String, Object> safeParams = HashMultimap.create();
+        Multimap<String, Object> unsafeParams = HashMultimap.create();
+        exchange.putAttachment(Parameters.SAFE_PARAMS_ATTACH_KEY, safeParams);
+        exchange.putAttachment(Parameters.UNSAFE_PARAMS_ATTACH_KEY, unsafeParams);
+        delegate.handleRequest(exchange);
```
carterkozak: I don't think this handler is necessary, rather than logging a warning when the map is not present, the `Parameters` code can add a new map attachment. Not all endpoints have parameters, so we don't need to create or apply these maps to the exchange. Doing so is a bit dangerous because this handler could be applied twice, and lose references to previous maps.

Most services will have a mix of conjure-undertow and other handlers (like servlets) which will not apply these attachments, so our request logging implementation always needs to handle null attachment values.

rubenfiszel: I'm fine with removing the handler

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: There is only one case where we are not guaranteed, headers have the name of the actual header name and not the header argument. I am also potentially interested to stash more things into the params thing (see the WC PR #1303)

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: WC was logging unsafe query and path params. I think the biggest value is CI and on-prem debugging.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: Sounds good!

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: Sounds goods!

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParametersHandler.java
rubenfiszel: See below, I am fine with removing this class.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
carterkozak: If I'm reading this correctly, we're logging everything as unsafe, where I think wc only logs things annotated with `@Unsafe` as unsafe parameters, right? We shouldn't log credential data as unsafe by default, because even logging a BearerToken (or other non-bearer credential) to disk is dangerous.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
carterkozak: bikeshed: The more I think about this, I'd prefer a standard map because we record this data in a standard map. My preference is to throw an exception when there are multiple attempts to record data under the same name.

This allows us to solve the multi-value problem where currently WC logs `["foo"]` as `"foo"` because it assumes everything might be a list. We've generated the code to solve this problem already, and we should record exactly the received value, avoiding datatype mismatches depending on the number of parameters provided.

Headers: We can use the header parameter name rather than the header name.

Alternative solution:

We can entirely avoid logic here, and change the attachment to `List<Arg<?>>` which does allow duplicate keys, and allow the consumer (request logging implementation) to decide how to handle that data.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: We are not logging AuthHeader in WC but WC consider every query and path params not safe as unsafe and log them. It is not taking into account any `@Unsafe` notation.

rubenfiszel: Let me actually make that more precise, we do log header params but only those in a whitelist.

carterkozak: Our header logging isn't connected to service parameter logging though. Our whitelisted set of headers are already logged regardless of endpoint.

Are you sure that path parameters area default logged as unsafe? I had thought they were omitted, but I may be misremembering.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: By throwing exception, you mean at conjure generator time, or at runtime? Would that mean the endpoint always error out?

re: headers, if we use the header parameter name then we cannot use the whitelist anymore as a safe guard but maybe we don't need that additional protection.

The `List<Arg>` is interesting, maybe that's actually the best solution after all.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Parameters.java
rubenfiszel: re: path parameters default:

That test snippet in our WC tests is relevant
```
        assertThat(requestLog2.getUnsafeParams()).contains(
                Pair.of("path2", "path2V"),
                Pair.of("query", "query1V"));
```

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowRuntime.java
```diff
@@ -32,4 +32,8 @@
 
     /** Provides the {@link AuthorizationExtractor} used to read auth tokens from request headers. */
     AuthorizationExtractor auth();
+
+    /** Provides the {@link SafeParamStorer} used to store safe parameters (header, query, body params) for further
+     * processing. (e.g logging) */
+    SafeParamStorer safeParamStorer();
```
iamdanfox: Not a huge fan of adding this here, as everything reachable from `UndertowRuntime` previously really neatly map to first-class concepts in Conjure... feels like this will set a precedent that turns the UndertowRuntime class into a bit of a grab-bag of disparate functionality.

Also adding methods to an interface that other people can implement technically breaks compile :/

iamdanfox: Not really sure how this is better than just making up an Attachment key and directly calling `exchange.putAttachment` with the correctly-presized map?

rubenfiszel: @cakofony thoughts on the above ?

I could put a key as a public static field of a new Attachments class conjure-undertow-lib.

carterkozak: The attachment would have to live in conjure-undertow-lib, despite being an implementation detail, the rest of our attachments live in the runtime. Attachment keys are based on object identity, so we must share the code that defines the key rather than a stringly matching property.

> Not a huge fan of adding this here, as everything reachable from UndertowRuntime previously really neatly map to first-class concepts in Conjure.

This is a justification for not supporting safe-logging marked parameters at all, which I wouldn't particularly mind, but I think a lot of folks would! If it's something that we actually want to support, the implementation belongs somewhere in the runtime imo.

I wonder if we can map this feature to a conjure concept?

```java
// invocation generated for any parameter with markers. Would we need a better pojo to describe parameter type (body, path, etc) and other metadata (java type) here?
// marker sets would be extracted to Endpoint level constants, and would not use guava
runtime.markers().markParameter("parameterName", parameterValue, ImmutableSet.of(Safe.class));
```

> Also adding methods to an interface that other people can implement technically breaks compile :/

That's why the javadoc for `UndertowRuntime` includes this, giving us (theoretically) the freedom add methods, just not to remove them.

```java
 * The {@link UndertowRuntime} and provided interfaces {@link BodySerDe}, {@link PlainSerDe}, and
 * {@link AuthorizationExtractor} are internal API, no guarantees are made for custom implementations.
```

rubenfiszel: @cakofony Interesting proposal.
If we go in this direction:
-  I'd like indeed to differentiate between the different type of parameters (either by passing an enum ParameterType) or by having different methods
- markers should be strings instead of class and represented as such in the IR.

rubenfiszel:
@iamdanfox @cakofony @uschi2000 This is getting stale. What should we do about this PR?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -519,12 +528,70 @@ private CodeBlock generateParameterCodeBlock(Stream<ArgumentDefinition> params,
...
                 }).collect(Collectors.toList()));
     }
 
+    private Optional<String> getStoreMethodAndParamName(ArgumentDefinition arg) {
+        boolean isSafe = arg.getMarkers().stream().anyMatch(type ->
+                type.accept(
+                        new DefaultTypeVisitor<Boolean>() {
+                            @Override
+                            public Boolean visitExternal(ExternalReference value) {
+                                return value.getExternalReference().getName().equals("Safe");
```
pkoenig10: This should probably get the class name of this marker to verify it is actually `com.palantir.logsafe.Safe` and not just another class named `Safe`. Unless we are making a plain `Safe` marker a first-class concept in conjure in which case we should update the spec.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -519,12 +528,70 @@ private CodeBlock generateParameterCodeBlock(Stream<ArgumentDefinition> params,
...
+                            @Override
+                            public Boolean visitDefault() {
+                                return false;
+                            }
+                        }));
+        if (isSafe) {
+            return arg.getParamType().accept(new ParameterType.Visitor<Optional<String>>() {
+                @Override
+                public Optional<String> visitHeader(HeaderParameterType value) {
+                    return Optional.of(value.getParamId().get());
```
pkoenig10: We probably should add checks here to prevent logging non-whitelisted headers even if they are incorrectly marked `Safe`.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
rubenfiszel: See https://github.com/palantir/conjure-java/pull/255#discussion_r260938457

I don't feel strongly about it but I think the case where it is marked as safe but with another imported conjure class and that the expected behavior is to be different is quite convoluted. There is also some thoughts to make markers plain strings so this would be more aligned with it.

rubenfiszel: also see https://github.com/palantir/conjure-java/pull/255#issuecomment-468175097 

btw, i'm refactoring the whole PR.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowRuntime.java
```diff
@@ -32,4 +32,8 @@
 
     /** Provides the {@link AuthorizationExtractor} used to read auth tokens from request headers. */
     AuthorizationExtractor auth();
+
+    /** Provides the {@link MarkedParam} used to store safe parameters (header, query, body params) for further
+     * processing. (e.g logging) */
+    MarkedParam markedParam();
```
carterkozak: Thoughts on renaming this method to `markers()`?

rubenfiszel: What about other types of elements that can be marked but that aren't concerned by this. Isn't it ambiguous to call this only markers?

carterkozak: I think this interface can support all markers, since they exist to provide metadata there's not a large difference between the marker types.

rubenfiszel: Sure but then we should probably either pass the marked element type to the mark method or provide a markParam, markX, etc.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -62,6 +74,7 @@ public AuthorizationExtractor auth() {
     public static final class Builder {
 
         private final List<Encoding> encodings = Lists.newArrayList();
+        private List<MarkedParam> markedParamActions = Lists.newArrayList();
```
carterkozak: nit: final

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowRuntime.java
rubenfiszel: Would you be satisfied with MarkedParam -> Markers, markerParam -> markers, mark -> markParam ?

carterkozak: That makes sense, though I think it may be cleaner to leave `mark` as is since marking other types of data would use a different method signature. I don't have a strong opinion, `markParam` is also reasonable.

conjure-java-core/src/test/resources/example-service.yml
```diff
@@ -68,6 +68,8 @@ services:
             param-id: Test-Header
             param-type: header
             type: string
+            markers:
+              - Safe
```
carterkozak, line 72: We might want a test with two markers on the same parameter

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -71,6 +84,12 @@ public Builder encodings(Encoding value) {
             return this;
         }
 
+        @CanIgnoreReturnValue
+        public Builder markedParamAction(MarkedParam value) {
```
carterkozak: We may want to use a different interface here, that way we can add new methods to `MarkedParam` without breaking existing ConjureUndertowRuntime configurations.

carterkozak:
I like the simplicity of this approach. It doesn't force special-case code specifically for safe/unsafe parameters and maintains the same level of separation we've had in the jersey implementation.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
rubenfiszel: done

conjure-java-core/src/test/resources/example-service.yml
rubenfiszel: done

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
rubenfiszel: done

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -89,6 +89,7 @@
     private static final String DESERIALIZER_VAR_NAME = "deserializer";
     private static final String SERIALIZER_VAR_NAME = "serializer";
     private static final String AUTH_HEADER_VAR_NAME = "authHeader";
+    private static final String MARKED_PARAM_VAR_NAME = "markers().mark";
```
carterkozak: markers().param?

rubenfiszel: yes ...

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParamMarkers.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.palantir.conjure.java.undertow.runtime;
+
+import io.undertow.server.HttpServerExchange;
+
+/**
+ * execute arbitrary actions given a marker on a mark.
+ */
+public interface ParamMarkers {
```
carterkozak: maybe singular `ParamMarker<T>` with `void mark(Class<T> markerClass`?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
pkoenig10: I feel pretty strongly that for now we should mimic the current behavior of the Jersey resources, which look [specifically and only for `com.palantir.logsafe.Safe`](https://github.palantir.build/foundry/sls-logging/blob/5.2.0/sls-logging-jersey/src/main/java/com/palantir/sls/logging/RequestLoggingFilter.java#L199). I think changing this makes sense, but we should have a discussion about updating the spec before making decision like this in arbitrary PRs.

carterkozak: @pkoenig10 That's the plan, this PR has been refactored so that we can retain the strictness of that check without special casing in conjure-undertow. wc will provide a `ParamMarker` implementation for `Safe.class`.

carterkozak:
I'm happy with this. Any thoughts before it merges @iamdanfox @uschi2000?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParamMarker.java
```diff
@@ -0,0 +1,27 @@
...
+package com.palantir.conjure.java.undertow.runtime;
+
+import io.undertow.server.HttpServerExchange;
+
+/**
+ * execute arbitrary actions given a marker on a param.
+ */
+public interface ParamMarker {
+
+    void mark(Class<?> markerClass, String parameterName, Object parameterValue, HttpServerExchange exchange);
```
uschi2000: this conflicts with the `[RFC] Markers and “safe” parameters in Conjure` (internal). I don't think we should cement the idea of markers being classes in this PR.

uschi2000: this API opens the door for pretty wild hacks, e.g., register some marker and then mutate the exchange. is that desirable?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Markers.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.undertow.lib;
+
+import io.undertow.server.HttpServerExchange;
+
+public interface Markers {
```
uschi2000: this name isn't good. `MarkerCallback`?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParamMarker.java
rubenfiszel: 1) I agree. What about providing the package and class name as strings?
2) Mutating the exchange is what we need to do for safe-logging to stash the args in an attachment. We could use a layer of abstraction to reduce the API surface but i'm not sure what it really buys us. In the case of WC, consumers will not be able to override the runtime.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Markers.java
rubenfiszel: MarkerCallback sounds good to me. @carterkozak thoughts?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ParamMarker.java
uschi2000: 2. ok. please add plenty of code docs when implementing the stashing and the collecting, because it's going to be pretty hard to follow how things fit together.

uschi2000: 1: maybe just the full/canonical class name?

rubenfiszel: 1. done
2. will do when we implement it in WC

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Markers.java
rubenfiszel: renamed to MarkerCallback

uschi2000:
lgtm

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -369,7 +369,7 @@ private ParameterSpec createServiceMethodParameterArg(TypeMapper typeMapper, Arg
 
     private static Set<AnnotationSpec> createMarkers(TypeMapper typeMapper, List<Type> markers) {
         checkArgument(markers.stream().allMatch(type -> type.accept(TypeVisitor.IS_REFERENCE)),
-                "Markers must refer to reference types.");
+                "MarkerCallback must refer to reference types.");
```
carterkozak:
```suggestion
                "Markers must refer to reference types.");
```

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -368,6 +371,19 @@ private CodeBlock endpointInvocation(EndpointDefinition endpointDefinition, List
         return code.build();
     }
 
+    private CodeBlock generateParamMarkers(
+            List<Type> markers,
+            String paramName,
+            TypeMapper typeMapper) {
+        return CodeBlocks.of(markers.stream().map(marker ->
+                CodeBlock.of("$1N.$2N($3S, $4S, $5N, $6N);",
+                        RUNTIME_VAR_NAME, MARKED_PARAM_VAR_NAME,
```
iamdanfox: Please could we just inline MARKED_PARAM_VAR_NAME here because it's not used in any other places?

conjure-java-core/src/test/resources/test/api/TestServiceEndpoints.java.undertow
```diff
@@ -696,6 +716,8 @@ public final class TestServiceEndpoints implements UndertowService {
                     exchange.getAttachment(PathTemplateMatch.ATTACHMENT_KEY).getParameters();
             ResourceIdentifier datasetRid =
                     runtime.plainSerDe().deserializeRid(pathParams.get("datasetRid"));
+            runtime.markers()
+                    .param("com.palantir.redaction.Safe", "datasetRid", datasetRid, exchange);
```
iamdanfox, line 720: :+1: this is really nice, really like how it doesn't back us into a corner in terms of future development :)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
rubenfiszel: done

iamdanfox (approved):
Looks great to me!

PR #260, opened 2019-03-04
[break][Undertow] Rewrite conjure-undertow API

carterkozak:
## All implementation has moved to the runtime module
* `conjure-undertow-lib` only contains interfaces.
* `ServiceContext` -> `UndertowRuntime`
* `StringDeserializers` -> `UndertowRuntime.plainSerDe()`
* `Auth.cookie/header` -> `UndertowRuntime.auth().cookie/header`

**Rationale**: All runtime behavior should be updated in lock step in `conjure-java-runtime`, this change allows the undertow runtime to behave similarly to jersey.

## Renamed Service to UndertowService
Registering generated service implementations using `.api(Service)` failed to make it clear what `Service` was. `.api(UndertowService)` is immediately understandable.

## `UndertowService` creates a list of `Endpoint`
`Registerable` and `EndpointRegistry` are no longer needed in this simplified model. Generated code is easier to understand and only contains one level of nested inner classes.

## Serializers and Deserializers are typed
Serializer and Deserializer objects are created eagerly when Endpoints are constructed in order to match the conjure specification more closely. Each endpoint receives exactly one body type, and returns exactly one type. 

This avoids unnecessary type mapping for each request.

## Immutable UndertowHandler
Endpoints must be added to the `UndertowHandler.Builder` before a `HttpHandler` is created to ensure behavior remains constant once a server has started.

## Example generated `EmptyPathServiceEndpoints.java`

Note that the output is more verbose than previous versions, however it's also less complex in both legibility runtime cost.

```java
@Generated("com.palantir.conjure.java.services.UndertowServiceHandlerGenerator")
public final class EmptyPathServiceEndpoints implements UndertowService {
    private final EmptyPathService delegate;

    private EmptyPathServiceEndpoints(EmptyPathService delegate) {
        this.delegate = delegate;
    }

    public static UndertowService of(EmptyPathService delegate) {
        return new EmptyPathServiceEndpoints(delegate);
    }

    @Override
    public List<Endpoint> endpoints(UndertowRuntime runtime) {
        return ImmutableList.of(new EmptyPathEndpoint(runtime, delegate));
    }

    private static final class EmptyPathEndpoint implements HttpHandler, Endpoint {
        private final UndertowRuntime runtime;

        private final EmptyPathService delegate;

        private final Serializer<Boolean> serializer;

        EmptyPathEndpoint(UndertowRuntime runtime, EmptyPathService delegate) {
            this.runtime = runtime;
            this.delegate = delegate;
            this.serializer = runtime.bodySerDe().serializer(new TypeToken<Boolean>() {});
        }

        @Override
            boolean result = delegate.emptyPath();
            serializer.serialize(result, exchange);
        }

        @Override
        public HttpString method() {
            return Methods.GET;
        }

        @Override
        public String template() {
            return "/";
        }

        @Override
        public String serviceName() {
            return "EmptyPathService";
        }

        @Override
        public String name() {
            return "emptyPath";
        }

        @Override
        public HttpHandler handler() {
            return this;
        }
    }
}
```


carterkozak:
code block in the readme needs to be updated

conjure-java-core/src/test/resources/test/api/TestServiceEndpoints.java.undertow.binary
```diff
@@ -50,7 +48,7 @@ public final class TestServiceEndpoints implements Service {
             @Override
             public void handleRequest(HttpServerExchange exchange) throws IOException {
                 BinaryResponseBody result = delegate.getBinary();
-                BinarySerializers.serialize(result, exchange);
+                context.serialize(result, exchange);
```
cakofony: rename to `serializeBody`/`deserializeBody`?

conjure-java-server-verifier/src/test/java/com/palantir/conjure/java/verification/server/undertest/UndertowServerUnderTestRule.java
```diff
@@ -40,7 +41,7 @@ protected void before() throws Throwable {
         Service endpoints = AutoDeserializeServiceEndpoints.of(service);
 
         ConjureHandler handler = new ConjureHandler();
-        ServiceContext context = ServiceContext.builder()
+        ServiceContext context = ConjureContext.builder()
                 .serializerRegistry(ConjureSerializerRegistry.getDefault())
```
carterkozak: There's no reason to expose the serializer registry, serializers should be added directly to the `ConjureContext`. By default, we should use the `ConjureSerializerRegistry.getDefault()` configuration.

conjure-java-core/src/integrationInput/java/com/palantir/product/EteBinaryServiceEndpoints.java
```diff
@@ -80,20 +74,20 @@ public void register(EndpointRegistry endpointRegistry) {
 
             @Override
             public void handleRequest(HttpServerExchange exchange) throws IOException {
-                AuthHeader authHeader = Auth.header(exchange);
-                InputStream body = BinarySerializers.deserializeInputStream(exchange);
+                AuthHeader authHeader = context.authHeader(exchange);
```
markelliot: this doesn't really seem like a 'context' function

carterkozak: I agree that the `ServiceContext` name doesn't quite fit anymore. We can rename in #261, ideas for a better fit?

markelliot: seems different than a naming problem -- what's the intent of the object? it seems pretty muddy, now.

carterkozak: The point is to provide implementations of all non-generated logic. Rather than provide a static utility class in lib, we add methods to this interface, which is provided to generated services. As demonstrated in #261 we delete the majority of the -lib module in favor of moving all logic into the runtime.
We could provide separate interfaces for these functions accessible from the context, however that makes the generated code a bit dirtier due to chained accessors.

markelliot: almost all of [ServiceContext](https://github.com/palantir/conjure-java/blob/58d3abc011a77e83a895a293f42c3b9ae33791fd/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServiceContext.java) is serde logic with exception of authHeader, authCookie and instrument. To me, there seem to be 3 distinct concerns:
 - Serialization
 - Parameter extraction for HttpServerExchange (or, more narrowly: Authorization)
 - "instrumentation" -- which seems superfluous given the signature of this method (it gets access to the implementation + its interface class, which is already available to anyone constructing a service

Given that, I'd propose:
 - remove instrument
 - move authHeader and authCookie to a new interface called AuthorizationExtractor
 - rename ServiceContext to SerDe

markelliot: ...possibly more correct to call it `UndertowSerDe`

carterkozak: In the future we may need to provide functionality that is not strict ser/de, for example tracking safe/unsafe parameters for request logging #255, or supporting asynchronous request processing by returning a `ListenableFuture`. In these cases we cannot plumb through new types without API breaks, so I'd prefer to add methods to the context. Perhaps `UndertowRuntime`?

markelliot: I think if you go with `UndertowRuntime` it'd be good to break it up into parts (serde, async, etc.) so that it doesn't become a dumping ground for stuff. (the code-gen to make `runtime.serde().<call you care about>` isn't a huge deal, IMO)

carterkozak: Works for me, I'll put something together.

rubenfiszel: I agree strongly with @markelliot about splitting and nesting specialized runtime components.

In general I have an issue with naming of runtime and lib. Here is a proposal of splitting and naming of conjure-java:
- conjure-generator
- conjure-lib-api
  - conjure-lib-api-serde
  - ...
- conjure-lib-runtime
  - conjure-lib-runtime-serde
  - ...

conjure-lib-api would be shared between the generator and the runtime.

conjure-lib-api would have to respect semantic versioning. 

For a given proejct, the generator, the runtime and dependencies (for instance WC) must rely on the same major version of the API.
Hence, the dependencies should aim to rely on an conjure-Iib-api as low as possible. 
The runtime can be on a higher version than the generator but not the opposite.
So the dependency constraints would become:
conjure-lib-api version: `dependencies <= generator <= runtime`

rubenfiszel (commented):
I think this is the right direction but we could take advantage of this PR to be more ambitious about the naming of the repo between runtime, api and generators

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointRegistry.java
```diff
@@ -27,6 +27,6 @@
  */
 public interface EndpointRegistry {
 
-    EndpointRegistry add(Endpoint description, HttpHandler handler);
+    void add(Endpoint endpoint);
```
carterkozak: Should we use `void register(Endpoint endpoint)` instead?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public static final class Builder {
+
+        private SerializerRegistry serializerRegistry;
+
+        private Builder() {}
+
+        @CanIgnoreReturnValue
+        public Builder serializerRegistry(SerializerRegistry value) {
```
carterkozak: Replace with a function to add individual `Serializer` instances

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowService.java
```diff
@@ -16,13 +16,19 @@
...
 /**
- * Creates a {@link Registrable} which may be registered with a web server. The server is responsible
- * for providing a {@link ServiceContext} allowing API implementors to add APIs using
+ * A {@link UndertowService} provides a list of {@link Endpoint endpoints} when it is
+ * given the {@link UndertowRuntime}. These {@link Endpoint endpoints} may be
+ * registered with an {@link EndpointRegistry} in order to be exposed by a
+ * web server. The server is responsible for providing an {@link UndertowRuntime} and
+ * orchestrating registration with the {@link EndpointRegistry} allowing API
+ * implementors to add APIs using
  * <code>server.api(MyServiceEndpoints.of(myServiceImpl)</code>.
```
markelliot: is the `MyServiceEndpoints` thing still accurate?

markelliot: Also it seems like we should describe that conjure-generated server stubs are implementations of `UndertowService` that contain an `Endpoint` in the returned list for each endpoint described in a given Conjure service.

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoints.java
```diff
@@ -1,57 +1,80 @@
...
+                return Methods.GET;
+            }
+
+            @Override
+            public final String template() {
+                return "/";
+            }
+
+            @Override
+            public final Optional<String> serviceName() {
```
markelliot: what are the cases when there's no service name?

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoints.java
```diff
@@ -1,57 +1,80 @@
...
+                return Optional.of("EmptyPathService");
+            }
+
+            @Override
+            public final Optional<String> name() {
```
markelliot: what are the cases when an endpoint wouldn't have a name?

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoints.java
carterkozak: These are optional so that we may phase out the name in the future if we need to. `serviceName` and `name` are only used internally to record metrics.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointRegistry.java
rubenfiszel: 👍

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowService.java
carterkozak: > is the MyServiceEndpoints thing still accurate?

Yes, the `of` method is static and generated on all `UndertowService` classes. Will update javadoc momentarily.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowService.java
```diff
@@ -16,13 +16,19 @@
...
  */
-public interface Service {
+public interface UndertowService {
 
-    Registrable create(ServiceContext context);
+    List<Endpoint> endpoints(UndertowRuntime runtime);
```
carterkozak, line 30: should this be `createEndpoints`, or are we happy with the `endpoints` method name?

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoints.java
```diff
@@ -1,58 +1,74 @@
...
+        }
+
+        @Override
+        public Optional<String> name() {
+            return Optional.of("emptyPath");
+        }
+
+        @Override
+        public HttpHandler handler() {
+            return this;
```
carterkozak, line 75: Each endpoint is a fair bit longer than before, however I find this less complex. If we're worried about LoC we can provide an abstract implementation and pass data to the ctor, however then we have to worry about api/abi compatibility on that class as well.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
```diff
@@ -16,57 +16,39 @@
...
-    static Endpoint put(String template, String serviceName, String name) {
-        return ImmutableEndpoint.of(Methods.PUT, template, Optional.of(serviceName), Optional.of(name));
-    }
+    /** Undertow {@link HttpHandler} which provides the endpoint implementation. */
+    HttpHandler handler();
 
-    static Endpoint delete(String template) {
-        return ImmutableEndpoint.of(Methods.DELETE, template, Optional.empty(), Optional.empty());
+    /** Simple name of the service which provides this endpoint. This data may be used for metric instrumentation. */
+    default Optional<String> serviceName() {
```
rubenfiszel: Could we make this non optional ? What are the cases where there is no name ?
If there are cases like this, I would almost prefer them to pass a fake "NO_SERVICE_NAME" if needed.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
```diff
@@ -16,57 +16,39 @@
...
+        return Optional.empty();
     }
 
-    static Endpoint delete(String template, String serviceName, String name) {
-        return ImmutableEndpoint.of(Methods.DELETE, template, Optional.of(serviceName), Optional.of(name));
+    /** Simple name of the endpoint method. This data may be used for metric instrumentation. */
+    default Optional<String> name() {
+        return Optional.empty();
```
rubenfiszel: same as above, could we make this non-optional

rubenfiszel (approved):
Clear and significative improvement in readability and complexity reduction.
2 things that I would prefer:
- make name and serviceName non-optional in Endpoint
- reduce some of the generate boilerplate around the endpoint definition.
Would be in favor of something akin to 
`Endpoint = (HttpHandler, EndpointDefinition)`
`EndpointDefinition = (Method, Template, ServiceName, Name)`
Such that you reduce some of the boilerplate with `EndpointDefinition.of(GET, path, serviceName, name)`
Not sure about the EndpointDefinition naming.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -64,27 +67,59 @@
...
+        private HttpHandler fallback = ResponseCodeHandler.HANDLE_404;
+
+        private Builder() { }
+
+        public Builder endpoints(Endpoint value) {
+            endpoints.add(Preconditions.checkNotNull(value, "Value is required"));
+            return this;
+        }
+
+        public Builder addAllEndpoints(Iterable<Endpoint> values) {
```
carterkozak, line 107: I used conjure bean style names. using `Builder endpoint(Endpoint value)` above (rather than `endpoints`) and calling this method `endpoints` without clearing existing values may be cleaner, but wouldn't match what we do for generated objects.

conjure-java-core/src/integrationInput/java/com/palantir/product/EteBinaryServiceEndpoints.java
```diff
@@ -1,127 +1,229 @@
...
+    public List<Endpoint> endpoints(UndertowRuntime runtime) {
+        return ImmutableList.of(
+                new PostBinaryEndpoint(runtime, delegate),
+                new GetOptionalBinaryPresentEndpoint(runtime, delegate),
+                new GetOptionalBinaryEmptyEndpoint(runtime, delegate),
+                new GetBinaryFailureEndpoint(runtime, delegate));
     }
 
-    private static final class EteBinaryServiceRegistrable implements Registrable {
+    private static final class PostBinaryEndpoint implements HttpHandler, Endpoint {
```
markelliot, line 43: now that these are static classes, could also consider making them package private and splitting into independent files. Will help deal with overall length of the code.

conjure-java-core/src/integrationInput/java/com/palantir/product/EteBinaryServiceEndpoints.java
```diff
@@ -1,127 +1,229 @@
...
-                if (result.isPresent()) {
-                    BinarySerializers.serialize(result.get(), exchange);
-                } else {
-                    exchange.setStatusCode(StatusCodes.NO_CONTENT);
-                }
+        @Override
+        public void handleRequest(HttpServerExchange exchange) throws IOException {
+            AuthHeader authHeader = runtime.auth().header(exchange);
+            Optional<BinaryResponseBody> result = delegate.getOptionalBinaryPresent(authHeader);
+            if (result.isPresent()) {
```
markelliot, line 102: it might be kind of nice to handle optionals explicitly in the runtime

markelliot: could imagine an explicit SerDe method which accepts `Optional<?>` and sets the NO_CONTENT on the status, for instance.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/SerDe.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /** Deserializes the request body into the requested type. */
+    <T> T deserialize(TypeToken<T> type, HttpServerExchange exchange) throws IOException;
+
+    /** Reads an {@link InputStream} from the {@link HttpServerExchange} request body. */
+    InputStream deserializeInputStream(HttpServerExchange exchange);
+
+    // query, path, and header parameter deserializers
+
+    BearerToken deserializeBearerToken(String in);
```
markelliot: probably worth at least a TODO to doc this now that it's public API

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowService.java
markelliot: I think 'endpoints' is probably fine?

conjure-java-core/src/integrationInput/java/com/palantir/product/EteBinaryServiceEndpoints.java
carterkozak: That would be nice, however it would need to support `Optional`, `OptionalInt` `OptionalDouble`, as well as `alias: optional<*`. This would also duplicate our jackson configuration for serializing optional values which is also used for nested fields. I don't think it's the right approach right now, and supporting it would be an additive change we can make later.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/SerDe.java
carterkozak: Good idea, will do

conjure-java-core/src/integrationInput/java/com/palantir/product/EteBinaryServiceEndpoints.java
carterkozak: Services are generated in user-defined packages, it's likely that user-defined code exists in the same package and can access package private components.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
carterkozak: Updated.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
carterkozak: done

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
rubenfiszel: I understand the arguments but I would still prefer `endpoint(Endpoint value)` and `endpoints(Endpoint... endpoints)`

dansanduleac: is that a worthwhile performance optimization? how much of an overhead does a 1-element array incur?

carterkozak:
I think this change is in a good place. A few changes we should consider before merging this break, as we won't be able to do this again:

## Packages

Currently everything lives in `com.palantir.conjure.java.undertow.lib`. This definitely makes sense for `BinaryResponseBody` which is implemented directly by conjure users, however we have two additional types of interfaces:

* Implemented only by generated code
  * `Endpoint`
  * `UndertowService`
* Implemented only by the runtime, and consumed exclusively by generated code
  * `UndertowRuntime`
  * `SerDe`
  * `AuthorizationExtractor`

Any suggestions for subpackage names for these? The latter could use an `internal` package matching existing conjure code, however it is a bit odd to implement an `internal` in a separate project (eventually conjure-java-runtime).

## Body Serializers

At the moment our body serializers and deserializers are modeled as `SerDe.serialize(value, data)` and `SerDe.deserialize(type, data)` which always requires runtime mapping per request from `type` or `value` to a serialization mechanism.
We should instead provide `SerDe.serializerFor(type)` and `SerDe.deserializerFor(type)` the result of which are stored as fields on each handler.

We can do this in a backwards compatible way later by delegating to `ser/deFor(type).ser/de(data)`, but we can avoid the shims if we implement at least the API portion in this change.

rubenfiszel:
@cakofony 

### Packages proposal

`conjure-undertow-lib -> conjure-undertow-api`

Implemented only by generated code: `conjure-undertow-api.generated-api`
Implemented only by the runtime, and consumed exclusively by generated code: `conjure-undertow-api.runtime-api`

### Body Serializers

I propose the following scheme:

```
interface Serializer<I> {
  Bytes serialize(I input, String contentType)
}

interface Deserializer<O> {
  O deserialize(Bytes bytes, String contentType)
}

interface SerDe {
  Serializer<I> serializerFor(Class<I> class)
  Deserializer<O> deserializerFor(Class<O> class)
}
```

such that generated code can be 
```
class MyHandler implements HttpHandler, Endpoint {
    private  final Serializer<QueryParamXType> queryParamXSerializer;
    MyHandler(UndertowRuntime runtime, ..) {
       queryParamXSerializer = runtime.serde().serializerFor(QueryParamXType.class)
       ...
   }
... 
}
```

And then SerializerRegistry would become (if still relevant) an implementation detail of the runtime..

What do you think?

carterkozak:
Yep, that's basically the design I'm thinking. Regarding packaging, changing modules will happen in the future, but for now I'm just thinking about changing java packages inside of the `conjure-undertow-lib` module.

carterkozak:
Body Serializers proposal is here: #265

carterkozak:
I'm happy with the state of this PR. Once this merges we will not allow breaking changes, so if you see anything that looks odd please flag here. I would appreciate another round of review before we merge. Thanks!

rubenfiszel:
@cakofony Another way to reduce boilerplate:
Have an abstract class Default/AbstractEndpoint which extends Endpoint and take all getters (except the HttpHandler) in the constructor. 
Make all handlers extends Default/AbstractEndpoint instead of Endpoint directly and set the parameters in the constructor through the super call.

rubenfiszel:
otherwise, LGTM

markelliot:
lgtm. can always make the switch to AbstractEndpoint or whatever later as it's an internal concern to the code-gen.

markelliot:
only other nit might be: `EmptyPathServiceEndpoints` -> `UndertowEmptyPathServiceHandler` or maybe just `UndertowEmptyPathService` (something that corresponds more to it being an implementation of `UndertowService`). The `*Endpoints` thing is probably fine, too.

carterkozak:
I'm happy with the `*Endpoints` name. Generated *Endpoints type is used as a static utility class and cannot be instantiated directly.

Agree that we shouldn't add the abstract implementation at this point, SLOC is likely only an issue for IDE indexing, and we can make changes if that becomes problematic. I like that our generated code is clear about what exactly it does, and the abstract class would have (slightly) more heap overhead at runtime (not that we should index on that).

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/PlainSerDe.java
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * These utilities are used to parse HTTP path, query, and header parameter values.
+ */
+public interface PlainSerDe {
+
+    // TODO(ckozak): Write javadoc
+
+    BearerToken deserializeBearerToken(String in);
+
+    BearerToken deserializeBearerToken(Iterable<String> in);
```
iamdanfox: I don't really understand the signature of this... seems weird that we're asking implementers of this to make decisions about which of the iterable of strings to pick?  Why not just assert there's only one and pass to the `BearerToken  deserializeBearerToken(String)` method?

Also why are others Nullable but not this one?

carterkozak: That is logic we would have to implement in generated code, it’s cleaner to move that burden to code we write in the runtime and reuse.

Re nullable: that’s probably a bug, I will verify

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/BodySerDe.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.common.reflect.TypeToken;
+import io.undertow.server.HttpServerExchange;
+import java.io.IOException;
+import java.io.InputStream;
+
+/** Request and response Deserialization and Serialization functionality used by generated code. */
+public interface BodySerDe {
+
+    /** Creates a {@link Serializer} for the requested type. Serializer instances should be reused. */
+    <T> Serializer<T> serializer(TypeToken<T> type);
```
iamdanfox: I've just noticed that `TypeToken` comes from Guava and is actually annotated `@Beta`, which I think means it might change across guava versions... 

Do you think it's worth lifting the TypeCapture#capture() method into our own lib? Or shall we just run this as-is, accepting the small risk that we'd need to update our interfaces if guava decides to break us?

carterkozak: @uschi2000 Curious of your thoughts on using guava TypeToken vs defining our own, I think this is a relic from your prototype/dialogue.
We use guava collections in generated code, but they're not `@Beta`, and not core api.

I think we might as well define our own to avoid awkward breaks in the future, I'll PR into this branch to try it out.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
```diff
@@ -0,0 +1,179 @@
...
+            this.encodings = encodings.stream()
+                    .map(encoding -> new EncodingSerializerContainer<>(encoding, token))
+                    .collect(ImmutableList.toImmutableList());
+            this.defaultEncoding = this.encodings.get(0);
+        }
+
+        @Override
+        public void serialize(T value, HttpServerExchange exchange) throws IOException {
+            Preconditions.checkNotNull(value, "cannot serialize null value");
+            EncodingSerializerContainer<T> container = getResponseSerializer(exchange);
```
iamdanfox, line 98: tiny bit subtle that the second param of `Serializer#serialize(T, HttpServerExchange)` is only intended for people to _write_ data, but here we're _reading_ the original request's content-type and changing behaviour based on it. 

Don't have a better proposal here - just something that made me look twice

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/PlainSerDe.java
carterkozak: Fixed

iamdanfox:
:+1: New names and package separation looks fantastic - I really like how all the conjure-undertow-lib interfaces precisely capture just the language concepts of Conjure, and everything else ends up separated!

Are we gonna tag this as conjure-java 3.0?

dansanduleac:
👍

carterkozak:
Getting rid of TypeToken: https://github.com/palantir/conjure-java/pull/273 (targeted into this branch)

carterkozak:
re 4.0.0 do we want to tag the break with this change, or when we move the runtime to `conjure-java-runtime`? Perhaps 4, than a relatively quick 5 for that change is safest.

PR #261, opened 2019-03-04
[break][Undertow] Delete deprecated APIs, and make breaking improvements

carterkozak:
* Removes deprecated utilities which have been migrated to `ServiceContext` in #261 
* `EndpointRegistry` is no longer fluent, preventing bugs where wrapped registry lambdas erroneously return a delegate registry rather than themselves.
* the `Service` interface and `Registerable` interface have been combined. `Service` provides a single `void register(ServiceContext context, EndpointRegistry registry)` and `Registerable has been removed, resulting in a simpler API.


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Service.java
```diff
@@ -17,12 +17,13 @@
...
 
 /**
- * Creates a {@link Registrable} which may be registered with a web server. The server is responsible
- * for providing a {@link ServiceContext} allowing API implementors to add APIs using
+ * Registers a service interface with a web server. The server is responsible
+ * for providing a {@link ServiceContext} and {@link EndpointRegistry} allowing
+ * API implementors to add APIs using
  * <code>server.api(MyServiceEndpoints.of(myServiceImpl)</code>.
  */
 public interface Service {
```
carterkozak, line 25: If we want to rename to `UndertowService`, this is the time to do it (assuming we decide to take a break here). @iamdanfox thoughts?

carterkozak:
Note that this is targeted into #260, not directly into develop

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoints.java
```diff
@@ -23,11 +22,11 @@ public static Service of(UndertowEmptyPathService delegate) {
     }
 
     @Override
-    public Registrable create(ServiceContext context) {
-        return new EmptyPathServiceRegistrable(context, delegate);
+    public void register(ServiceContext context, EndpointRegistry registry) {
```
markelliot, line 25: what's the rationale behind passing in an EndpointRegistry instead of returning an EndpointRegistry?

carterkozak: Many services are registered to a single EndpointRegistry, and we have no guarantee that each service has a unique prefix.

markelliot: so an alternative approach would be returning an "EndpointList" or having the ability to merge many EndpointRegistry objects. Why do the conditions you mention demand that we invert the control here?

I think, generally, I'd prefer to see us use a more declarative approach, rather than passing in an object that gets configured, we ask this class to produce required metadata and then something external to it uses that information to do the configuring.

carterkozak: That is exactly what `Registerable` did. Perhaps renaming `Registerable` to `EndpointList` would be better, though you noted that you didn't like the double-initialization of the declarative model before.

markelliot: Actually it was the same objection: Registrable returned a thing that accepts an EndpointRegistry and performs registration of endpoints. I'd like to return a thing that that an EndpointRegistry would iterate over to perform endpoint registration.

```
interface Registrable {
    Iterable<EndpointDefinition> endpoints();
}

interface EndpointDefinition {
    Endpoint endpoint();
    Handler handler();
}
```

Or, we could simplify a bit:

```
interface Service {
        Iterable<EndpointDefinition> endpoints(ServiceContext context);
}
```

carterkozak: Ack, I will make the second change.

carterkozak: @rubenfiszel Thoughts on including `HttpHandler` in `Endpoint` to avoid the `EndpointDefinition` tuple type?

iamdanfox: @cakofony alternatively perhaps it just returns a `Map<Endpoint, HttpHandler>`? (but no strong preference)

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/SerializerRegistry.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
```
rubenfiszel, line 2: @dfox argued that our legal dept recommend to not do this.

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoints.java
carterkozak: I'm not a huge fan of using maps in the API -- that could silently lose data if we attempt to combine two results, where duplicate Endpoints in a list would fail upon registration to the registry, which is cleaner behavior.

Currently our `Endpoint` is just the metadata describing where the endpoint can be reached, the `HttpHandler` _is_ the endpoint. The more I think about this the better I like combining the handler into our `Endpoint` object.
Endpoint registry becomes `void register(Endpoint endpoint)` rather than `register(Endpoint, HttpHandler)`

markelliot: can also consider having EndpointHttpHandler and putting this metadata into the HttpHandler implementations, and returning a list of those.

rubenfiszel: I am in favor of the proposal of @iamdanfox: `Map<Endpoint, HttpHandler>`. Endpoint is nice to have as a pure metadata tuple and is perfectly suited as a key.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointRegistry.java
```diff
@@ -27,6 +27,6 @@
  */
 public interface EndpointRegistry {
 
-    EndpointRegistry add(Endpoint description, HttpHandler handler);
+    void add(Endpoint description, HttpHandler handler);
```
rubenfiszel, line 31: One aspect I like less about this is that now EndpointRegistry cannot be implemented in an immutable fashion anymore.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/SerializerRegistry.java
carterkozak: Huh, spotlessApply did this

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyPathServiceEndpoints.java
carterkozak: I prefer using a tuple type over Map, the hash and equality methods on `Endpoint` shouldn't have any bearing on this functionality.

carterkozak:
Closing this PR and making changes on the base #260 PR

PR #265, opened 2019-03-06
[Undertow] Refactor Serializers API

carterkozak:
Each endpoint only serializes and desrializes at most one type, so we provide type information to each endpoint at construction time. This allows implementations to potentially avoid doing type lookup work at runtime.

## Overview

The old `Serializer` interface is renamed to `Encoding`. Serializer implies formatting a structured object, where an encoding has additional metadata and supports serialization and deserialization.

The `lib` package provides a `Serializer<T>` and `Deserializer<T>` type which are eagerly initialized when endpoints are constructed, more accurately matching the conjure definition where there's no reason to allow dynamic type information inside of a single endpoint.


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Encoding.java
```diff
@@ -0,0 +1,68 @@
...
+
+    interface Serializer<T> {
+
+        /**
+         * Serializes the given object and writes the serialized representation to the given output stream.
+         * Implementations must not close the stream. Inputs must never be null.
+         */
+        void serialize(T value, OutputStream output) throws IOException;
+
+    }
```
carterkozak, line 77: This change introduces both an `Encoding.Serializer/Deserializer` (runtime only) as well as `Serializer/Desrializer` in the lib. I considered naming them differently, but decided that the `Encoding` outer class provides enough context to differentiate the two, and the terms Serializer and Deserializer make it easy to understand their purpose.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/SerDe.java
```diff
@@ -36,15 +36,15 @@
 /** Serialization functionality consumed by generated code. */
 public interface SerDe {
```
rubenfiszel: Could we maybe extract the param deserializers into a ParamDeserializers interface and make SerDe extends it ?

carterkozak: I don't think that buys us a lot, the logical extreme is to put everything back on the `ConjureRuntime` interface. I'd prefer to have as few separate interfaces in the lib module as possible, which will make it easier to think about backcompat.

rubenfiszel: If we were to go on an extreme with this, you could have BodySerDe, ParamSerDe and StaticTypeSerDe and have SerDe extends all three.

carterkozak: Yep, weighted against adding more interfaces to `lib` I don't think it's super helpful, but I don't have a strong opinion if you disagree.

rubenfiszel: Personally I don't really mind more interface as long as they allow better compartmentalization.
Maybe it's just missing more explicit javadoc (you could potentially have an handler in the javadoc that use all feature of SerDe) but it is not trivial to follow on a first read.
I feel like intuitively reading this you would think there is a lot of factorization that could be done with all those deserialize methods (even though it's not the case). So some comments about some of the design choice would help.

rubenfiszel (approved):
I like very much the resulting architecture. I think overall it is missing a bit of high-level docs but it is very clean and once you understand the design and generated code it is easy enough to convince yourself this is well optimized at runtime and without evident warts.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/EncodingRegistry.java
```diff
@@ -0,0 +1,185 @@
...
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.util.HeaderValues;
+import io.undertow.util.Headers;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+/** Orchestrates serialization and deserialization of response and request bodies. */
+final class EncodingRegistry {
```
rubenfiszel: Another design possible instead of `(Encoding default, List<Encoding> encodings)` is `(Encoding default, BiMap<Format, Encoding> formatToEncoding, ContentTypeString -> Format contentTypeToFormat) ` and have the encoding class not have to implement: 
```
    String getContentType();

    boolean supportsContentType(String contentType);
```

Personally, I find the latter easier to follow since it is a bit more explicit that an encoding is tied to a serde mechanism like json or cbor and that the content type mapping is a separate concern.

conjure-java-core/src/integrationInput/java/com/palantir/product/EteBinaryServiceEndpoints.java
```diff
@@ -46,11 +47,12 @@ public static UndertowService of(UndertowEteBinaryService delegate) {
 
         private final UndertowEteBinaryService delegate;
 
-        private final TypeToken<InputStream> bodyType = new TypeToken<InputStream>() {};
+        private final Deserializer<InputStream> deserializer;
 
         PostBinaryEndpoint(UndertowRuntime runtime, UndertowEteBinaryService delegate) {
             this.runtime = runtime;
             this.delegate = delegate;
+            this.deserializer = runtime.serde().deserializer(new TypeToken<InputStream>() {});
```
carterkozak: bug: this is unnecessary here

conjure-java-core/src/integrationInput/java/com/palantir/product/EteBinaryServiceEndpoints.java
```diff
@@ -46,11 +47,12 @@ public static UndertowService of(UndertowEteBinaryService delegate) {
 
         private final UndertowEteBinaryService delegate;
 
-        private final TypeToken<InputStream> bodyType = new TypeToken<InputStream>() {};
```
carterkozak, line 49: was unused, should never have been here

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/SerDe.java
carterkozak: edit: commented on the wrong thread

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/EncodingRegistry.java
carterkozak: this time on the right thread:

I think we can punt this until later, the `Encoding` interface is public, but only to allow us to test support for other content types outside of this repository. It shouldn't be considered a major break for the runtime module to make changes to Encoding because implementors must understand the risk. I'll add that to the javadoc :-)

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/SerDe.java
carterkozak: I think you're right on this one. perhaps we should move the query/path/header deserializaton methods to a `Plain` interface since these are values which use the conjure Plain format. e.g. `runtime.plain().deserializeString(in)`, thoughts?

carterkozak: alternatively add `Plain` to the method names, e.g. `runtime.serde().deserializePlainString(in)`
It feels odd not to put those on the `SerDe` interface

rubenfiszel: I would prefer `runtime.serde().plain().deserializeString(in)` and make `Plain` a field of handlers.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/EncodingRegistry.java
rubenfiszel: Fair. On the other hand, I don't think we will want to break later with this since it's a rather minor refactor.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/SerDe.java
carterkozak: We can continue this discussion in #260 (target of this PR) since this PR doesn't modify SerDe.<plain deserializer functions>

PR #273, opened 2019-03-08
Replace guava TypeToken with an internal conjure TypeMarker

carterkozak:


iamdanfox:
:+1: This is a lot more reassuring than people possibly relying on the 1400 line https://github.com/google/guava/blob/master/guava/src/com/google/common/reflect/TypeToken.java

PR #276, opened 2019-03-09
[improvement] conjure-undertow deserializes collections to guava immutable collections

### LINKED ISSUE #272: Server request body parameters of types map, set, and list should not be mutable
## What happened?

Using both conjure-jersey and conjure-undertow server generators I can modify request body parameter with type `List`, `Set`, or `Map`in server implementations.

Human defined implementations of conjure interfaces should only receive immutable list, set, and map objects. Note I am not suggesting that we should change the signature of service methods, only the implementation we pass.
If we choose to use the guava ImmutableList, ImmutableSet, and ImmutableMap, we will also gain validation that null values are not allowed (see #105)

carterkozak:
No change to implemented interfaces, this simply prevents inputs
from being modified.

fix #272

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EteServiceEndpoints.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EteServiceEndpoints.java
@@ -1,6 +1,8 @@
 package com.palantir.product;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.palantir.conjure.java.lib.SafeLong;
 import com.palantir.conjure.java.undertow.lib.BinaryResponseBody;
 import com.palantir.conjure.java.undertow.lib.Deserializer;
@@ -25,6 +27,7 @@ import java.util.Deque;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import javax.annotation.Generated;
 
 @Generated("com.palantir.conjure.java.services.UndertowServiceHandlerGenerator")
@@ -65,7 +68,8 @@ public final class EteServiceEndpoints implements UndertowService {
                 new EnumQueryEndpoint(runtime, delegate),
                 new EnumListQueryEndpoint(runtime, delegate),
                 new OptionalEnumQueryEndpoint(runtime, delegate),
-                new EnumHeaderEndpoint(runtime, delegate));
+                new EnumHeaderEndpoint(runtime, delegate),
+                new ComplexNestedWrappersEndpoint(runtime, delegate));
     }
 
     private static final class StringEndpoint implements HttpHandler, Endpoint {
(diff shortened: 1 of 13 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -291,22 +293,33 @@ private CodeBlock endpointInvocation(EndpointDefinition endpointDefinition, List
...
+                // avoid compilation errors without sacrificing immutability.
+                TypeName handlerType = mappers.handlerRequest().getClassName(bodyParam.getType()).box();
+                TypeName serviceType = mappers.request().getClassName(bodyParam.getType()).box();
+                boolean requiresCast = requiresForcefulCast(handlerType, serviceType);
+                if (requiresCast) {
+                    code.add("@$1T($2S)", SuppressWarnings.class, "unchecked");
+                }
+                code.addStatement(requiresCast
+                                ? "$1T $2N = ($1T) (Object) $3N.deserialize($4N)" : "$1T $2N = $3N.deserialize($4N)",
+                        serviceType, bodyParam.getArgName().get(), DESERIALIZER_VAR_NAME, EXCHANGE_VAR_NAME);
```
carterkozak, line 311: This piece is more complicated than I'd like, however I'd rather take this complication in a conjure generator over debugging problems caused by mutable state.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowRequestBodyClassNameVisitor.java
```diff
@@ -55,7 +50,8 @@ public TypeName visitPrimitive(PrimitiveType primitiveType) {
 
     @Override
     public TypeName visitOptional(OptionalType optionalType) {
-        // TODO(ckozak): Support this? I don't know that jersey does.
+        // optional<binary> requests are not allowed, unlike responses there is no way to differentiate
+        // present 0-byte binary from empty optional.
```
carterkozak, line 54: Updated comment, the mentioned case is already validated by the conjure parser.

conjure-java-core/src/integrationInput/java/com/palantir/product/EteServiceEndpoints.java
```diff
@@ -1261,4 +1265,75 @@ public HttpHandler handler() {
...
+                                    new TypeMarker<Optional<Map<String, List<Set<String>>>>>() {});
+        }
+
+        @Override
+        public void handleRequest(HttpServerExchange exchange) throws IOException {
+            AuthHeader authHeader = runtime.auth().header(exchange);
+            @SuppressWarnings("unchecked")
+            Optional<Map<String, List<Set<String>>>> body =
+                    (Optional<Map<String, List<Set<String>>>>)
+                            (Object) deserializer.deserialize(exchange);
```
carterkozak, line 1304: Nasty cast from the code highlighted above.

carterkozak:
Can just do this instead:
```java
                .registerModule(new SimpleModule("immutable")
                        .addAbstractTypeMapping(List.class, ImmutableList.class)
                        .addAbstractTypeMapping(Map.class, ImmutableMap.class)
                        .addAbstractTypeMapping(Set.class, ImmutableSet.class))
```

PR #277, opened 2019-03-10
Don't warn about immutable bytes unless objects are generated

carterkozak:
The binary type is not allowed in headers/query/path parameters,
and streaming types are used for the request and response
bodies. The flag only modifies output of object generation.


markelliot (commented):
Ideally we would only emit the warning when binary is used in an object definition.

carterkozak:
I'd prefer to associate the warning with an infrastructure change (conjure generator upgrade or addition of conjure to a new project) than a change which adds a new endpoint.

iamdanfox:
:+1: Not convinced this will really affect many projects (as I don't know of any use cases that just generate jersey), but associating with a conjure-generator upgrade seems like a fair enough justification

PR #281, opened 2019-03-11
[improvement] The conjure-undertow runtime deserializes immutable collections

carterkozak:
conjure-undertow endpoints which expect List, Map, and Set types (including nested
in optional, and nested in alias types) receive guava immutable
implementations.

For context, see https://github.com/palantir/conjure-java-runtime/pull/973

### DIFF
--- a/conjure-java-server-verifier/src/test/resources/ignored-test-cases.yml
+++ b/conjure-java-server-verifier/src/test/resources/ignored-test-cases.yml
@@ -27,3 +27,7 @@ server:
       - 'null'
     getOptionalAnyAliasExample:
       - 'null'
+    getListAnyAliasExample:
+      - '[null, 0, "content", true, [1,2,3], {"key":3}]'
+    getSetAnyAliasExample:
+      - '[null, 0, "content", true, [1,2,3], {"key":3}]'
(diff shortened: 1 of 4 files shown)


markelliot:
wondering if it'd make sense to first get a a handle on how much this might impact people's existing serde before cutting over?

carterkozak:
@markelliot This is in response to a conversation on https://github.com/palantir/conjure-java-runtime/pull/973
I should have included that in the description for this PR

markelliot:
even so, if we want to encourage users to switch from conjure-jaxrs to conjure-undertow we probably need to understand how frequently nulls show up in their collections so they have a handle on the risk they're taking in performing the transition

carterkozak:
Sure, that's the tradeoff noted in the conjure-java-runtime comment. Conjure-undertow already provides additional validation which aligns well with guava collection validation, for example jersey allows null parameters and return values, but this generator does not.
I appreciate that we want to limit the difference in behavior between the two, and enable strictness when we're certain it doesn't break existing workflows, but I fear that will never happen.

carterkozak:
Putting this on hold until we find a path forward, it would be regrettable if we weren't able to provide this safety to all conjure consumers.

markelliot:
I'd love for us to add a small amount of logging to the conjure-jaxrs stuff to assess impact of the change -- I'm in favor of the direction/want the benefit, but without understanding the impact it seems like it may become a critical blocker/create critical failures.

carterkozak:
Proposal here: https://github.com/palantir/conjure-java-runtime/pull/979

PR #284, opened 2019-03-15
[fix] Log error name in exception handler

pkoenig10:
## Before this PR
Error names are not logged when handling errors from undertow endpoints.

## After this PR
Error names are logged when handling errors from undertow endpoints.

This was a small regression when switching to undertow endpoints. This behavior was added to the jersey exception mappers in https://github.com/palantir/conjure-java-runtime/pull/827.


carterkozak (approved):

ferozco:
👍

PR #293, opened 2019-03-23
[fix] JDK 8 binary compatibility

schlosna:
## Before this PR

`NoSuchMethodError` when `conjure-java` is built on JDK 9+ and run on JDK 8, for example:

```
java.lang.NoSuchMethodError: java.nio.ByteBuffer.rewind()Ljava/nio/ByteBuffer;
	at com.palantir.foo.bar.Key$Builder.encodedKey(Key.java:155)
```

## After this PR
==COMMIT_MSG==

Generate JDK 8 compatible binary types

JDK 9 added covariant return types to ByteBuffer (and other *Buffer)
classes for methods including flip, rewind, clear, mark, limit, reset,
position. Unfortunately, building against JDK 9+ with a target of JDK 8 can lead to `java.lang.NoSuchMethodError: java.nio.ByteBuffer.rewind()Ljava/nio/ByteBuffer;`

See [lucene-dev thread on this issue](https://mail-archives.apache.org/mod_mbox/lucene-dev/201503.mbox/%3CCAOdYfZWTWzBq9=kc-ECvi9_jLQDuS7_k7wcX+Uem2=y9VvRFKw@mail.gmail.com%3E)

This workaround avoids such problems by explicitly casting to the base `Buffer` type before rewinding.

Use [JEP 247](https://openjdk.java.net/jeps/247) to specify JDK target release on JDK 9+ in gradle compiler arguments.

See also [gradle `releaseCompatibility` RFC](https://github.com/gradle/gradle/issues/2510) 

==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,6 +70,11 @@ subprojects {
     tasks.withType(JavaCompile) {
         options.errorprone.errorproneArgs += ['-Xep:PreferSafeLoggingPreconditions:OFF', '-Xep:PreferSafeLoggableExceptions:OFF']
         options.compilerArgs += ['-Werror']
+        if (JavaVersion.current() > JavaVersion.VERSION_1_8) {
+            // Link against JDK 8 APIs per https://openjdk.java.net/jeps/247
+            // see also gradle `releaseCompatibility` RFC https://github.com/gradle/gradle/issues/2510
+            options.compilerArgs += ['--release', '8']
+        }
     }
 
     // Run `./gradlew test -Drecreate=true` to recreate all the expected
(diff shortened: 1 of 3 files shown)


ferozco (commented):
Does it make sense to only make the code changes here, and pickup the gradle changes through baseline?

iamdanfox:
:+1:

ferozco:
👍

schlosna:
@ferozco & @iamdanfox -- do we want to make a similar change to https://github.com/palantir/gradle-conjure or just apply the release compatibility plugin?

iamdanfox:
Would like to push baseline-release-compatibility  - i just merged this as a stopgap so that we can give people better codegen!

PR #294, opened 2019-03-25
[break] Consistently unbox external type imports for boxed primitives 

carterkozak:
Note: We do not recommend using external type imports, they bring
significant risk. However while they are supported we should avoid
generating code that would not pass a code review.

## Before this PR
Java primitive external type imports were previously handled
sporadically.

## After this PR
Now the type DefaultClassNameVisitor handles
unboxing consistently across the project.

==COMMIT_MSG==
External type imports for boxed java primitives are unboxed in generated code.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/ExternalLongExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/ExternalLongExample.java
@@ -4,10 +4,12 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSetter;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.palantir.conjure.java.lib.internal.ConjureCollections;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -20,12 +22,16 @@ public final class ExternalLongExample {
 
     private final Optional<Long> optionalExternalLong;
 
+    private final List<Long> listExternalLong;
+
     private volatile int memoizedHashCode;
 
-    private ExternalLongExample(long externalLong, Optional<Long> optionalExternalLong) {
-        validateFields(optionalExternalLong);
+    private ExternalLongExample(
+            long externalLong, Optional<Long> optionalExternalLong, List<Long> listExternalLong) {
+        validateFields(optionalExternalLong, listExternalLong);
         this.externalLong = externalLong;
         this.optionalExternalLong = optionalExternalLong;
+        this.listExternalLong = Collections.unmodifiableList(listExternalLong);
     }
 
     @JsonProperty("externalLong")
@@ -38,6 +44,11 @@ public final class ExternalLongExample {
         return this.optionalExternalLong;
     }
 
+    @JsonProperty("listExternalLong")
+    public List<Long> getListExternalLong() {
+        return this.listExternalLong;
+    }
+
     @Override
     public boolean equals(Object other) {
         return this == other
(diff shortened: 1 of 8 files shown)


ferozco:
Although this is technically a break to code gen, I don't think in practice anyone would be effected so it seems reasonable to move forward with this

ferozco:
@iamdanfox 	@dansanduleac any concerns with this?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceInterfaceGenerator.java
```diff
@@ -83,8 +83,7 @@ private MethodSpec generateServiceInterfaceMethod(
 
         ServiceGenerator.getJavaDoc(endpointDef).ifPresent(content -> methodBuilder.addJavadoc("$L", content));
 
-        endpointDef.getReturns().ifPresent(type -> methodBuilder.returns(
-                UndertowTypeFunctions.unbox(returnTypeMapper.getClassName(type))));
```
iamdanfox, line 87: Should be safe to delete this `unbox` method too now right?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/DefaultClassNameVisitor.java
```diff
@@ -153,7 +153,8 @@ public TypeName visitReference(com.palantir.conjure.spec.TypeName type) {
     @Override
     public TypeName visitExternal(ExternalReference externalType) {
         String conjurePackage = externalType.getExternalReference().getPackage();
-        return ClassName.get(conjurePackage, externalType.getExternalReference().getName());
+        ClassName typeName = ClassName.get(conjurePackage, externalType.getExternalReference().getName());
+        return typeName.isBoxedPrimitive() ? typeName.unbox() : typeName;
```
iamdanfox, line 157: :+1: this seems like a reasonable short-term workaround, especially because it doesn't really have downsides for users who don't have Long external imports.

conjure-java-core/src/integrationInput/java/com/palantir/product/ExternalLongExample.java
```diff
@@ -0,0 +1,224 @@
...
+        this.listExternalLong = Collections.unmodifiableList(listExternalLong);
+    }
+
+    @JsonProperty("externalLong")
+    public long getExternalLong() {
+        return this.externalLong;
+    }
+
+    @JsonProperty("optionalExternalLong")
+    public Optional<Long> getOptionalExternalLong() {
```
iamdanfox, line 43: @cakofony do you envisage wanting to change this to be an OptionalLong at some point? (I know we do this for conjure-native integers already in OptionalIntegerExample.java).

Or is that codepath something you'd put on the gotham fork?

carterkozak: Yes, we do that in certain versions of this generator already, however I think that may be a bit too special for external type imports, and is likely a better fit for an integer64 type once that is implemented.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceInterfaceGenerator.java
carterkozak: I thought I had, oops!

carterkozak: Updated

iamdanfox:
This is an ABI-break in our generated code, as getters for java.long.Long used to return boxed values and now return unboxed ones, but it's not a wire break.

We just re-crawled `internal-conjure-release` and verified that nobody in Foundry has an external type defined for `java.lang.Long`. This suggests that we won't hit the nasty diamond dependency case from this change, so it's probably OK to sneak it into a release without a major rev.

:+1:

iamdanfox:
Actually hang on - looks like our searches missed a couple...

Using our `foundry/conjure-local-template` repo, I ran `./gradlew copyIr` and `rg --no-ignore -U -l "externalReference.*\n.*Long"` and it shows that exactly two published IR definitions use external long imports, both of which are related to Gotham.

Likely still don't need to major rev, but would like to phrase the PR as '[break]' just in case!

PR #309, opened 2019-03-29
[feature] JAX-RS client Conjure generator

carterkozak:
**Why**
Many use cases are better suited for blocking clients. There is no
doubt that asynchronous clients can be more powerful, however
they're often both unnecessary and dangerous. Threads are hard,
and it's best to avoid thinking about them until we need to.

The existing Jersey generator is sufficient in many cases,
but it's entirely broken for streaming binary data. Because
the Jersey generator outputs code for both clients and servers,
it must be a jack of all trades and master of none. For example,
there's no reason for server interfaces to scramble argument
order or provide default methods, but they do so because it's
helpful when the interface is reused by clients. There's also
no reason for client interfaces to return a server-side
`StreamingOutput` from binary methods because the type is
unhelpful to clients. By decoupling client and server
interfaces we allow ourselves to generate more accurate
code for both.

**What**
This change adds a `--jaxRsClient` flag to the CLI to generate
client-specific interfaces. Our JaxRsClient interfaces differ
in a few ways:
1. Generated interfaces names are suffixed with `JaxRsClient`
   to differentiate them from Jersey services and avoid classpath
   conflicts.
2. `returns: binary` results in methods which return InputStream
3. `returns: optional<binary>` results in methods which return
   `Optional<InputStream>`
4. The generator ignores `--requireNotNullAuthAndBodyParams`
   because feign doesn't use doesn't use
   `javax.validation.constraints`

**Future Work**
I plan to update `conjure-java-runtime` JaxRsClients to avoid fully
buffering binary request and response bodies on heap. This is not
a blocker for the generator because we have decided that buffering
binary request bodies is acceptable, so we can temporarily allow
the same behavior for responses.

==COMMIT_MSG==
[feature] JAX-RS client Conjure generator
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JaxRsClientGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JaxRsClientGenerator.java
@@ -1,0 +1,48 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.services;
+
+import com.palantir.conjure.java.FeatureFlags;
+import com.palantir.conjure.spec.ConjureDefinition;
+import com.squareup.javapoet.ClassName;
+import com.squareup.javapoet.JavaFile;
+import com.squareup.javapoet.ParameterizedTypeName;
+import java.io.InputStream;
+import java.util.Optional;
+import java.util.Set;
+
+public final class JaxRsClientGenerator implements ServiceGenerator {
+
+    private final JaxRsServiceFactory delegate;
+
+    public JaxRsClientGenerator(Set<FeatureFlags> featureFlags) {
+        this.delegate = new JaxRsServiceFactory(
+                getClass(),
+                featureFlags,
+                serviceName -> serviceName + "JaxRsClient",
+                // Clients return InputStream and Optional<InputStream>
+                ClassName.get(InputStream.class),
+                ParameterizedTypeName.get(Optional.class, InputStream.class),
+                // Feign clients do not validate based on javax.validation.constraints
+                false);
+    }
+
+    @Override
+    public Set<JavaFile> generate(ConjureDefinition conjureDefinition) {
+        return delegate.generate(conjureDefinition);
+    }
+}
--- a/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
+++ b/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
@@ -27,7 +27,6 @@ import com.palantir.conjure.java.services.Retrofit2ServiceGenerator;
 import com.palantir.conjure.java.services.UndertowServiceGenerator;
 import com.palantir.conjure.java.types.ObjectGenerator;
 import com.palantir.conjure.spec.ConjureDefinition;
-import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.File;
 import java.io.IOException;
 import java.util.List;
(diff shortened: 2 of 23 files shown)


uschi2000:
Instead of spending points here, @cakofony , how do you feel about pushing dialogue over the line?

carterkozak:
I agree that dialogue is our strongest client generator for the future, and I would enjoy working on it. However, I have a large application which remains on conjure 1.10.3 that I need to prioritize upgrading to latest before I have cycles to commit to dialogue.

We can match the structure of dialogue blocking client interfaces to jaxrs client interfaces in order to quickly replace the implementation with minimal code churn when we're ready to do so.

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -127,31 +133,31 @@ public void run() {
...
+                            .emit(conjureDefinition, config.outputDirectory());
                 }
                 if (config.generateUndertow()) {
-                    undertowGenerator.emit(conjureDefinition, config.outputDirectory());
+                    new UndertowServiceGenerator(config.featureFlags())
+                            .emit(conjureDefinition, config.outputDirectory());
                 }
             } catch (IOException e) {
-                throw new RuntimeException("Error parsing definition", e);
+                throw new SafeRuntimeException("Error parsing definition", e);
```
rubenfiszel: What is the purpose of safe exceptions here?

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -127,31 +133,31 @@ public void run() {
...
+                    if (!config.featureFlags().contains(FeatureFlags.UseImmutableBytes)) {
+                        System.err.println("[WARNING] Using deprecated ByteBuffer codegen, please enable the "
+                                + "--useImmutableBytes feature flag to opt into the preferred implementation");
+                    }
+                    new ObjectGenerator(config.featureFlags()).emit(conjureDefinition, config.outputDirectory());
+                }
+                if (config.generateJaxRsClient()) {
+                    new JaxRsClientGenerator(config.featureFlags()).emit(conjureDefinition, config.outputDirectory());
                 }
                 if (config.generateJersey()) {
```
rubenfiszel, line 147: thoughts about adding ServiceGenerator to a list based on the flags and then `list.foreach(ServiceGenerator::emit)` ?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JaxRsClientGenerator.java
```diff
@@ -0,0 +1,48 @@
...
+import com.squareup.javapoet.ParameterizedTypeName;
+import java.io.InputStream;
+import java.util.Optional;
+import java.util.Set;
+
+public final class JaxRsClientGenerator implements ServiceGenerator {
+
+    private final JaxRsServiceFactory delegate;
+
+    public JaxRsClientGenerator(Set<FeatureFlags> featureFlags) {
```
rubenfiszel, line 32: seems a bit awkward to use composition in this manner.
Thoughts about abstract JaxRsGenerator and two subclass JaxRsClientGenerator and JaxRsServiceGenerator ?

rubenfiszel (approved):
2 small nits, otherwise lgtm.
I agree with the purpose of the PR and that even though we should prioritize dialogue, this still seem very valuable considering it require minimal work and give us more flexibility.

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
carterkozak: We use them elsewhere, and it doesn't produce a different message in this case. I prefer to consistently use safe exception types across our codebases so we can avoid accidentally using the wrong ones where it matters.

In practice this one makes no difference.

rubenfiszel: Mixed feelings about this. Safe imho is a logging concept and should not be used in non-relevant contexts.

carterkozak: Reverting the exception type change because it's not relevant to this PR.

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
carterkozak: I think there's something clever here using a `BiConsumer<Definition,Output>` and chaining via `andThen(ServiceGenerator::emit)`, but I think the code is easier to follow as is.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JaxRsClientGenerator.java
carterkozak: I wanted stronger separation between the `JaxRsGenerator` and `JerseyServiceGenerator` classes. I find abstract classes more difficult to follow, so I opted for a package-private final shared utility.

PR #324, opened 2019-04-04
[proposal][wip] conjure-undertow async request processing (Callback)

carterkozak:
Target generated interface:

```java
interface MyService {
  void getFoo(AuthHeader authHeader, AsyncContext<Foo> context);
}
```

impl:

```java
class MyResource {
  @Override
  void getFoo(AuthHeader authHeader, AsyncContext<Foo> context) {
    context.complete(Foo.builder().build());
  }
}
```

Where the blocking implementation would take the form:

```java
interface MyService {
  Foo getFoo(AuthHeader authHeader);
}
```


uschi2000:
Is this something you need urgently, @carterkozak ? I would like to let the general conjure and undertow dust settle a bit before we introduce more stuff.

carterkozak:
I don't need it this month, but support for some kind of asynchronous request processing will block using generators in this repository. This PR is mostly unimplemented and meant as a demonstration of one of the options highlighted in my design document. I'd like to build consensus on the approach we want to take before it becomes the only blocker to avoid rushing into design decisions we may regret.

Happy to discuss the specific uses offline.

PR #334, opened 2019-04-11
[improvement] Add a feature flag to disable parameter sorting in service method generation

dsd987:
## Before this PR
Service methods from the service method generators will always sort parameters to help with backwards compatibility. However, this can make it very difficult for applications upgrading from older versions of conjure-java as parameter sorting has the potential to change the signature of all service methods.

## After this PR
There is now a new feature flag which allows parameter sorting for service methods to be skipped and instead just follow the order as specified in the conjure definition. The flag will allow developers to more easily upgrade versions and continue to use conjure-java generators by providing a consistent ordering of method parameters.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
@@ -140,11 +140,14 @@ public final class JerseyServiceGenerator implements ServiceGenerator {
                 .map(endpoint -> generateServiceMethod(endpoint, returnTypeMapper, argumentTypeMapper))
                 .collect(Collectors.toList()));
 
-        serviceBuilder.addMethods(serviceDefinition.getEndpoints().stream()
-                .map(endpoint -> generateCompatibilityBackfillServiceMethods(endpoint, returnTypeMapper,
-                        argumentTypeMapper))
-                .flatMap(Collection::stream)
-                .collect(Collectors.toList()));
+        // When parameters are not sorted, do not generate the backwards compatibility service methods
+        if (!featureFlags.contains(FeatureFlags.DisableParameterSorting)) {
+            serviceBuilder.addMethods(serviceDefinition.getEndpoints().stream()
+                    .map(endpoint -> generateCompatibilityBackfillServiceMethods(endpoint, returnTypeMapper,
+                            argumentTypeMapper))
+                    .flatMap(Collection::stream)
+                    .collect(Collectors.toList()));
+        }
 
         return JavaFile.builder(serviceDefinition.getServiceName().getPackage(), serviceBuilder.build())
                 .skipJavaLangImports(true)
@@ -288,7 +291,8 @@ public final class JerseyServiceGenerator implements ServiceGenerator {
         Optional<AuthType> auth = endpointDef.getAuth();
         createAuthParameter(auth, withAnnotations).ifPresent(parameterSpecs::add);
 
-        ParameterOrder.sorted(endpointDef.getArgs()).forEach(def ->
+        boolean disableParameterSorting = featureFlags.contains(FeatureFlags.DisableParameterSorting);
+        ParameterOrder.sorted(endpointDef.getArgs(), disableParameterSorting).forEach(def ->
                 parameterSpecs.add(createServiceMethodParameterArg(typeMapper, def, withAnnotations)));
         return ImmutableList.copyOf(parameterSpecs);
     }
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceGeneratorTests.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceGeneratorTests.java
@@ -97,6 +97,25 @@ public final class JerseyServiceGeneratorTests extends TestBase {
         validateGeneratorOutput(files, Paths.get("src/test/resources/test/api"), ".jersey.binary_as_response");
     }
 
+    @Test
+    public void testSortedMethodParameters() throws IOException {
+        ConjureDefinition def = Conjure.parse(
+                ImmutableList.of(new File("src/test/resources/example-param-sorting.yml")));
+        List<Path> files = new JerseyServiceGenerator(Collections.emptySet())
+                .emit(def, folder.getRoot());
+        validateGeneratorOutput(files, Paths.get("src/test/resources/test/api"), ".jersey.sorted_params");
+    }
+
+    @Test
+    public void testUnsortedMethodParameters() throws IOException {
+        ConjureDefinition def = Conjure.parse(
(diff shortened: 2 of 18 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -288,7 +291,8 @@ private MethodSpec createCompatibilityBackfillMethod(
         Optional<AuthType> auth = endpointDef.getAuth();
         createAuthParameter(auth, withAnnotations).ifPresent(parameterSpecs::add);
 
-        ParameterOrder.sorted(endpointDef.getArgs()).forEach(def ->
+        boolean disableParameterSorting = featureFlags.contains(FeatureFlags.DisableParameterSorting);
+        ParameterOrder.sorted(endpointDef.getArgs(), disableParameterSorting).forEach(def ->
```
carterkozak, line 295: We should rename this utility function to `ordered`, it's odd to say `sort(<values>, dontReallySort)`

conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceGeneratorTests.java
```diff
@@ -97,6 +97,25 @@ public void testBinaryReturnResponse() throws IOException {
         validateGeneratorOutput(files, Paths.get("src/test/resources/test/api"), ".jersey.binary_as_response");
     }
 
+    @Test
+    public void testSortedMethodParameters() throws IOException {
+        ConjureDefinition def = Conjure.parse(
+                ImmutableList.of(new File("src/test/resources/example-param-sorting.yml")));
+        List<Path> files = new JerseyServiceGenerator(Collections.emptySet())
+                .emit(def, folder.getRoot());
+        validateGeneratorOutput(files, Paths.get("src/test/resources/test/api"), ".jersey.sorted_params");
```
carterkozak, line 106: Aren't there existing example service methods which we can reuse without a new conjure yml definition? The default `.jersey` outputs should be sufficient to test this case, I don't think this test adds anything

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceGeneratorTests.java
```diff
@@ -79,6 +79,25 @@ public void testBinaryReturnInputStream() throws IOException {
         validateGeneratorOutput(files, Paths.get("src/test/resources/test/api"), ".undertow.binary");
     }
 
+    @Test
+    public void testSortedMethodParameters() throws IOException {
+        ConjureDefinition def = Conjure.parse(
+                ImmutableList.of(new File("src/test/resources/example-param-sorting.yml")));
+        List<Path> files = new UndertowServiceGenerator(Collections.emptySet())
+                .emit(def, folder.getRoot());
+        validateGeneratorOutput(files, Paths.get("src/test/resources/test/api"), ".undertow.sorted_params");
```
carterkozak, line 88: same here

readme.md
```diff
@@ -30,6 +30,8 @@ The recommended way to use conjure-java is via a build tool like [gradle-conjure
                      Generate service interfaces for Undertow with class names prefixed 'Undertow'
         --useImmutableBytes
                      Generate binary fields using the immutable 'Bytes' type instead of 'ByteBuffer'
+        --disableParameterSorting
+                     Generate service interfaces that respect the order of method parameters in the conjure definition
```
carterkozak, line 34: Should we document this flag, since it's more of an escape hatch?

carterkozak:
Would it be possible for us to move all parameter sorting to the top level? We could re-build the `ConjureDefinition` object with all endpoint parameters sorted so that we can avoid that complexity entirely inside of generator code. We would still need to decide whether or not to add backcompat methods based on the flag.

markelliot (changes requested):
I think we should prefer the one-time break here as something that only affects compilation and not wire-format.

carterkozak:
@markelliot When I make API changes which block services from taking upgrades, I am expected to fix those services. Who is responsible for fixing this break?

A couple additional questions you may know the answer to:
Is there any reason to generate default methods and re-order arguments for server-side interfaces?
Why don't we sort parameters for retrofit clients?

markelliot:
I’d rate the compile breaks as pretty different to the wire breaks, and the severity/impact modulates the time+effort needed to help correct consumers. The wire breaks require complex/multi-step migrations that affect not just the direct consumer of Conjure but also the consumers of whatever that thing are (and it’s why we’d push for instrumenting impact).

When we made the change this PR feature flags off many months ago we assessed impact + discussed whether it needed to be flagged, we assessed that breaks were already regularly occurring in compilation, and as we surveyed Conjure users, discovered that a one-time break was strongly + widely preferred to the ongoing (severe, difficult to track down) compilation breaks from addition of optional parameters. I’d also note that this break was wound in to many of the other compile breaks added during the open-sourcing cleanup phase of this project. Even so, when we made the break several months ago, we helped migrate/clean up any consumers that asked.

Default methods for server-side interfaces: we have, since the start of Conjure, used the Jersey interface as both the server and the client interface. When we, in the past, varied the client and server behavior, a number of our internal services which act as strongly-typed proxies/dispatchers flagged (loudly) that the client/server in those cases needed to behave the same.

Retrofit clients: news to me, perhaps the change was not ported correctly to all places.

carterkozak:
> I’d rate the compile breaks as pretty different to the wire breaks, and the severity/impact modulates the time+effort needed to help correct consumers.

Yes, we're on the same page, but that's not what we're discussing in this PR.

> When we made the change this PR feature flags off many months ago we assessed impact + discussed whether it needed to be flagged, we assessed that breaks were already regularly occurring in compilation, and as we surveyed Conjure users, discovered that a one-time break was strongly + widely preferred to the ongoing (severe, difficult to track down) compilation breaks from addition of optional parameters.

I don't recall anyone reaching out to us to discuss the original change.
Why is the calculus different between this feature flag and something like https://github.com/palantir/conjure-java/pull/158#issuecomment-456030390 ? Should that feature flag be removed?

> I’d also note that this break was wound in to many of the other compile breaks added during the open-sourcing cleanup phase of this project.

Yes, and we're still working through these breaks. It's not our goal to add feature flags for all of them, but as you've stated, if we cannot upgrade to track the latest releases, compilation breaks will continue to stack up making upgrades even more difficult. We're not asking for feature flags for all compilation (or even runtime) breaks, we're looking for a way to avoid falling further behind.

> Even so, when we made the break several months ago, we helped migrate/clean up any consumers that asked.

This would be wonderful, who should we reach out to for assistance?

> When we, in the past, varied the client and server behavior, a number of our internal services which act as strongly-typed proxies/dispatchers flagged (loudly) that the client/server in those cases needed to behave the same

I didn't realize that this was something we specifically supported, thanks!

> Retrofit clients: news to me, perhaps the change was not ported correctly to all places.

Filed #339

We should catch up offline to discuss our path forward.

PR #345, opened 2019-04-17
[break] Immutable 'Bytes' type is always used for non-streamed binary values

carterkozak:
## Before this PR
By default we used dangerous mutable binary data representations.

## After this PR
==COMMIT_MSG==
Immutable 'Bytes' type is always used for non-streamed binary values
==COMMIT_MSG==

## Possible downsides?
This causes a compile break for projects which use non-streamed `binary` values and have not already opted into `--useImmutableBytes`


markelliot:
possible to do this now that we have refaster?

iamdanfox:
The key thing we want to avoid is when conjure-generated API libraries are used deep inside other libraries.  Then people can get into the nasty state where they pull in some library (e.g. fsclient) which expects the old signature, but they've dialled up their catalog dep to a version that gets the new binary signature, leading to a NoSuchMethodError.

I think if we could enable gradle-revapi on all java servers then we'd be in business (as people would see a red build from this conjure upgrade).

PR #346, opened 2019-04-17
[undertow][improvement] Generated undertow handlers use final variables

carterkozak:
## After this PR
==COMMIT_MSG==
Generated undertow handlers use final variables
==COMMIT_MSG==

This provides additional compile-time safety ensuring that we don't
accidentally set the same variable twice.

## Possible downsides?
A couple more characters per line.


rubenfiszel (changes requested):
As discussed offline, I would prefer addressing directly the problem that this change protect us against (naming collisions).

PR #347, opened 2019-04-17
[tests only] Unpin conjure-java-runtime jaxrs version

carterkozak:
This causes quite a few verifier tests to fail, but allows us to
catch new test failures from CJR changes moving forward.

These failures were originally introduced by:
https://github.com/palantir/conjure-java-runtime/pull/865

==COMMIT_MSG==
Unpin conjure-java-runtime jaxrs version
==COMMIT_MSG==


ferozco:
👍

PR #348, opened 2019-04-17
[fix][undertow] sanitize bad argument names

carterkozak:
## Before this PR
Previously some parameter names resulted in undertow handlers which failed to compile.

## After this PR
We sanitize variable names.

==COMMIT_MSG==
The Undertow generator no longer produces code that cannot compile when given certain inputs.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -16,6 +16,7 @@
 
 package com.palantir.conjure.java.services;
 
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.MoreCollectors;
 import com.palantir.conjure.java.ConjureAnnotations;
@@ -28,6 +29,7 @@ import com.palantir.conjure.java.undertow.lib.Serializer;
 import com.palantir.conjure.java.undertow.lib.TypeMarker;
 import com.palantir.conjure.java.undertow.lib.UndertowRuntime;
 import com.palantir.conjure.java.undertow.lib.UndertowService;
+import com.palantir.conjure.java.util.JavaNameSanitizer;
 import com.palantir.conjure.java.util.ParameterOrder;
 import com.palantir.conjure.java.visitor.DefaultTypeVisitor;
 import com.palantir.conjure.java.visitor.MoreVisitors;
@@ -77,7 +79,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import javax.lang.model.element.Modifier;
 import org.apache.commons.lang3.StringUtils;
 
@@ -89,8 +90,11 @@ final class UndertowServiceHandlerGenerator {
     private static final String DESERIALIZER_VAR_NAME = "deserializer";
     private static final String SERIALIZER_VAR_NAME = "serializer";
     private static final String AUTH_HEADER_VAR_NAME = "authHeader";
-
     private static final String COOKIE_TOKEN_VAR_NAME = "cookieToken";
+    private static final String RESULT_VAR_NAME = "result";
+
+    private static final ImmutableSet<String> RESERVED_PARAM_NAMES = ImmutableSet.of(
+            EXCHANGE_VAR_NAME, DELEGATE_VAR_NAME, RUNTIME_VAR_NAME, DESERIALIZER_VAR_NAME, SERIALIZER_VAR_NAME);
 
     private final Set<FeatureFlags> experimentalFeatures;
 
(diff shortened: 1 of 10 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -768,4 +774,13 @@ private static String deserializeFunctionName(Type type) {
                 .filter(arg -> arg.getParamType().accept(ParameterTypeVisitor.IS_BODY))
                 .collect(MoreCollectors.toOptional());
     }
+
+    private static String sanitizeVarName(String input, EndpointDefinition endpoint) {
+        String value = JavaNameSanitizer.sanitizeParameterName(input, endpoint);
+        if (RESERVED_PARAM_NAMES.contains(value)
+                || (endpoint.getReturns().isPresent() && RESULT_VAR_NAME.equals(value))) {
```
rubenfiszel, line 786: I would just add result to the RESERVED_PARAM_NAMES to remove complexity at the cost of over-sanitizing in some cases.

conjure-java-core/src/test/resources/dangerous-name-service.yml
```diff
@@ -0,0 +1,33 @@
...
+            type: string
+          deserializer:
+            param-type: query
+            type: string
+          delegate:
+            param-type: query
+            type: string
+          result:
+            param-type: query
+            type: string
```
rubenfiszel, line 51: what happens if you also declare params of the name authHeader_, long_, runtime_, serializer_, ... (or are those not valid conjure names)

conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
```diff
@@ -30,11 +37,48 @@
...
-        return SourceVersion.isName(name) && !RESERVED_FIELD_NAMES.contains(name) ? name : name + "_";
+    /** Returns a valid java name based on the input. */
+    public static String sanitize(String name) {
+        return SourceVersion.isName(name) ? name : escape(name);
+    }
+
+    /** Sanitizes parameter names taking into account default auth type parameter names. */
+    public static String sanitizeParameterName(String input, EndpointDefinition endpoint) {
+        String value = sanitize(input);
+        Optional<String> maybeAuthParamName = endpoint.getAuth()
```
rubenfiszel, line 51: Same as above, I would just add authHeader and cookieToken to the list of reserved keywords.

conjure-java-core/src/test/resources/dangerous-name-service.yml
carterkozak: `Parameter names in endpoint paths and service definitions must match pattern ^[a-z][a-z0-9]*([A-Z0-9][a-z0-9]+)*$`

The same isn't true for method names, so `int` is renamed to `int_`, which may conflict with another defined endpoint `int_`. That case is a little bit _too_ weird, so I'm fine letting that one fail for now. We'll only automatically fix so many bad things for you :-)

conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
carterkozak: I've seen endpoints which use `auth: none` and define an `authHeader` header parameters, so I'd prefer not to mangle those unnecessarily.
Originally I planned to add these and `result` to the restricted list, but this example made me reconsider.

conjure-java-core/src/test/resources/dangerous-name-service.yml
rubenfiszel: Is the parameter names pattern actually enforced?
A possible fix for method names is to always add `_` if the name end with `_`.

carterkozak: Parameter name regex is enforced when the conjure schema file is parsed.I don't follow your solution for method names, the problem is if I have a method named `int` as well as `int_`. the `int_` method is fine, but attempting to escape `int` produces a method name which matches the existing `int_` method.

conjure-java-core/src/test/resources/dangerous-name-service.yml
```diff
@@ -0,0 +1,33 @@
...
+    package: com.palantir.product
+    base-path: /
+
+    endpoints:
+      int:
+        http: GET /
+        auth: header
+        args:
+          authHeader:
+            param-type: query
```
rubenfiszel, line 19: what about path, header and body params ? Only query seems covered.

rubenfiszel: btw, I think body params are not sanitized

conjure-java-core/src/test/resources/dangerous-name-service.yml
rubenfiszel: @carterkozak Yes, that's why if in addition to escape `int` to `int_` you also escape `int_` to `int__` it works. To avoid collision you need the function to be bijective, which mean you need to also displace every element of the codomain

rubenfiszel: (I edited the n-2 comment which was weirdly formatted and as a result hard to understand)

rubenfiszel:
Are body param sanitized? Does it also work well with the markers invocation ?

conjure-java-core/src/test/resources/dangerous-name-service.yml
carterkozak: I'd rather avoid renaming things that aren't themselves invalid, my preference is to fail if we cannot provide a simple fix.

carterkozak:
Good call, I missed one body param variable rename. Fixed + tests.

conjure-java-core/src/test/resources/dangerous-name-service.yml
rubenfiszel: Isn't my solution a simple fix :) ? Also those are not more or less invalid than then one that collide with the reserved keywords.

carterkozak: There are existing services with methods named things like `boolean_` which would be modified.

rubenfiszel:
I am not sure you handle markers invocation correctly. Can we add some to the test?

conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
rubenfiszel: That's fair but tbh I don't really see much downside of over-sanitizing.

carterkozak:
Ah good call, I was sending the sanitized name as the marker parameter name, the string value shouldn't be escaped.

conjure-java-core/src/test/resources/dangerous-name-service.yml
rubenfiszel: Right now is probably the ideal time to break, after it will be too late. But fair enough if you think the cost is too high.

rubenfiszel (approved):
There is still a gap for method names when an existing method name is equal to the sanitization of another but this PR is still a net improvement without any downside compared to the current state. 

Thanks a lot for doing it!

iamdanfox (approved):
LGTM!

PR #352, opened 2019-04-19
[fix] Alias 'valueOf' generation validates nested aliases recursively

carterkozak:
## Before this PR
Both examples added to `example-types.yml` failed to compile.

## After this PR
==COMMIT_MSG==
Certain nested aliases no longer result in Java which cannot compile.
==COMMIT_MSG==

closes #130


ferozco:
👍

PR #354, opened 2019-04-20
[improvement] Beans with no fields generate toString methods returning constant values

carterkozak:
## After this PR
==COMMIT_MSG==
Beans with no fields generate toString methods returning constant values
==COMMIT_MSG==


carterkozak:
Out of curiosity, @iamdanfox do you recall why this was added? https://github.com/palantir/conjure-java/blob/902a33ef67dd82dc961e46ec80a995bc39478d22/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java#L102-L107

Tests don't fail without it, but it may work differently on older builds of jackson.

ferozco (approved):

PR #356, opened 2019-04-21
[improvement] Simpler alias valueOf implementation

carterkozak:
We no longer duplicate code between `of` and `valueOf`. `valueOf` simply
parses a string and passes the result to `of`.
Alias.of(null) results in a safe-logging npe rather than using
`Objects.requireNonNull`. Slightly simplified constructors to use
the result of the null check.
Add a `valueOf` function to alias of external type imports which both
declare a base type allowed to have a valueOf, and have a
`public static valueOf` function on the imported type (only applies
to types known by the generator).

## After this PR
==COMMIT_MSG==
Simpler alias valueOf implementation
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
@@ -21,17 +21,19 @@ import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.conjure.java.ConjureAnnotations;
 import com.palantir.conjure.java.visitor.MoreVisitors;
 import com.palantir.conjure.spec.AliasDefinition;
+import com.palantir.conjure.spec.ExternalReference;
 import com.palantir.conjure.spec.PrimitiveType;
 import com.palantir.conjure.spec.Type;
 import com.palantir.conjure.visitor.TypeDefinitionVisitor;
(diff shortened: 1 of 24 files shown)


rubenfiszel (approved):

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -215,16 +222,32 @@ private static CodeBlock valueOfFactoryMethodForPrimitive(
                 "Unsupported primitive type: " + primitiveType + "for `valueOf` method.");
     }
 
+    /**
+     * Detects if the class is available in the generator classpath and has a public static <code>valueOf(String)
+     * </code>.
+     */
+    private static boolean hasValueOfFactory(com.palantir.conjure.spec.TypeName className) {
```
ferozco, line 231: I believe that with this change we would be implicitly exposing the dependencies of conjure-java as API since changes to the dependencies could impact the generated code. What case did you have in mind for this?

carterkozak: I have several `alias: Long` where `Long` is an external-type-import of java.lang.Long. I can bound this to primitives.

ferozco:
👍

carterkozak:
tyvm

PR #361, opened 2019-04-24
[undertow] Asynchronous request processing

carterkozak:
This is an implementation of the ListenableFuture based approach

Target generated interface:

```java
interface MyService {
  ListenableFuture<Foo> getFoo(AuthHeader authHeader);
}
```

impl:

```java
class MyResource {
  @Override
  ListenableFuture<Foo> getFoo(AuthHeader authHeader) {
    return Futures.immediateFuture(Foo.builder().build());
  }
}
```

Where the blocking implementation would take the form:

```java
interface MyService {
  Foo getFoo(AuthHeader authHeader);
}
```

Sample generated Endpoints:
https://github.com/palantir/conjure-java/blob/eb25f77a0bd28c04df894ffd7dce34498f19ce35/conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncTargetGeneratedEndpoints.java#L161-L181

==COMMIT_MSG==
The Undertow server target supports asynchronous request processing.
==COMMIT_MSG==


conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncTargetGeneratedService.java
```diff
@@ -0,0 +1,30 @@
...
+import com.palantir.conjure.java.undertow.lib.BinaryResponseBody;
+import com.palantir.tokens.auth.AuthHeader;
+
+// TODO(ckozak): Delete and replace with generated code when the generator is ready
+public interface AsyncTargetGeneratedService {
+    ListenableFuture<String> string(AuthHeader authHeader);
+
+    ListenableFuture<Integer> integer(AuthHeader authHeader);
+
+    ListenableFuture<BinaryResponseBody> binary(AuthHeader authHeader);
```
bavardage: so on which thread does `BinaryResponseBody#write` end up being called?

I guess this is back on an io thread?

carterkozak: `BinaryResponseBody#write` is always executed on the worker pool, blocking I/O is not allowed on the I/O threads, they're for non-blocking I/O only.

carterkozak: A slow client will cause `BinaryResponseBody#write` to take a very long time, so it must run on the large pool.

bavardage: Ok - yeah - that makes sense. I now see where we dispatch back to the worker pool.

bavardage: just in https://github.com/palantir/conjure-java/pull/361/files#diff-c3b5cc874401c390b8d96224a8786de5R70 ?

carterkozak: Yep, exactly

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -0,0 +1,82 @@
...
+
+        // TODO(ckozak): potential optimization: avoid both dispatching and scheduling a timeout if the future is
+        // already complete.
+
+        // Dispatch the registration task. We're executing inside of root handler, so the dispatched task will
+        // execute on this thread after the current handler completes execution. This allows us to avoid racing
+        // registration with completion.
+        exchange.dispatch(() -> {
+            XnioExecutor.Key timeoutKey = exchange.getIoThread()
+                    .executeAfter(() -> future.cancel(false), timeout.toMillis(), TimeUnit.MILLISECONDS);
```
markelliot: is this what cancels work if the caller closes their connection?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -68,13 +72,25 @@ public AuthorizationExtractor auth() {
...
 
+        private Duration asyncTimeout = Duration.ofMinutes(3);
         private final List<Encoding> encodings = Lists.newArrayList();
         private final List<ParamMarker> paramMarkers = Lists.newArrayList();
 
         private Builder() {}
 
+        @CanIgnoreReturnValue
+        public Builder asyncTimeout(Duration value) {
+            asyncTimeout = Preconditions.checkNotNull(value, "Value is required");
```
markelliot: probably a bit more useful for the message to read `asyncTimeout is required`

markelliot:
Looks cool! Should we consider having some type that we own internally, for purposes of discussion here, let's call it `Promise<T>` that we use instead of forcing use of Guava? I think it's possible we may want to limit how much we end up depending on 3rd party stuff in API implementations.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
carterkozak: Only after the timeout has elapsed. This prevents services which return a future that is never completed from leaving requests active indefinitely.

I can add a listener to cancel the future if the server detects that the client has closed a connection/stream, but that's an optimization we can add outside of the initial implementation.

carterkozak:
I agree that we should avoid third party dependencies where it's feasible to do so, but the counterargument is that we may create N new standards to solve the same problem, and steepen the learning curve for developers to use this feature.

Any approach we take using the `Promise<T>` model (including ListenableFuture) requires developers to understand how to create and use a concrete `Promise` implementation. If we pick a type which developers are already familiar with, it will reduce friction. There's an additional cost to implement supporting functionality as well as teaching devs how to use it.

An alternative approach is to use java 8 [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html) which solves the same set of problems as ListenableFuture. The API isn't as elegant as the guava type, and integration with the fork-join pool may yield unexpected results, but we can add error-prone checks to disallow usage that results in execution on the fork-join pool. Unlike ListenableFuture, it's relatively difficult to propagate cancellation back through a chain of CompletableFutures, which will be problematic when we time out operations or detect dropped connections.

The benefit of Guavas ListenableFuture is the breadth of supporting functionality
https://github.com/google/guava/wiki/ListenableFutureExplained
Using guava functionality to create and transform futures results in free cancellation propagation.

carterkozak:
@markelliot Most of my response was spent avoiding your question about implementing our own `Promise` type, and listed alternatives. The upside is greater control, and if we do it well, potentially a more ergonomic API (this worked well for `Refreshable<T>`). The place where I'm getting stuck is that this problem isn't sufficiently different from `ListenableFuture`, so I don't see a great deal of upside in building our own `Promise<T>` aside from shedding the guava dependency. The cost will be relatively high as we would need to reimplement guava functionality, as well as adapters between `Promise<T>` and guava types.

@bavardage @j-baker Do you have thoughts/opinions on the target return type?

uschi2000:
I'm not actually terribly concerned with exposing ListenableFuture because these are not cross-service/repo APIs: the guava dependency would be localized to my service only, and consumers of my API don't have to know that internally I'm using a ListenableFuture.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -0,0 +1,99 @@
...
+final class ConjureAsyncRequestProcessing implements AsyncRequestProcessing {
+
+    private static final Executor DIRECT_EXECUTOR = MoreExecutors.directExecutor();
+    private static final AttachmentKey<ListenableFuture<?>> FUTURE = AttachmentKey.create(ListenableFuture.class);
+    // If the request is ended before the future has completed, cancel the future to signal that work
+    // should be stopped. This occurs when clients cancel requests or connections are closed.
+    private static final ExchangeCompletionListener COMPLETION_LISTENER = (exchange, nextListener) -> {
+        ListenableFuture<?> future = exchange.getAttachment(FUTURE);
+        if (future != null && !future.isDone()) {
+            future.cancel(false);
```
uschi2000: why false?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -0,0 +1,99 @@
...
+
+final class ConjureAsyncRequestProcessing implements AsyncRequestProcessing {
+
+    private static final Executor DIRECT_EXECUTOR = MoreExecutors.directExecutor();
+    private static final AttachmentKey<ListenableFuture<?>> FUTURE = AttachmentKey.create(ListenableFuture.class);
+    // If the request is ended before the future has completed, cancel the future to signal that work
+    // should be stopped. This occurs when clients cancel requests or connections are closed.
+    private static final ExchangeCompletionListener COMPLETION_LISTENER = (exchange, nextListener) -> {
+        ListenableFuture<?> future = exchange.getAttachment(FUTURE);
+        if (future != null && !future.isDone()) {
```
uschi2000: there's a race here between isDone and cancel (unless you can convince me that no other thread can possibly access this attachment)

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        nextListener.proceed();
+    };
+
+    private final Serializer<SerializableError> exceptionSerializer = ConjureExceptions.serializer();
+
+    private final Duration timeout;
+
+    ConjureAsyncRequestProcessing(Duration timeout) {
+        this.timeout = Preconditions.checkNotNull(timeout, "Timeout is required");
```
uschi2000: IMO, don't need the null checks

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowRuntime.java
```diff
@@ -35,4 +35,7 @@
 
     /** Provides the {@link MarkerCallback} to execute arbitrary actions given a marker on a param. */
     MarkerCallback markers();
+
+    /** Provides the {@link AsyncRequestProcessing} used to detach execution from the request thread. */
+    AsyncRequestProcessing async();
```
uschi2000, line 40: sync implementations don't need the async stuff. possible to separate?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -0,0 +1,99 @@
...
+        Preconditions.checkNotNull(future, "future");
+        Preconditions.checkNotNull(returnValueWriter, "returnValueWriter");
+        Preconditions.checkNotNull(exchange, "exchange");
+
+        // TODO(ckozak): potential optimization: avoid both dispatching and scheduling a timeout if the future is
+        // already complete.
+
+        // Attach data to the exchange in order to reuse a stateless listener
+        exchange.putAttachment(FUTURE, future);
+        // Cancel the future if the exchange is completed before the future is done
```
uschi2000, line 123: what happens if the future tries to write to the exchange after the exchange has been completed? presumably we'll get an error thrown on the future thread?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -0,0 +1,99 @@
...
+        exchange.addExchangeCompleteListener(COMPLETION_LISTENER);
+
+        XnioExecutor.Key timeoutKey = exchange.getIoThread()
+                .executeAfter(() -> future.cancel(false), timeout.toMillis(), TimeUnit.MILLISECONDS);
```
uschi2000: why false?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -0,0 +1,99 @@
...
+        future.addListener(timeoutKey::remove, DIRECT_EXECUTOR);
+        // Dispatch the registration task. We're executing inside of root handler, so the dispatched task will
+        // execute on this thread after the current handler completes execution. This allows us to avoid racing
+        // registration with completion.
+        // This leaves the exchange in a dispatched state preventing the calling Connectors.executeRootHandler
+        // from terminating this request while the future awaits completion.
+        exchange.dispatch(() -> Futures.addCallback(future, new FutureCallback<T>() {
```
uschi2000, line 136: the javadoc indicates that you should check whether `isInIoThread`

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
uschi2000: no need to check whether the future is already done?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -0,0 +1,99 @@
...
+        // Attach data to the exchange in order to reuse a stateless listener
+        exchange.putAttachment(FUTURE, future);
+        // Cancel the future if the exchange is completed before the future is done
+        exchange.addExchangeCompleteListener(COMPLETION_LISTENER);
+
+        XnioExecutor.Key timeoutKey = exchange.getIoThread()
+                .executeAfter(() -> future.cancel(false), timeout.toMillis(), TimeUnit.MILLISECONDS);
+        future.addListener(timeoutKey::remove, DIRECT_EXECUTOR);
+        // Dispatch the registration task. We're executing inside of root handler, so the dispatched task will
+        // execute on this thread after the current handler completes execution. This allows us to avoid racing
```
uschi2000: I *think* I get it, but I'm not sure; can you try to reword this explanation?
(is the problem that we need to call exchange.dispatch before the root handler terminates so that the connection isn't closed?)

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncRequestProcessingTest.java
```diff
@@ -0,0 +1,184 @@
...
+        server = Undertow.builder()
+                .addHttpListener(12345, "localhost")
+                .setHandler(handler)
+                .build();
+        server.start();
+    }
+
+    @After
+    public void after() throws InterruptedException {
+        if (server != null) {
```
uschi2000: remove null check

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncRequestProcessingTest.java
```diff
@@ -0,0 +1,184 @@
...
+            server.stop();
+        }
+        if (scheduler != null) {
```
uschi2000: remove null check

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncRequestProcessingTest.java
```diff
@@ -0,0 +1,184 @@
...
+        };
+        HttpHandler httpHandler = exchange -> {
+            OptionalInt maybeDelay = runtime.plainSerDe()
+                    .deserializeOptionalInteger(exchange.getQueryParameters().get("delay"));
+            final ListenableFuture<String> future;
+            if (maybeDelay.isPresent()) {
+                int delay = maybeDelay.getAsInt();
+                future = Futures.transform(
+                        scheduler.schedule(() -> { }, delay, TimeUnit.MILLISECONDS),
+                        ignored -> "Completed " + "after " + delay + "ms",
```
uschi2000: remove one +

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncRequestProcessingTest.java
```diff
@@ -0,0 +1,184 @@
...
+            exchange.getResponseSender().send(returnValue);
+        };
+        HttpHandler httpHandler = exchange -> {
+            OptionalInt maybeDelay = runtime.plainSerDe()
+                    .deserializeOptionalInteger(exchange.getQueryParameters().get("delay"));
+            final ListenableFuture<String> future;
+            if (maybeDelay.isPresent()) {
+                int delay = maybeDelay.getAsInt();
+                future = Futures.transform(
+                        scheduler.schedule(() -> { }, delay, TimeUnit.MILLISECONDS),
```
uschi2000: I think we need to help users of the async framework define a sensible executor for their work. maybe a `ConjureAsyncRuntime` could provide such a thing?

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncRequestProcessingTest.java
```diff
@@ -0,0 +1,184 @@
...
+        this.scheduler = MoreExecutors.listeningDecorator(Executors.newSingleThreadScheduledExecutor());
+        UndertowRuntime runtime = ConjureUndertowRuntime.builder()
+                // Use a 1 second timeout so we can wait for the timeout to be hit without tests running too long.
+                .asyncTimeout(Duration.ofSeconds(1))
+                .build();
+        ReturnValueWriter<String> writer = (returnValue, exchange) -> {
+            exchange.getResponseHeaders().put(Headers.CONTENT_TYPE, "text/plain");
+            exchange.getResponseSender().send(returnValue);
+        };
+        HttpHandler httpHandler = exchange -> {
```
uschi2000: I'd be interested in seeing the API that a service author would interact with. (I assume they won't get to see the exchange)

uschi2000: ignore me, saw the thing down below

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncTargetGeneratedService.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.palantir.conjure.java.undertow.runtime;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.conjure.java.undertow.lib.BinaryResponseBody;
+import com.palantir.tokens.auth.AuthHeader;
+
+// TODO(ckozak): Delete and replace with generated code when the generator is ready
+public interface AsyncTargetGeneratedService {
+    ListenableFuture<String> string(AuthHeader authHeader);
```
uschi2000: picking up the question from above: how am I expected to construct these futures? on what thread pool should I run them?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
carterkozak: Thread interruption can result in unexpected behavior which I'd prefer to avoid forcing service implementors to handle. The common case for this feature is not for work running on a thread, but for queued async tasks where the interruption flag makes no difference.
I will extract a boolean constant and add a comment.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
carterkozak: The `isDone` check is unnecessary, `cancel(boolean)` is threadsafe and will no-op if the future has completed in some other way.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowRuntime.java
carterkozak: the `UndertowRuntime` interface is only meant to be consumed by generated code. Sync implementations will not be generated to invoke `async()`. I think additional separation will make the code unwieldy, but you may have something in mind that I haven't considered.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
carterkozak: same as the other future, replied above

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
carterkozak: Writing the error response back to a client cannot work if the request has completed, but that's no different from the blocking case when a client drops. Either way we log the failure gracefully.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
carterkozak: No, the `isDone` check in the completion listener was unnecessary, I have removed it. If the future is done, `cancel` is a no-op.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
carterkozak: That is correct, I've updated the comment.  Does it make more sense this way?

```java
// Dispatch the registration task, this accomplishes two things:
// 1. Puts the exchange into a 'dispatched' state, otherwise the request will be terminated when
//    the endpoint HttpHandler returns. See Connectors.executeRootHandler for more information.
// 2. Avoids racing completion with the current handler chain. The dispatched task will be executed
//    after this Endpoints HttpHandler returns. Otherwise an unlucky future could race
//    HttpServerExchange.isInCall transitioning from true -> false, causing hte dispatch task not
//    to execute.
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
carterkozak: I will add class level javadoc describing the thread model. The description in HttpServerExchange.dispatch javadoc describes how to move work from an I/O thread to a task thread, but that's not what we want to accomplish here -- we already know we're on a task thread, but we need to mark the exchange as dispatched while we wait for the future to complete. I updated the comment, hopefully that makes the intent clearer.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncTargetGeneratedService.java
carterkozak: In the ideal case, you do not construct them yourself, but use a ListenableFuture from something else (a dialog async client future, [ListeningExecutorService](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/util/concurrent/ListeningExecutorService.html)), perhaps with some transformation.

https://github.com/google/guava/wiki/ListenableFutureExplained#creation

The conjure async request processing feature is designed to allow any ListenableFuture to be sufficient regardless of how the service author constructs it. On completion, we dispatch back to the worker pool to serialize the result (success or failure) the same way we would in the blocking case, regardless of how and where the future executed. The only thing that we execute on the thread which completes the future is `exchange.dispatch(args)`, which sends work to the task pool.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncRequestProcessingTest.java
carterkozak: Our server library already provides an executor factory, and service implementations don't get access to the `ConjureRuntime` and by extension the `ConjureAsyncRuntime`.

We shouldn't provide executors to services because we expect this feature to be used to reduce thread use. If the goal is to run work on another thread, then return the result, there's no reason to return a future, one can execute that work synchronously in the service method.

In some cases it may be useful for specific types of endpoints to use executors to allow certain types of tasks to queue for execution on a bounded number of threads, but I don't think this framework is the correct place to manage that executor.

carterkozak:
ready for another review.

carterkozak:
Feature complete implementation, still some work to clean things up a bit before it can be merged.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowTypeFunctions.java
```diff
@@ -259,6 +265,18 @@ public Boolean visitUnknown(String unknownType) {
         }
     }
 
+    // TODO(ckozak): Better algorithm to opt into async functionality
```
carterkozak: Looking for better ideas to opt into this feature.

My preference is to allow it to be configured per endpoint without requiring us to generate entire apis as either blocking or asynchronous. I don't want developers to split a single service API into multiple services for convenience when some endpoints benefit from asynchronous processing.

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -474,11 +484,39 @@ public void testEnumHeaderParameter() {
...
+                .isInstanceOf(RemoteException.class)
+                .satisfies(throwable -> assertThat(((RemoteException) throwable).getError().parameters())
+                        .containsEntry("type", "failedFuture"));
+    }
+
+    @Test
+    public void testAsyncOptional() {
+        Optional<String> input = Optional.of("Hello, World!");
+        assertThat(asyncServiceClient.optionalEnumQueryAsync(AuthHeader.valueOf("authHeader"), input)).isEqualTo(input);
+    }
```
carterkozak: These belong in AsyncRequestProcessingTest, we don't need to modify this test.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowTypeFunctions.java
uschi2000: I'm somewhat skeptical this is the right choice, because it complicates the conjure configuration space and the workaround is as simple as `return Futures.immediateFuture(syncReturnValue)`

uschi2000 (commented):
fine with moving forward with this.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncRequestProcessingTest.java
uschi2000: I think this is quite easy to abuse / needs a lot of understanding of undertow internals and threading strategies. Thoughts on how to avoid bad "pseudo-async" uses? Maybe we should write 2-3 examples of "good" and "bad".

carterkozak: > I think this is quite easy to abuse / needs a lot of understanding of undertow internals and threading strategies

The intent of the design was to avoid needing to understand how the server works behind the scenes, and provide something that works as expected. The thread model I described should only need to be understood by developers writing code for the undertow generator and runtime.

Could you describe how this could be abused? I think there are ways it could be used inefficiently, but that's true of existing generator implementations as well.

I think I may have created confusion in this test, the handler implementation uses hand-written handler code because I had not written the generator yet at this point. All uses of the async component should be done through code we generate, which provides little to no room for abuse. I'm planning to replace these tests with a set which use generated handler code instead.

> Thoughts on how to avoid bad "pseudo-async" uses? Maybe we should write 2-3 examples of "good" and "bad".

Good idea, do you think the readme is a good place for these, or perhaps a separate `docs/undertow_async.md` document? Alternatively I can embed examples in javadoc.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowTypeFunctions.java
carterkozak: I agree that this is dirty, and I don't think it would be problematic for foundry if we used a global switch (e.g. `--undertowListenableFutures`).

I'm responsible for a project with ~1000 endpoints, the majority of which exist in a single module, ~60 of these endpoints require asynchronous request processing in order to avoid quickly saturating the server worker pool.

Thoughts on this approach?
1. Implement the `--undertowListenableFutures` CLI flag to make every method return ListenableFuture.
2. Implement second `--experimentalUndertowAsyncMarkers` flag which can be enabled to opt into asynchronous processing per-endpint. I will tighten the marker values to accept only an external-type-import annotation to avoid allowing markers to grow beyond existing usage on parameters.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncRequestProcessingTest.java
uschi2000: I'd put it in the README. Naively, if I saw I should implement a
```
interface MyService {
  ListenableFuture<String> foo();
}
```
, I would make a `MyResource`, have it eat an ExecuterService, and then in my `MyServer` init method create something like a `Executors.newFixedThreadPool` with some arbitrary number, and then compute my futures on that thread pool. And then, I'd think, I'd have a perfectly async service with great performance. I imagine that's not what you want me to do :)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowTypeFunctions.java
uschi2000: ok

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AsyncRequestProcessingTest.java
carterkozak: Will do, thank you for the example :-)

carterkozak:
Remaining work:
* ~~Update AsyncRequestProcessingTest to use generated code instead of adding cruft to UndertowServiceEteTest~~
* Provide readme examples of both good and bad uses of uses of this feature.

carterkozak:
Ready for review.

carterkozak:
Friendly ping for review @uschi2000 @markelliot @j-baker @bavardage

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -68,22 +72,34 @@ public AuthorizationExtractor auth() {
...
     }
 
+    @Override
+    public AsyncRequestProcessing async() {
+        return async;
+    }
+
     public static final class Builder {
 
+        private Duration asyncTimeout = Duration.ofMinutes(3);
```
carterkozak, line 82: Do we want to increase this to match `ClientConfigurations.DEFAULT_READ_TIMEOUT` (5 minutes)? I chose 3 because it's a relatively low number, and it will be harder to decrease this moving forward than it is to increase. I also considered using 2 minutes based on https://github.com/palantir/conjure-java-runtime/pull/966

5 minutes is the safest bet to avoid breaking the proxy use case, but we may leverage this to reduce call times.

Thoughts?

uschi2000 (approved):

carterkozak:
I am planning to merge this tomorrow morning (EDT) unless there is additional feedback. If you would like additional time please let me know.

readme.md
```diff
@@ -220,6 +222,89 @@ public static void main(String[] args) {
 }
 ```
 
+### Asynchronous Request Processing
```
rubenfiszel, line 225: Great docs!

rubenfiszel (approved):
As spoken offline, my personal preference would have had to use a generator flag to set the default mode, and endpoint markers to override. Currently, you can only override with "async" and there is no "sync" markers.
However, I understand that in order to disincentivize the use of markers, this implementation prefers the override to only go one way, from sync to async.

rubenfiszel:
Really excited to see this feature in action!

PR #362, opened 2019-04-25
[undertow] Extract logic from the conjure exception handling HttpHandler

carterkozak:
This is a refactor, no logic or public API change.

Exception serialization logic moved from
`ConjureExceptionHandler.handleRequset(HttpServerExchange)` into
an internal utility function `ConjureExceptions.handle`.

This is a pre-req for async request processing, see #361


carterkozak:
Goal is to reduce the size of the change in #361 so we can focus on the important pieces.

rubenfiszel (approved):

PR #367, opened 2019-05-07
[fix] Box external type before calling valueOf in UndertowServiceHandlerGenerator

dsd987:
## Before this PR
Use of external imports that have primitive types would cause parameter code blocks generated by the UndertowServiceHandlerGenerator to not compile by calling valueOf on the primitive type.

## After this PR
==COMMIT_MSG==
Parameter code generated by the UndertowServiceHandlerGenerator now correctly boxes external types.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EteService.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EteService.java
@@ -79,7 +79,7 @@ public interface EteService {
 
     @GET
     @Path("base/externalLong/{param}")
-    String externalLongPath(
+    long externalLongPath(
             @HeaderParam("Authorization") @NotNull AuthHeader authHeader,
             @PathParam("param") long param);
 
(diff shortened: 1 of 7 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -603,7 +603,7 @@ private CodeBlock decodePlainParameterCodeBlock(Type type, TypeMapper typeMapper
                     "$1T $2N = $3T.valueOf($4N.plainSerDe().deserializeString($5N.get($6S)))",
                     typeMapper.getClassName(type),
                     resultVarName,
-                    typeMapper.getClassName(type),
+                    typeMapper.getClassName(type).box(),
```
carterkozak, line 606: oops!

carterkozak (approved):
Thanks!

conjure-java-core/src/integrationInput/java/com/palantir/product/EteService.java
```diff
@@ -77,6 +77,12 @@ String path(
             @HeaderParam("Authorization") @NotNull AuthHeader authHeader,
             @PathParam("param") String param);
 
+    @GET
+    @Path("base/externalLong/{param}")
+    String externalLongPath(
```
carterkozak: Missing an implementation in EteResource.java

dsd987: Shoot! Will add that in! Also going to change it to return Long as well. Accidentally put string.

PR #368, opened 2019-05-07
[fix] Fix deserialization for aliases of external type imports

carterkozak:
Regression introduced between 1.10.3 and 3.5.

JsonCreator factory methods are handled as 'native' deserializers,
bypassing all ObjectMapper configuration and deserializers.
https://github.com/FasterXML/jackson-databind/issues/2318

In cases when both a `valueOf` and `of` method exist for an
external type import, we do not apply the `JsonCreator`
annotation, allowing jackson to choose the most accurate
deserializer function (either `of` or `valueOf` depending
on the input node type).

## After this PR
==COMMIT_MSG==
Fix deserialization for aliases of external type imports
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/ExternalLongAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/ExternalLongAliasExample.java
@@ -1,5 +1,6 @@
 package com.palantir.product;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import javax.annotation.Generated;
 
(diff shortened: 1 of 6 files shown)


dansanduleac:
How does this work for aliases of string? do you get two methods `of` and `valueOf` that both take String?

carterkozak:
Hmm, I think eeb0f801dd92c044366090e57e38fb840291a257 breaks that, I will revert.

carterkozak:
Actually, disregard. `alias<external string>` doesn't get a valueOf, so it doesn't actually break. I can add tests for that case. I'm not sure why anyone would external-import string instead of using the builtin type

carterkozak:
Updated, ready for review :-)

dansanduleac (approved):

PR #378, opened 2019-05-13
[WIP][improvement] Undertow deprecation marking

carterkozak:
## After this PR
==COMMIT_MSG==
Implement first-class support for endpoint deprecation
==COMMIT_MSG==


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowRuntime.java
```diff
@@ -38,4 +38,7 @@
 
     /** Provides the {@link AsyncRequestProcessing} used to detach execution from the request thread. */
     AsyncRequestProcessing async();
+
+    /** Provides the {@link DeprecationCallback} to execute for each request given a deprecated endpoint. */
+    DeprecationCallback deprecated();
```
carterkozak: `deprecation()`

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/DeprecationCallback.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.undertow.lib;
+
+import io.undertow.server.HttpServerExchange;
+
+public interface DeprecationCallback {
+
+    void deprecated(
```
carterkozak, line 27: javadoc

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -103,6 +112,12 @@ public Builder paramMarker(ParamMarker value) {
             return this;
         }
 
+        @CanIgnoreReturnValue
+        public Builder deprecated(DeprecationCallback value) {
+            deprecated = Preconditions.checkNotNull(value, "deprecated is required");
+            return this;
+        }
```
carterkozak, line 119: I don't think this should be configurable (much like the plain serde interface)

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/DeprecationCallback.java
```diff
@@ -0,0 +1,29 @@
...
+            String packagePath,
```
carterkozak: I don't think the package is very helpful, just endpoint name and service name should be sufficient. Alternatively we could replace the two of those with the `Endpoint` object.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowRuntime.java
rubenfiszel: done

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/DeprecationCallback.java
rubenfiszel: done

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/DeprecationCallback.java
rubenfiszel: I removed the packagePath.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
rubenfiszel: the plain SerDe interface is completely implemented in conjure-undertow-lib which I disagree we should be doing for this. Plain SerDe implemented here makes sense because we do want to have tight control over the result of SerDe (else the webserver would not work correctly). On the other hand, this is similar to the markerCallbacks and I feel like clients should be free to implement this how they please.

carterkozak: I think this is more similar to our exception handling code than it is to markers. While some consumers would prefer to implement their own exception handling, we don't allow it because it allows for unnecessarily complex and confusing code.
Much like exception handling, we should decide how deprecation should be handled, and do it properly for all implementations.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -91,6 +91,7 @@
     private static final String RUNTIME_VAR_NAME = "runtime";
     private static final String DESERIALIZER_VAR_NAME = "deserializer";
     private static final String SERIALIZER_VAR_NAME = "serializer";
+    private static final String DEPRECATED_CALLBACK_VAR_NAME = "deprecation";
```
carterkozak: deprecation().deprecated

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
rubenfiszel: In this case, we would need to pass the DefaultTaggedRegistry. Should we pass it to this builder?

carterkozak: Yep, that would be great!

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
rubenfiszel: actually I renamed every deprecatiated to depreciation. So this would be `deprecation.deprecation`.

Is it bad?

carterkozak: I think `deprecation()` gives us the interface which handles functionality related to deprecated things, the `deprecated(args)` method marks an endpoint as being deprecated. That's how I modeled it in my head anyhow.

rubenfiszel:
@carterkozak In lights of the fact that it is now only adding a safe param to the exchange, it is extremely similar to the paramMarker and so is implemented in the same fashion.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -103,6 +112,12 @@ public Builder paramMarker(ParamMarker value) {
             return this;
         }
 
+        @CanIgnoreReturnValue
+        public Builder deprecation(DeprecationCallback value) {
+            deprecation = Preconditions.checkNotNull(value, "paramMarker is required");
```
carterkozak, line 117: s/paramMarker/deprecationCallback

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/DeprecationCallback.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package com.palantir.conjure.java.undertow.lib;
+
+import io.undertow.server.HttpServerExchange;
+
+public interface DeprecationCallback {
+
+    /**
+     * Called when a deprecated endpoint (whose deprecated field contains some documentation)
```
carterkozak, line 24: > whose deprecated field contains some documentation

Not sure that's relevant since it's required to mark an endpoint as deprecated.

carterkozak (commented):
As an alternative, what if we added a new method to `Endpoint`?

```java
default Optional<String> deprecated() {
    return Optional.empty();
}
```

We could override this for anything that is deprecated, then update `ConjureHandler` to wrap only deprecated endpoints with a handler to apply the callback.

rubenfiszel:
I like the endpoint change proposal. I will experiment with it.

rubenfiszel:
#405

PR #385, opened 2019-05-16
[fix] Box external types before calling valueOf when used in optionals in UndertowServiceHandlerGenerator

dsd987:
## Before this PR
Use of external imports that have primitive types that are used in optionals would cause parameter code blocks generated by the UndertowServiceHandlerGenerator to not compile by calling valueOf on the primitive type.

## After this PR
==COMMIT_MSG==
Parameter code generated by the UndertowServiceHandlerGenerator now correctly boxes external types used in optionals.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -678,7 +678,7 @@ final class UndertowServiceHandlerGenerator {
                 functionName,
                 paramsVarName,
                 paramId,
-                typeMapper.getClassName(getComplexType(type))));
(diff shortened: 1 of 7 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -678,7 +678,7 @@ private CodeBlock decodePlainParameterCodeBlock(Type type, TypeMapper typeMapper
                 functionName,
                 paramsVarName,
                 paramId,
-                typeMapper.getClassName(getComplexType(type))));
+                typeMapper.getClassName(getComplexType(type)).box()));
```
carterkozak, line 681: Nice catch

carterkozak:
:+1:

PR #391, opened 2019-05-22
[improvement] SafeLong is now Comparable

j-baker:
```java
         return Long.compare(left.get().longValue(), right.get().longValue());
```

is the sort of code that is pretty sad :)

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
@@ -19,6 +19,7 @@ package com.palantir.conjure.java.lib;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.Preconditions;
+import java.util.Comparable;
 import org.immutables.value.Value;
 
 /**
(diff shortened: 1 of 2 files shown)


carterkozak:
Devils advocate: Does this open the door for comparable `alias<safelong>`? `alias<integer>`, etc?

j-baker:
Yes, it does, and I'd like that, too?

j-baker:
:)

carterkozak:
If we’re comfortable with that complexity, I’m happy with this.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
```diff
@@ -56,4 +56,8 @@ public final String toString() {
         return Long.toString(longValue());
     }
 
+    @Override
+    public final int compareTo(SafeLong other) {
+        return Long.compare(longValue(), other.longValue());
```
dansanduleac, line 61: what if other is null? do we want some special handling?

dansanduleac (approved):

conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
iamdanfox: So I think this is actually pretty unlikely due to how we've really tried to purge nulls from our codebases, and also seems OK according to the javadoc of Comparable:

```
     * @throws NullPointerException if the specified object is null
```

j-baker: yep - I don't deal much with nulls, but my general understanding is that special null handling is useful only when it makes a tricky error easier (e.g. if the NPE happens 10 lines down, in a different method). Here you're getting an NPE on line 1 anyway.

Issue #399, opened 2019-05-30
Generated code should transform markdown to html javadoc


carterkozak:
## What happened?

The conjure specification suggests that docs may use markdown, however service authors tend to use syntax for their favorite/current language instead, which leads to low quality documentation across other targets.

## What did you want to happen?

Markdown in conjure yml should be rendered into html for javadoc.

carterkozak:
Fixed by #400

PR #400, opened 2019-05-30
[improvement] Javadoc html is rendered from conjure docs markdown

carterkozak:
## Before this PR
Javadoc fields were populated with the string value from conjure definitions, resulting in issues described in #399.

## After this PR
==COMMIT_MSG==
[improvement] Javadoc html is rendered from conjure docs markdown
[fix] Generate `@param` javadoc based on ArgumentDefinition docs 
==COMMIT_MSG==

## Possible downsides?
It's possible that existing docs may be parsed into HTML in unexpected ways.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/util/Javadoc.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/util/Javadoc.java
@@ -34,15 +34,24 @@ public final class Javadoc {
             // Disable paragraph tags, prefer raw text instead. Otherwise
             // all javadoc will be wrapped in paragraphs.
             .nodeRendererFactory(context -> new CoreHtmlNodeRenderer(context) {
+
+                private boolean firstParagraph = true;
+
                 @Override
                 public void visit(Paragraph paragraph) {
-                    visitChildren(paragraph);
+                    // Exclude the first set of paragraph tags for javadoc style.
+                    if (firstParagraph) {
+                        firstParagraph = false;
+                        visitChildren(paragraph);
+                    } else {
+                        super.visit(paragraph);
+                    }
                 }
             })
             .build();
 
     public static String render(Documentation documentation) {
-        String rawDocumentation = documentation.get();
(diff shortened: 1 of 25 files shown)


rubenfiszel:
Cool feature!

Side question: what about `@param` and `@return` javadoc annotations. Should we auto-fill them or enable that with a flag?

carterkozak:
Eh, I'm not terribly concerned with those. My primary goal is to dissuade developers from writing html in the conjure docs fields!

conjure-java-core/src/main/java/com/palantir/conjure/java/util/Javadoc.java
```diff
@@ -0,0 +1,54 @@
...
+
+/** Provides utility functionality for rendering javadoc. */
+public final class Javadoc {
+
+    private static final Parser parser = Parser.builder().build();
+    private static final Renderer renderer = HtmlRenderer.builder()
+            // Escaping existing HTML may break docs that rendered fine previously.
+            .escapeHtml(false)
+            // Disable paragraph tags, prefer raw text instead. Otherwise
+            // all javadoc will be wrapped in paragraphs.
```
markelliot, line 38: Does javadoc respect line breaks? If not, think you have to have the paragraph tags (and would ideally skip the very first set).

carterkozak: The rendered javadoc/html does not respect line breaks, however applying paragraph tags would be a significant change to existing documentation in ways that may or may not make it more readable, for example if it already uses html tags specifically for javadoc.
My goal for this change was to provide support for the format we suggest in the specification without significant changes to existing generated code, for now.

carterkozak: Added support for line breaks.

carterkozak:
Added the fix for #401 to this PR since they're very closely related.

dansanduleac:
👍

PR #402, opened 2019-05-30
[fix] Fixing formatting violation to comply with baseline rules.

dsd987:
## Before this PR
There is a compile-time break caused by a formatting issues enforced by an updated version of baseline.

## After this PR
Formatting is fixed and library can compile.


carterkozak:
:+1:

PR #405, opened 2019-06-01
[improvement] Add deprecated existence and documentation to Endpoint

rubenfiszel:
## Before this PR

`Endpoint` does not contain deprecated information of existence nor its corresponding documentation

## After this PR

`Endpoint` contains deprecated information of existence nor its corresponding documentation

### DIFF
--- a/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
+++ b/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
@@ -102,7 +102,7 @@ public interface Endpoint {
 
         @CanIgnoreReturnValue
         public Builder deprecated(Optional<String> value) {
-            deprecated = Preconditions.checkNotNull(value, "name is required");
+            deprecated = Preconditions.checkNotNull(value, "deprecated is required");
             return this;
         }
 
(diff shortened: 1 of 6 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
```diff
@@ -93,12 +100,19 @@ public Builder name(String value) {
             return this;
         }
 
+        @CanIgnoreReturnValue
+        public Builder deprecated(Optional<String> value) {
+            deprecated = Preconditions.checkNotNull(value, "name is required");
+            return this;
+        }
+
         public Builder from(Endpoint endpoint) {
```
carterkozak, line 110: nit: `@CanIgnoreReturnValue`

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
```diff
@@ -49,6 +50,11 @@
     /** Simple name of the endpoint method. This data may be used for metric instrumentation. */
     String name();
 
+    /** Is present if the method is deprecated, and contains its corresponding deprecating documentation. */
+    default Optional<String> deprecated() {
```
iamdanfox, line 54: :+1: in Jersey-land, this information would be loaded using reflection, but since we're not doing any reflection it looks nice to add it here.

Slight question as to whether we should make this a wrapped string (`DeprecationDocs` or something)

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
```diff
@@ -93,12 +100,20 @@ public Builder name(String value) {
             return this;
         }
 
+        @CanIgnoreReturnValue
+        public Builder deprecated(Optional<String> value) {
+            deprecated = Preconditions.checkNotNull(value, "name is required");
```
iamdanfox:
```suggestion
            deprecated = Preconditions.checkNotNull(value, "deprecated is required");
```

iamdanfox (approved):

PR #406, opened 2019-06-04
[improvement][undertow] Add EndpointHandlerWrapper for filter-like features

rubenfiszel:
## Changes in this PR

This PR enables features that are similar to the concept of Jersey filters. Undertow is meant to be used by chaining `HttpHandler`s. The first `HttpHandler` applies logic and then pass it to the next handler. With this structure, handlers always have control over the control flow and can apply logic, before and after processing by the next handlers. This is similar to the desired behavior of jersey Filters. 

After that PR, it is possible add arbitrary behaviours pre and post processing of the underlying resource by leveraging a concept similar to Undertow's `HandlerWrapper`: the `EndpointHandlerWrapper`. It wraps the handler of an endpoint, taking potentially advantage of all informations contained in an `Endpoint`.

Below an example of an EndpointHandlerWrapper with pre and post processing.

```java
EndpointHandlerWrapper myHandlerWrapper = 
    endpoint -> 
      if (shouldWrap) {
        Optional.of(exchange -> {
          preProcessingLogic(endpoint, exchange);
          endpoint.handler().handleRequest(exchange);
          postLogic(endpoint, exchange)
          exchange.addCompletionListener(postCompletionLogic(endpoint))})
      } else {
        Optional.empty()
      }
```

- added `EndpointHandlerWrapper` as a functional interface of the form `HttpHandler wrap(Endpoint endpoint)`. **Warning** `EndpointHandlerWrapper`s are meant to be used only by library writers and webserver maintainers.
- added `addEndpointHandlerWrapperBeforeBlocking(EndpointHandlerWrapper wrapper)` to the `ConjureHandler` class to give the possibility of adding **non-blocking** `EndpointHandlerWrapper` to all registered `UndertowService`s.

The utilities to easily wrap an `UndertowService` with an `EndpointHandlerWrapper` have been purposefully not included as this is a behavior that we do not want to encourage.
One can still use the following snipper to achieve this behavior.

```java
    public static UndertowService map(UndertowService service, EndpointHandlerWrapper wrapper) {
        return runtime ->
                ImmutableList.copyOf(
                        service.endpoints(runtime).stream()
                                .map(endpoint -> Endpoints.map(endpoint, wrapper))
                                .collect(ImmutableList.toImmutableList()));
    }
```

## Possible downsides?

None


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/EndpointWrapper.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.undertow.lib;
+
+public interface EndpointWrapper {
+
+    Endpoint wrap(Endpoint endpoint);
```
markelliot: What’s the use-case for being able to totally transform Endpoint definitions?

rubenfiszel: In most case you are only interested to wrap the inner handler but in this case it would still be pretty convenient:`endpoint -> endpoint.mapHandler(handler -> { ...})`
1. It makes the api more intuitive/consistent
2. it gives us flexibility for the future if we add new fields to Endpoint
3. There are some actual case where you might be interested to rewrite the path and maybe the name (for instance if you want to have a v2 of an api that change only a few things, add "2" to the ServiceName and change the path.

carterkozak: My preference is to avoid utility functionality in conjure-undertow-lib, which should be the simplest possible target for our generated code. Similarly I don't think we want to make it convenient to wrap endpoints. It's a necessity for some scenarios, but also something to be avoided unless there are no other options.

The problem we need to solve is providing a replacement for servlet and jersey filter mechanisms, for example providing a filter which can respond with 503 until the service reports that it has been initialized.

Something along these lines will be more specific:
```java
interface EndpointFilter {

    /** 
     * Returns a handler wrapping {@link Endpoint#handler()} if the given endpoint is matched,
     * otherwise an empty optional if this filter does not apply to the endpoint.
     */
    Optional<HttpHandler> filter(Endpoint endpoint);
}
```

We may want to implement that in a new module in conjure-java-runtime-api so we can avoid exposing an api dependency on conjure-undertow-runtime

rubenfiszel: I have mixed feelings about this (not the part about the package location but EndpointFilter).

It seems weird to me that we call it a `Filter` which is a jersey concept when this is just a plain HandlerWrapper. It also requires consumers to infer that they are expected to wrap the Endpoint#handler() but actually nothing forbid them to just return a completely non-related `HttpHandler`

carterkozak: It maps closely to a jersey DynamicFeature, which is one of the most powerful concepts from jersey. It makes it easy for implementors to pre-compute more than a standard servlet filter, which is what I'm going for. I think it's a bit dangerous if we return `Endpoint` because it allows implementors to remove the deprecated field or modify the http methd. In the future if we add new functionality to Endpoint, we may need to expand the filter/wrapper concept, but I think that is likely to be safer than allowing full modification of the Endpoint.

Agreed that there's nothing to forbid returning a completely different handler, but that's no different from returning an Endpoint. There's no way around this if we want to allow try/finally blocks around the implementation, which allows us to code more safely.

rubenfiszel: I still don't see the benefits of mapping to a jersey concept when we have the huge elegance in conjure-undertow of being able to represent Endpoint and UndertowService as nice static data containers that can be seen as Functors and for whose the map function is a natural extension.

The so-called filters on a Service is easier to understand as a wrapper to all endpoints than as "the equivalent of the jersey concept of applying pre and post subroutine to jax-rs resources".

The most natural signature of an Endpoint transformation is `Endpoint -> Endpoint`. It is very clear what it is doing and it seems unlikely that users would be confused by its use. If they really want to change the http method, it is weird, but it is also ultimately their choice. There is no way they would change the http method without realising it.

That being said, if it really bother you, could we consider transforming `EndpointWrapper` to `EndpointHandlerWrapper { HandlerWrapper wrap(Endpoint) }` and UndertowService to 

```
UndertowService {
   ...
   UndertowService mapHandler(EndpointHandlerWrapper) 
}
```

rubenfiszel:
@carterkozak I probably have to add tests but what do you think of the refactor?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -32,44 +32,59 @@
...
- * among {@link ConjureHandler#WRAPPERS other features}.
+ * among {@link ConjureHandler#WRAPPERS_BEFORE_BLOCKING features before blocking} and
+ * {@link ConjureHandler#WRAPPERS_AFTER_BLOCKING features after blocking}.
  */
 public final class ConjureHandler implements HttpHandler {
 
-    private static final ImmutableList<BiFunction<Endpoint, HttpHandler, HttpHandler>> WRAPPERS =
-            ImmutableList.<BiFunction<Endpoint, HttpHandler, HttpHandler>>of(
+    private ImmutableList<EndpointHandlerWrapper> wrappersJustBeforeBlocking =
+            ImmutableList.of();
```
carterkozak: nit: I'd prefer for this to be ordered after static fields.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -83,11 +98,19 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
     }
 
     private void register(Endpoint endpoint) {
-        HttpHandler current = endpoint.handler();
-        for (BiFunction<Endpoint, HttpHandler, HttpHandler> wrapper : WRAPPERS) {
-            current = wrapper.apply(endpoint, current);
+        Endpoint current = endpoint;
+        current = stackEndpointHandlerWrapper(current, WRAPPERS_AFTER_BLOCKING);
+        current = stackEndpointHandlerWrapper(current, wrappersJustBeforeBlocking.reverse());
+        current = stackEndpointHandlerWrapper(current, WRAPPERS_BEFORE_BLOCKING);
```
carterkozak: What if we construct a single EndpointHandlerWrapper in `Builder.build()` that we use once here?

```java
private void register(Endpoint endpoint) {
    routingHandler.add(endpoint.method(), endpoint.template(), wrapper.wrap(current));
}
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/EndpointHandlerWrapper.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.palantir.conjure.java.undertow.runtime;
+
+import com.palantir.conjure.java.undertow.lib.Endpoint;
+import io.undertow.server.HandlerWrapper;
+
+public interface EndpointHandlerWrapper {
+
+    /** This should return the {@link HandlerWrapper} that should be applied to the {@link Endpoint#handler()}. **/
+    HandlerWrapper wrap(Endpoint endpoint);
```
carterkozak: No reason for one wrapper to return a second wrapper.
```suggestion
    HttpHandler wrap(Endpoint endpoint);
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -32,44 +32,59 @@
...
+            endpoint -> handler -> new ConjureExceptionHandler(handler)
     ).reverse();
 
+    public ConjureHandler addHandlerWrapperBeforeBlocking(EndpointHandlerWrapper wrapper) {
+        wrappersJustBeforeBlocking = ImmutableList.<EndpointHandlerWrapper>builder()
+                .addAll(wrappersJustBeforeBlocking)
+                .add(wrapper)
+                .build();
+        return this;
+    }
```
carterkozak: ConjureHandler is meant to be immutable, configuration should occur on the builder

Maybe call the builder method `initialNonBlockingEndpointHandlerWrappers(EndpointHandlerWrapper)` with strongly worded documentation warning against performing _any_ blocking operations in this wrapper. We should also call out the order we expect them to be invoked in.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Endpoints.java
```diff
@@ -0,0 +1,36 @@
...
+
+public final class Endpoints {
+
+    private Endpoints() {}
+
+    /**
+     * Build am {@link Endpoint} who is a copy of the {@link Endpoint} but whose handler
+     * has been wrapped by an {@link HandlerWrapper}.
+     */
+    public static Endpoint mapHandler(Endpoint endpoint, HandlerWrapper wrapper) {
```
carterkozak: Thoughts on naming this method `wrap`? It doesn't exclude us from eventually adding a `wrap` method for the full endpoint because the parameter types will differ.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Endpoints.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.undertow.runtime;
+
+import com.palantir.conjure.java.undertow.lib.Endpoint;
+import io.undertow.server.HandlerWrapper;
+
+public final class Endpoints {
```
carterkozak:
```suggestion
/** Utility functionality for conjure {@link Endpoint} objects. */
public final class Endpoints {
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/UndertowServices.java
```diff
@@ -0,0 +1,39 @@
...
+     * {@link io.undertow.server.HandlerWrapper} resulting of the application of each
+     * {@link com.palantir.conjure.java.undertow.lib.Endpoint} to an
+     * {@link EndpointHandlerWrapper}.
+     */
+    public static UndertowService map(UndertowService service, EndpointHandlerWrapper wrapper) {
+        return runtime ->
+                ImmutableList.copyOf(
+                        service.endpoints(runtime).stream()
+                                .map(endpoint -> Endpoints.mapHandler(endpoint, wrapper.wrap(endpoint)))
+                                .collect(Collectors.toList()));
```
carterkozak: nit: remove ImmutableList.copyOf in favor of:

```suggestion
                                .collect(ImmutableList.toImmutableList()));
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -32,44 +32,59 @@
...
+    // will not propagate to the wrapped service.
+    private static final ImmutableList<EndpointHandlerWrapper> WRAPPERS_AFTER_BLOCKING =
+            ImmutableList.<EndpointHandlerWrapper>of(
+            endpoint -> handler -> new BlockingHandler(handler),
+    // Logging context and trace handler must execute prior to the exception
+    // to provide user and trace information on exceptions.
+            endpoint -> handler -> new LoggingContextHandler(handler),
+            endpoint -> handler -> new TracedOperationHandler(
+            handler, endpoint.method() + " " + endpoint.template()),
+            endpoint -> handler -> new ConjureExceptionHandler(handler)
```
carterkozak: It's probably time to move these into the build method to apply custom wrappers while maintaining readability.
That way we can build a single list of wrappers and apply additional configured wrappers between the right ones in one place.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -32,44 +32,59 @@
...
+import java.util.Collection;
 import java.util.List;
 import java.util.Set;
-import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
 /**
  * Conjure routing mechanism which can be registered as an Undertow {@link HttpHandler}.
  * This handler takes care of exception handling, tracing, and web-security headers
- * among {@link ConjureHandler#WRAPPERS other features}.
```
carterkozak, line 43: nit: I'd just remove the link from this, it's not important to call out which features happen at what point in the lifecycle because that's an implementation detail.

carterkozak (commented):
Needs tests for `Endpoints` and `UndertowServices` utility functions, especially since `UndertowServices` isn't otherwise used.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/EndpointHandlerWrapper.java
rubenfiszel: I think it's actually better with an `HandlerWrapper` because then what it is actually returning is a "Filter" instead of a "Filtered Handler".

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Endpoints.java
rubenfiszel: wrap seems to me to only makes sense when you are returning the full object wrapped itself.
The `map` here comes from the fact that Endpoint is a `Functor` (actually a x-functor where x is the number of fields). I like `map` here :)

carterkozak: In that case can we rename the method from `mapHandler` to `map` since the paramter types tell us that it's only capable of mapping handlers?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/EndpointHandlerWrapper.java
carterkozak: I don't agree, the `Endpoint` is a single instance of a single endpoint, there's no reason to wrap multiple handlers with the same wrapper. Returning the handler directly results in an API that's easier to understand.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
rubenfiszel: done

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
rubenfiszel: Added doc, but kept the name.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Endpoints.java
rubenfiszel: Sure.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
rubenfiszel: done

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/EndpointHandlerWrapper.java
rubenfiszel: I don't agree but I did it anyway

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Endpoints.java
rubenfiszel: I thought it was less confusing with `mapHandler` but renamed to `map`

markelliot:
would it be possible to get a top-line description of why we want this functionality, either on the PR, in docs or in code?

rubenfiszel:
@markelliot updated the title and description

markelliot:
@rubenfiszel thanks! The description sounds pretty great, but isn't quite what the code gets us right now. Could we do something like this?
```
interface EndpointHandlerWrapper<C> {
    C beforeExecution(Endpoint endpoint, HttpServerExchange exchange);
    C afterExecution(C context, Endpoint endpoint, HttpServerExchange exchange);
    C completion(C context, Endpoint endpoint);
}
```

Somewhat inspired by https://github.com/palantir/tritium/blob/develop/tritium-api/src/main/java/com/palantir/tritium/event/InvocationEventHandler.java.

The idea is that it makes it impossible to mess up/not call the original handler, still gives options to code an [aspect](https://en.wikipedia.org/wiki/Aspect-oriented_programming).

If we're after filters and want the ability to stop processing, may want to have something that's more explicitly a predicate `boolean shouldExecute(...)`.

carterkozak:
@markelliot I'm somewhat skeptical of that approach because it doesn't allow us to try/finally around execution, which can lead to leaks.

While I agree that it may be cleaner for some functionality, it has drawbacks for others. I want to avoid over-engineering a new api unless it's a strict improvement or would solve problems we have encountered.

A few benefits of the current proposal:
1. It works similarly to other undertow constructs, allowing for simple integration with existing components.
2. We can choose to selectively wrap endpoints, leaving others completely unmodified.
3. The endpoint is provided to the wrapper, not necessarily to each request, this way implements are pushed to do setup once on initialization, not for every request.

Given that, I would propose that we build the aop extension in a way that it can produce a `EndpointHandlerWrapper` (`HttpHandler wrap(Endpoint)`). Our WC API may expose only the aop approach for witchcraft consumers to filter on, while leaving itself the more expressive and potentially more dangerous wrapper api.

markelliot:
The try/finally semantics are exactly why one would do this — and it’s otherwise not possible to guarantee that all implementers will correctly perform that kind of wrapping unless we’re a little more structured. 

The current proposal also seems way more likely to leak things like Endpoint (my sample was directional, drop the args you don’t want to have), and things like selective application can easily be solved with an additional predicate. 

I’m not a fan of the big hammer of being able to wholesale replace an Endpoint, and don’t think this is the appropriate way to allow developers to add cross-cut functionality.

rubenfiszel:
@markelliot I share the concerns of @carterkozak .The experience we have had with WC is that the Undertow model of HttpHandler/HandlerWrapper is pleasant to write in, concise/readable and give us sufficient flexibility to achieve a diverse set of features so far without having us to go through any escape hatch.

One of the goal of having your proposed interface is to make it "impossible to mess up/not call the original handler". However, in some cases, the control that you would want to apply is precisely to not call the original handler. 

That being said, I agree that onSuccess, onFailure is a better try/catch replacement and that your proposal bring more safety and structure. I think in an ideal world, what we would want is to be able to use the PR's EndpointHandlerWrapper for WC's internals and core libraries, but to force other final consumers to go through your proposed interface of EndpointHandlerWrapper if they desire to implement their own filter-like logic. 

It is possible to achieve some of this desired goal by taking advantage of the fact that your proposed EndpointHandlerWrapper can extend the PR's one and that we can strongly recommend  final consumers to extend your more structured interface when implementing their custom filter logic.

markelliot:
Okay, perhaps we could make clear the possible downsides of the approach proposed here:
* Implementors may fail to correctly call the delegate handler (docs presently fail to make this clear)
* Implementors will likely assume that the delegate handler will never throw
* Implementations of wrappers can wholesale replace the existing HttpHandler, and implementors can abuse this to modify the behavior of Conjure endpoints arbitrarily
* Implementations of wrappers are not guided to separate initialization logic from per-endpoint logic (docs presently fail to make this clear)
* Implementors of wrappers may not consider whether their implementation should always wrap or sometimes wrap (not only do the docs presently fail to make this clear, but @rubenfiszel recently submitted an internal proposal which failed to do this)
* We're exposing a very general purpose override, and once exposed, we will not be able to remove this override easily, even if it is used poorly or abusively

rubenfiszel:
@markelliot @carterkozak @iamdanfox 
Updated the PR and PR description, changed the `EndpointHandlerWrapper` to `Endpoint -> HttpHandler`. Removed the utilities from being usable by users. Made the doc more explicit about the use and dangers of `EndpointHandlerWrapper`.
Ready for re-review.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Endpoints.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.undertow.runtime;
+
+import com.palantir.conjure.java.undertow.lib.Endpoint;
+
+/** Utility functionality for conjure {@link Endpoint} objects. */
+public final class Endpoints {
```
carterkozak: should be package private

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowConjureFilter.java
```diff
@@ -0,0 +1,36 @@
...
+    default HttpHandler wrap(HttpHandler next) {
+        return exchange -> handle(
+                exchange,
+                () -> {
+                    next.handleRequest(exchange);
+                    return null;
+                });
+    }
+    void handle(HttpServerExchange exchange, Callable<Void> callNextHandler);
+}
```
carterkozak: unused class

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/EndpointHandlerWrapper.java
```diff
@@ -0,0 +1,40 @@
...
+     * incorrectly. Make sure to not forget passing the request to the next handlers through
+     * calling the endpoint's {@link HttpHandler#handleRequest(HttpServerExchange)}.
+     * This function is used to implement filter-like behavior:
+     * - Adding logic pre and post handling by the next handlers, by adding the logic before and after
+     *  {@link HttpHandler#handleRequest(HttpServerExchange)}
+     * - Adding logic post completion of the request by all handlers
+     * {@link io.undertow.server.HttpServerExchange#addExchangeCompleteListener(ExchangeCompletionListener)}
+     * - Catching and handling exceptions surrounding {@link HttpHandler#handleRequest(HttpServerExchange)} with
+     * try ... catch ... finally.
+     **/
```
carterkozak: Might want to move this javadoc to the class level, and provide a concise description on the method `/** May wrap {@link Endpoint#handler()} if this wrapper applies, otherwise empty. */`

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -96,6 +74,52 @@ public static Builder builder() {
...
+         * For blocking operations, please wrap the UndertowService thenmselves
+         * If you call this multiple time, the last wrapper will be applied first, meaning it will be wrapped the
+         * previously added {@link EndpointHandlerWrapper}s.
+         */
+        public Builder addEndpointHandlerWrapperBeforeBlocking(EndpointHandlerWrapper wrapper) {
+            wrappersJustBeforeBlocking = ImmutableList.<EndpointHandlerWrapper>builder()
+                    .addAll(wrappersJustBeforeBlocking)
+                    .add(wrapper)
+                    .build();
+            return this;
```
carterkozak: Can we use a mutable list in the builder and add the wrapper here?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -96,6 +74,52 @@ public static Builder builder() {
...
+        private ImmutableList<EndpointHandlerWrapper> wrappersJustBeforeBlocking =
+                ImmutableList.of();
+
+        /**
+         * This MUST only be used for non-blocking operations that are meant to be run on the io-thread.
+         * For blocking operations, please wrap the UndertowService thenmselves
+         * If you call this multiple time, the last wrapper will be applied first, meaning it will be wrapped the
+         * previously added {@link EndpointHandlerWrapper}s.
+         */
+        public Builder addEndpointHandlerWrapperBeforeBlocking(EndpointHandlerWrapper wrapper) {
```
carterkozak: Mind adding a test for this?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -128,7 +152,23 @@ public Builder fallback(HttpHandler value) {
 
         public HttpHandler build() {
             checkOverlappingPaths();
-            return new ConjureHandler(fallback, endpoints);
+            EndpointHandlerWrapper current = endpoint -> Optional.empty();
+            current = stackEndpointHandlerWrapper(current, WRAPPERS_AFTER_BLOCKING.reverse());
+            current = stackEndpointHandlerWrapper(current, wrappersJustBeforeBlocking);
+            current = stackEndpointHandlerWrapper(current, WRAPPERS_BEFORE_BLOCKING.reverse());
```
carterkozak: Can we inline creation of a single List<EndpointHandlerWrapper> in this method, removing both `WRAPPERS_AFTER_BLOCKING` and `WRAPPERS_BEFORE_BLOCKING`? That way I think it will be easier to follow the order of operations.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -32,48 +32,27 @@
...
+    private final EndpointHandlerWrapper stackedEndpointHandlerWrapper;
     private final RoutingHandler routingHandler;
 
-    private ConjureHandler(HttpHandler fallback, List<Endpoint> endpoints) {
+    private ConjureHandler(
+            HttpHandler fallback,
+            List<Endpoint> endpoints,
+            EndpointHandlerWrapper stackedEndpointHandlerWrapper) {
         this.routingHandler = Handlers.routing().setFallbackHandler(fallback);
+        this.stackedEndpointHandlerWrapper = stackedEndpointHandlerWrapper;
```
carterkozak: Can we avoid providing the wrapper list to the ConjureHandler, and instead wrap endpoints in `build()`?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Endpoints.java
rubenfiszel: done

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowConjureFilter.java
rubenfiszel: done

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/EndpointHandlerWrapper.java
rubenfiszel: done

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
rubenfiszel: done

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
rubenfiszel: done

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
rubenfiszel: done

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureHandlerTest.java
```diff
@@ -80,6 +102,17 @@ public void after() {
         server.stop();
     }
 
+    @Test
+    public void invokesWrapperBeforeBlocking() {
+        when(wrapperObserver.control()).thenReturn(1);
+        execute();
+        //check that the first wrapper (the one that adds 1) is called before the one that adds 2.
+        assertThat(ImmutableList.copyOf(wrappersBeforeBlockingCallOrder).reverse())
```
carterkozak: isn't this backwards?

rubenfiszel: In what sense? Adding to the list add to the front so you have to reverse it to get the original order

carterkozak: https://docs.oracle.com/javase/8/docs/api/java/util/List.html#add-E-

> Appends the specified element to the end of this list

carterkozak:
Pushed commit https://github.com/palantir/conjure-java/pull/406/commits/00820bb40d3e68e6d221e43dc054f2c8a31ec8cf
Thoughts on that? I like having all of the wrappers in a single place.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureHandlerTest.java
rubenfiszel: I was so sure lists were doing append so that add would be O(1). Good to know!
Your fix is right then.

rubenfiszel:
@carterkozak LGTM. I did a small aesthetic change of using `add(E ...)`

rubenfiszel:
I'm just worried that it's less clear in this structure that the pre-defined ConjureHandler `EndpointHandlerWrapper`s are consts.

carterkozak:
👍

PR #407, opened 2019-06-04
Fix broken readme link to UndertowService

tdeitch:
## Before this PR
Link points to `Service.java`, which does not exist.

## After this PR
==COMMIT_MSG==
Fix broken readme link to UndertowService
==COMMIT_MSG==

## Possible downsides?
I could have linked to the wrong file.


carterkozak:
Thanks! 👍

PR #409, opened 2019-06-05
[improvement] conjure-undertow binary deserialization exception matches json

carterkozak:
Previously a standard exception was used rather than a
FrameworkException. Now both produce the correct 415 status code.

## After this PR
==COMMIT_MSG==
conjure-undertow binary deserialization exception matches json
==COMMIT_MSG==

## Possible downsides?
none


carterkozak:
Not likely to make much difference, just a minor issue I noticed while looking at something else.

rubenfiszel (approved):

PR #412, opened 2019-06-07
[fix] Undertow generator properly handles query params of optional of alias of external

carterkozak:
## Before this PR
Previously the generator would throw because it expected a Type to
be an alias when it was an external reference.

## After this PR
==COMMIT_MSG==
Undertow generator properly handles query params of optional of alias of external
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -772,7 +772,7 @@ final class UndertowServiceHandlerGenerator {
             //   * optional<alias that resolves to a primitive>
             //   * alias that follows one of these rules (recursive definition)
             Type aliasedType = UndertowTypeFunctions.getAliasedType(inType, typeDefinitions);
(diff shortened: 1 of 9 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -772,7 +772,7 @@ private static CodeBlock createConstructorForTypeWithReference(Type inType, Stri
             //   * optional<alias that resolves to a primitive>
             //   * alias that follows one of these rules (recursive definition)
             Type aliasedType = UndertowTypeFunctions.getAliasedType(inType, typeDefinitions);
-            if (aliasedType.accept(TypeVisitor.IS_PRIMITIVE)) {
+            if (aliasedType.accept(TypeVisitor.IS_PRIMITIVE) || aliasedType.accept(MoreVisitors.IS_EXTERNAL)) {
```
carterkozak, line 775: this line is the fix, the rest is generated code/testing

iamdanfox (approved):
Thanks @carterkozak!

PR #425, opened 2019-06-24
[improvement] conjure-undertow endpoints provide tracing spans around body serde

carterkozak:
## Before this PR
No differentiation between resource execution, and I/O + serde 

## After this PR
==COMMIT_MSG==
conjure-undertow endpoints provide tracing spans around body serde
==COMMIT_MSG==

## Possible downsides?
Up to 2 additional spans for every sampled call (only if request/response body is sent)

### DIFF
--- a/changelog/@unreleased/pr-425.v2.yml
+++ b/changelog/@unreleased/pr-425.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: conjure-undertow endpoints provide tracing spans around body serde.
+  links:
+  - https://github.com/palantir/conjure-java/pull/425
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
@@ -129,7 +129,7 @@ final class ConjureBodySerDe implements BodySerDe {
 
         EncodingSerializerContainer(Encoding encoding, TypeMarker<T> token) {
             this.encoding = encoding;
-            this.serializer = encoding.serializer(token);
+            this.serializer = TracedEncoding.wrap(encoding).serializer(token);
         }
     }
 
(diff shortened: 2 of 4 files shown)


markelliot:
I’d be concerned that always tracing this will result in a very significant increase in the total number of spans we collect, and so as a result it’s not clear that this should be the default setup.

rubenfiszel:
Agreed with @markelliot that providing it as default without any possibility to opt-out might be a bit strong. What about making it a builder setting flag in UndertowRuntime that we would opt-in for WC?

carterkozak:
I've gotten a lot of questions about where time is spent on the server between when the client sends a request, and resource implementation. This PR aims to provide that data.

Looking at the first sample I can find from a traced OkHttp invocation, there appear to be six unique spans:
`OkHttp: execute-run`
`OkHttp: POST /users`
`OkHttp: execute-enqueue`
`OkHttp: acquire-limiter-enqueue`
`OkHttp: dispatcher`
`OkHttp: acquire-limiter-run`

If we're concerned about creating too much trace information, perhaps we should either:
1) Reduce our configured/default sample rate slightly to hit a target rate
2) Remove some client spans in favor of balancing traced operations with the server
3) Support something like log levels for tracer (I don't think this is a good idea, it would make both development and usage more complicated than necessary).

rubenfiszel:
That's fair. I agree with you that providing this info should be the default for all our services. I'm less worried about the rate of tracing than the deep coupling that it entails if this is hardcoded in conjure-java.
The reasoning for me to make it opt-in is that it is not required as a core feature. The opinionated choices to provide it by default to all our services make more sense to me when located in Witchcraft by passing a flag to the ConjureUndertowRuntime.

rubenfiszel (approved):
Thinking about it more, changed my mind, this should be the default without any flag.

iamdanfox (approved):
I think I'm in favour of this too, would be nice to have data that can rule out serialization as a cause of slowness when debugging support tickets. I think this is also low risk because:

- only O(10) services use conjure-undertow generated handlers (most are still on jersey), so it won't double our trace rate
- trace-log-viewer also has functionality to auto-hide short spans, so if serialization is frequently trivial it shouldn't cause too much visual noise
- the non-sampled codepath for tracing is now more optimized (from [2.16.0 onwards](https://github.com/palantir/tracing-java/releases/tag/2.16.0)), so this should impose low overhead in the 99% of cases

PR #446, opened 2019-07-11
Deserialize null fields as empty collections

ferozco:
## Before this PR
While the [Conjure wire spec](https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#561-coercing-json-null--absent-to-conjure-types) specified that absent _or_ null JSON keys for collections (`list`, `set`, `map`, `optional`) should be coerced into their empty variants conjure-java implementation only correctly handled absent fields and would fail when attempting to coerce an explicit null.

This issue had not previously been encountered since conjure-java servers always respond with empty collections (`[]`, `{}`) or omit empty optionals. Recently, the issue surfaced since Golang server's serialize empty collections as null.

## After this PR
==COMMIT_MSG==
Correctly coerce null fields into empty collections
==COMMIT_MSG==

## Possible downsides?
It appears as if these cases had been explicitly omitted from the [verifier](https://github.com/palantir/conjure-verification/blob/develop/master-test-cases.yml#L179). I'm not sure why they were omitted so there may a downside I'm not aware of. 

We will want to clarify why the test cases were omitted and potentially re-introduce them

### DIFF
--- a/changelog/@unreleased/pr-446.v2.yml
+++ b/changelog/@unreleased/pr-446.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Correctly coerce null fields into empty collections
+  links:
+  - https://github.com/palantir/conjure-java/pull/446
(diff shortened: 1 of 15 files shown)


CRogers (approved):

dansanduleac:
:+1:

PR #465, opened 2019-07-24
[improvement] Adds a visitor builder for union type visitors

aioobe:
## Before this PR
Visitors are used to do case splits on union types. For example

    String sound = pet.accept(new Animal.Visitor<String>() {

        @Override
        public String visitDog(String name) {
            return "Woof!";
        }

        @Override
        public String visitCat(String name) {
            return "Meow!";
        }

        @Override
        public String visitUnknown(String unknownType) {
            return "beep?";
        }
    });

There's a lot of cruft here unfortunately. We have basically three semantically interesting lines, out of 17 lines in total. Judging from a big code base that uses Conjure, short handler methods are quite common, so the signal-to-noise ratio is quite low in practice.

## After this PR
The `UnionGenerator` now includes a builder for the `Visitor` class, which allows us to do:

    String sound = pet.accept(Animal.Visitor.<String>builder()
            .dog(name -> "Woof!")
            .cat(name -> "Meow!")
            .unknown(unknownType -> "beep?"));

The solution uses staged builders so case omissions are caught in compile time (thanks @carterkozak). A single builder object is used through out the build, and the separate stages are enforced simply through interfaces (thanks @dxiao).

## Possible downsides?
None that I can think of.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
@@ -13,6 +13,7 @@ import com.palantir.logsafe.Preconditions;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Function;
 import javax.annotation.Generated;
 
 @Generated("com.palantir.conjure.java.types.UnionGenerator")
@@ -72,6 +73,40 @@ public final class SingleUnion {
         T visitFoo(String value);
 
         T visitUnknown(String unknownType);
+
+        static <T> FooStageVisitorBuilder<T> builder() {
+            return new VisitorBuilder<T>();
+        }
+    }
+
+    private static class VisitorBuilder<T>
+            implements FooStageVisitorBuilder<T>, UnknownStageVisitorBuilder<T> {
+        private Function<String, T> fooVisitor;
+
+        public UnknownStageVisitorBuilder<T> foo(Function<String, T> fooVisitor) {
+            this.fooVisitor = fooVisitor;
+            return this;
+        }
+
+        public Visitor<T> unknown(Function<String, T> unknownVisitor) {
+            return new Visitor<T>() {
+                public T visitFoo(String value) {
+                    return fooVisitor.apply(value);
+                }
+
+                public T visitUnknown(String value) {
+                    return unknownVisitor.apply(value);
+                }
+            };
+        }
+    }
+
+    public interface FooStageVisitorBuilder<T> {
+        UnknownStageVisitorBuilder<T> foo(Function<String, T> fooVisitor);
+    }
+
+    public interface UnknownStageVisitorBuilder<T> {
+        Visitor<T> unknown(Function<String, T> unknownVisitor);
     }
 
     @JsonTypeInfo(
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
@@ -15,7 +15,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.function.Function;
 import javax.annotation.Generated;
-import javax.annotation.Nonnull;
 
 @Generated("com.palantir.conjure.java.types.UnionGenerator")
 public final class SingleUnion {
@@ -88,14 +87,13 @@ public final class SingleUnion {
 
         private Function<String, T> unknownVisitor;
 
-        public UnknownStageVisitorBuilder<T> foo(@Nonnull Function<String, T> fooVisitor) {
+        public UnknownStageVisitorBuilder<T> foo(Function<String, T> fooVisitor) {
             Preconditions.checkNotNull(fooVisitor, "fooVisitor cannot be null");
             this.fooVisitor = fooVisitor;
             return this;
         }
 
-        public CompletedStageVisitorBuilder<T> unknown(
-                @Nonnull Function<String, T> unknownVisitor) {
+        public CompletedStageVisitorBuilder<T> unknown(Function<String, T> unknownVisitor) {
             Preconditions.checkNotNull(unknownVisitor, "unknownVisitor cannot be null");
             this.unknownVisitor = unknownVisitor;
             return this;
@@ -115,11 +113,11 @@ public final class SingleUnion {
     }
 
     public interface FooStageVisitorBuilder<T> {
-        UnknownStageVisitorBuilder<T> foo(@Nonnull Function<String, T> fooVisitor);
+        UnknownStageVisitorBuilder<T> foo(Function<String, T> fooVisitor);
     }
 
     public interface UnknownStageVisitorBuilder<T> {
-        CompletedStageVisitorBuilder<T> unknown(@Nonnull Function<String, T> unknownVisitor);
+        CompletedStageVisitorBuilder<T> unknown(Function<String, T> unknownVisitor);
     }
 
     public interface CompletedStageVisitorBuilder<T> {
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
@@ -80,15 +80,26 @@ public final class SingleUnion {
     }
 
     private static class VisitorBuilder<T>
-            implements FooStageVisitorBuilder<T>, UnknownStageVisitorBuilder<T> {
+            implements FooStageVisitorBuilder<T>,
+                    UnknownStageVisitorBuilder<T>,
+                    CompletedStageVisitorBuilder<T> {
         private Function<String, T> fooVisitor;
 
+        private Function<String, T> unknownVisitor;
+
         public UnknownStageVisitorBuilder<T> foo(Function<String, T> fooVisitor) {
+            Preconditions.checkNotNull(fooVisitor, "fooVisitor cannot be null");
             this.fooVisitor = fooVisitor;
             return this;
         }
 
-        public Visitor<T> unknown(Function<String, T> unknownVisitor) {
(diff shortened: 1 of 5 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
```diff
@@ -88,6 +89,73 @@ public String toString() {
...
+            this.barVisitor = barVisitor;
+            return this;
+        }
+
+        public UnknownStageVisitorBuilder<T> baz(Function<Long, T> bazVisitor) {
+            this.bazVisitor = bazVisitor;
+            return this;
+        }
+
+        public Visitor<T> unknown(Function<String, T> unknownVisitor) {
```
rubenfiszel: clever shorthand but I would still separate `unknown` from the `build` method

rubenfiszel: I would also `checkNotNull` for every methods in the build method

conjure-java-core/src/integrationInput/java/com/palantir/product/UnionTypeExample.java
```diff
@@ -124,6 +125,140 @@ public String toString() {
...
+                }
+
+                public T visitUnknown(String value) {
+                    return unknownVisitor.apply(value);
+                }
+            };
+        }
+    }
+
+    public interface StringExampleStageVisitorBuilder<T> {
```
rubenfiszel: this seems like a cumbersome pattern. You probably want to only implement a `from` method. What is the use-case you had in mind?

aioobe: Could you clarify? How would that `from` method look?

aioobe: Do you mean something like the `of(...)` method for immutables? Like

    Visitor<String> from(Function<String, String> catVisitor,
                         Function<String, String> dogVisitor,
                         Function<String, String> unknownVisitor) {
        ...
    }

?

rubenfiszel:
Use @NonNullable but otherwise LGTM

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -190,30 +201,310 @@ private static MethodSpec generateEquals(ClassName unionClass) {
...
+                .addFields(allVisitorBuilderFields(memberTypeMap, visitResultType))
+                .addMethods(allVisitorBuilderSetters(enclosingClass, visitResultType, memberTypeMap))
+                .addMethod(builderBuildMethod(visitor, visitResultType, memberTypeMap))
+                .build();
+    }
+
+    /**
+     * Generates setter methods for the Visitor builder on the following format.
+     *
+     * NextMemberStage&lt;T&gt; memberName(Function&lt;MemberType, T&gt; memberVisitor) {
```
uschi2000, line 267: needs some `<pre>`, no?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -190,30 +201,310 @@ private static MethodSpec generateEquals(ClassName unionClass) {
...
+     *     this.memberVisitor = memberVisitor;
+     *     return this;
+     * }
+     *
+     * @param unionClass The enclosing class
```
uschi2000: unless this has changed recently, the code style in this repo favours explanation of parameters as part of the text. IMO, it's also fine to keep these pretty short in single-use internal/private methods.

uschi2000 (commented):
I generally like the idea, thanks! Few questions:
(1) Do you want to eventually deprecate&remove the interface-based visitors?
(2) What does the stack trace look like with the builder-based approach?
(3) People often use partial abstract implements of visitors, e.g., `AlwaysFailVisitor` or `DefaultFooVisitor` and then selectively override some of the methods. It looks like that doesn't work well with the staged builders introduced here, so the answer to (1) should be "no"?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
aioobe: Good points. PR updated accordingly.

(1+3) Right, the interface is probably good to keep. It allows handler classes to do things like `something.accept(this)` and it probably makes it slightly nicer to implement more complex stateful handlers.

(2) A sample stack trace...

    Exception in thread "main" java.lang.RuntimeException: test
        at conjuretesting.Main.lambda$main$1(Main.java:34)
        at org.aioobe.TestAnimal$VisitorBuilder$1.visitCat(TestAnimal.java:126)
        at org.aioobe.TestAnimal.accept(TestAnimal.java:45)
        at conjuretesting.Main.main(Main.java:32)

...where the top stack entry points at the throw statement in the lambda expression. So quite good I'd say.

Happy to do another iteration if anything else comes up. Otherwise, what would the next step be? How do I deal with the changelog check?

uschi2000: Either add a changelog entry manually or click the `Generate changelog entry` box towards the top of the comment section.

uschi2000: I'm happy with this PR, but let's give @iamdanfox a chance to take a look.

conjure-java-core/src/integrationInput/java/com/palantir/product/UnionTypeExample.java
```diff
@@ -124,6 +125,161 @@ public String toString() {
...
+
+        private Function<Integer, T> thisFieldIsAnIntegerVisitor;
+
+        private Function<Integer, T> alsoAnIntegerVisitor;
+
+        private Function<Integer, T> ifVisitor;
+
+        private Function<Integer, T> newVisitor;
+
+        private Function<Integer, T> interfaceVisitor;
```
carterkozak: Should we use IntFunction and DoubleFunction for int and double primitive types?

ferozco: Agree with carter here, we avoid boxing in other places and it would be nice to do so here as well

ferozco:
Overall this looks great! Thanks for the contribution @aioobe. Could you address @carterkozak comment and then we'll get this in

aioobe:
Added special handling for ints and doubles as suggested by @carterkozak. Also added `@Nonnull` annotations as suggested by @rubenfiszel.

I noticed there's also `SAFELONG`. Do we want to use `LongFunction` as well? (And unpack values with `SafeLong.longValue`.) Personally, I'd say yes, but I don't know the exact use cases for `SAFELONG` and why the API doesn't expose it as a `long`.

We could also get away without boxing booleans, but there's no `BooleanFunction` so we'd have to include such a functional interface ourselves. Worth it? Personally I'd say, naah, not worth it.

Happy to hear your thoughts.

carterkozak:
> I noticed there's also SAFELONG. Do we want to use LongFunction as well

No, `Function<SafeLong, T>` is more accurate. We should represent types as bestwe can without doing anything extra or clever.

> We could also get away without boxing booleans, but there's no BooleanFunction so we'd have to include such a functional interface ourselves. Worth it? Personally I'd say, naah, not worth it.

I agree. If we had a standard interface available we should use it, but given it doesn't exist, it's fine not to.

conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
```diff
@@ -87,13 +88,14 @@ public String toString() {
 
         private Function<String, T> unknownVisitor;
 
-        public UnknownStageVisitorBuilder<T> foo(Function<String, T> fooVisitor) {
+        public UnknownStageVisitorBuilder<T> foo(@Nonnull Function<String, T> fooVisitor) {
```
carterkozak: Do we use Nonnnull anotations elsewhere? I thought we assumed non-null by default unless otherwise specified (and no public conjure components allow nulls).

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -395,10 +398,16 @@ private static ClassName visitorStageInterfaceName(ClassName unionClass, String
...
      */
     private static TypeName visitorObjectTypeName(TypeName memberType, TypeName visitResultType) {
-        return ParameterizedTypeName.get(
-                ClassName.get(Function.class),
-                memberType.box(),
-                visitResultType);
+        if (memberType.equals(TypeName.INT)) {
+            return ParameterizedTypeName.get(ClassName.get(IntFunction.class), visitResultType);
+        } else if (memberType.equals(TypeName.DOUBLE)) {
+            return ParameterizedTypeName.get(ClassName.get(DoubleFunction.class), visitResultType);
```
carterkozak, line 392: 👍

conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
aioobe: Couldn't find any trace of `@Nonnull` elsewhere in fact, so I dropped it. We still have the `Preconditions.checkNotNull`.

aioobe:
@ferozco I think this is in good shape now. What's the next step? Should I add a changelog entry? Just follow the existing pattern of putting a file in the `@unreleased` directory?

iamdanfox (approved):
This looks really cool, thanks for putting it together @aioobe

carterkozak:
One quick thing: I think we should sort visitor stages alphabetically since order can change without breaking the wire format, that way modifying the conjure definition in a non-breaking way doesn't break java consumers!

aioobe:
Visitor builder stages are now in alphabetical order. Good catch @carterkozak.

Issue #78, opened 2018-10-10
values() method for enum-types in Java


ilyanep:
This would be super valuable. Right now, I'm finding myself reaching to the embedded `Value` type, which feels kind of inelegant and I have to filter out the `UNKNOWN` value.

For example, retrieving the string value names for display in a drop-down window:
```
    public static <T extends Enum<T>> String[] valueNamesWithoutUnknown(Class<T> underlyingConjureEnumType) {
        List<String> ret = Lists.newArrayList();
        for (T val : underlyingConjureEnumType.getEnumConstants()) {
            if (!val.name().equals("UNKNOWN")) {
                ret.add(val.name());
            }
        }
        return ret.toArray(new String[0]);
    }
```

Also, writing a test that creates jobs with different types, which includes not only having to filter out `UNKNOWN` but also having to convert back into the actual type itself:
```
        List<PeeringActionType.Value> peeringTypes = Lists.newArrayList(PeeringActionType.Value.values());
        // <snip>
        for (PeeringActionType.Value type : peeringTypes) {
            if (type != PeeringActionType.MINIMAL_EXPORT.get() && type != PeeringActionType.Value.UNKNOWN) {
                QueuedPeeringJobMetadata first = createPeeringJobMetadata(i++, 14, PeeringActionType.valueOf(type.name()));
                // <snip>
            }
        }
```

carterkozak:
Fixed by #768

Issue #79, opened 2018-10-10
Enum types in Java should implement some common interface


ilyanep:
Some sort of interface a la `Enum<T>`, call it `ConjureEnum<T>`. It doesn't even necessarily need the same methods (except for a `values` method as in #78, and the `valueOf` method [perhaps with the guarantee that ConjureEnum#valueOf applied to ConjureEnum#toString always gives you back the same value])

This is useful in a few cases where I want to write some utility code that applies generally to enums, but after replacing with Conjure enums, I have to either abandon that or pass two classes (the internal enum value class as well as the outer class).

Examples:

Getting string values for printing in a drop-down menu:
```
    public static <T extends Enum<T>> String[] valueNamesWithoutUnknown(Class<T> underlyingConjureEnumType) {
        List<String> ret = Lists.newArrayList();
        for (T val : underlyingConjureEnumType.getEnumConstants()) {
            if (!val.name().equals("UNKNOWN")) {
                ret.add(val.name());
            }
        }
        return ret.toArray(new String[0]);
    }
```

We frequently have enum classes with util methods that round-trip the values through an id (e.g. for serializing to a DB), so we unit test that the round-trip is stable:
```
public class PeeringConjureEnumUtilBaseTest<T, U extends Enum<U>> extends BaseTest {
    // <snip>
    protected final void testIdRoundTrips(Function<T, Long> getIdFunc,
                                Function<Long, T> fromIdFunc,
                                Class<U> clazz,
                                Function<U, T> fromEnumFunc) {
        try (AutoCloseableSoftAssertions softAssert = new AutoCloseableSoftAssertions()) {
            for (U enumValue : clazz.getEnumConstants()) {
                if (!enumValue.name().equals("UNKNOWN")) {
                    T value = fromEnumFunc.apply(enumValue);
                    try {
                        T roundTrippedValue = fromIdFunc.apply(getIdFunc.apply(value));
                        softAssert.assertThat(roundTrippedValue)
                                .as("Round-tripping through converters should provide same value.")
                                .isEqualTo(value);

                    } catch (Throwable e) {
                        softAssert.assertThat(true)
                                .as("Failed to roundtrip value %s with exception %s",
                                        value,
                                        e.getMessage())
                                .isFalse();
                    }
                }
            }
        }
}
```

In a property/configuration class where I know the values will take on the values of an enum and I want to validate that as well as return the right value:
```
    @Override
[...]

PR #466, opened 2019-07-24
[Improvement] Add new common features for enums

ilyanep:
## Before this PR

Conjure-generated enums are missing several important features of native Java enums:
- The ability to iterate over all possible values
- A way to know that a given class is a Conjure-generated enum class and therefore to know that it will have a way to get its values and call valueOf.

## After this PR
==COMMIT_MSG==
Conjure-generated enums now all implement a common interface, `ConjureEnum<E extends ConjureEnum<E>>`, and also have a new static method, `values`, that returns an array of all possible values for the enum.
==COMMIT_MSG==

## Possible downsides?

None that I can think of. The extra code represents a slight amount of extra maintenance burden on the repo.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
@@ -75,6 +75,10 @@ public final class EnumExample {
         }
     }
 
+    public static EnumExample[] values() {
+        return new EnumExample[] {ONE, TWO, ONE_HUNDRED};
+    }
+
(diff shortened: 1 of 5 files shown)


ilyanep:
I should mention that this fixes #78 and #79

carterkozak:
Hi @ilyanep, would you mind using separate PRs for #78 and #79? This way it's easier to have constructive conversations about each issue without stepping on our own toes!

ilyanep:
Sure thing. The interface calls a method added in #78 so there will still be some dependence but I'll split it out.

PR #467, opened 2019-07-24
[Improvement] Adds values() type to all generated enum classes

### LINKED ISSUE #78: values() method for enum-types in Java

ilyanep:
Fixes #78

## Before this PR
There isn't a good existing way to loop over all valid values for a conjure-generated enum type in the same way that I can do for Java native types. This frequently results in having to reach down into the underlying native Java enum class in a way that feels like a leaky abstraction.

## After this PR
==COMMIT_MSG==
Conjure-generated enum types now have a `values()` method, which returns a list of all valid values for that type.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
@@ -75,6 +75,10 @@ public final class EnumExample {
         }
     }
 
+    public static EnumExample[] values() {
+        return new EnumExample[] {ONE, TWO, ONE_HUNDRED};
+    }
+
     public <T> T accept(Visitor<T> visitor) {
         switch (value) {
             case ONE:
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
@@ -3,6 +3,9 @@ package com.palantir.product;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.Preconditions;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Locale;
 import javax.annotation.Generated;
 
@@ -27,6 +30,9 @@ public final class EnumExample {
 
     /** Value of 100. */
     public static final EnumExample ONE_HUNDRED = new EnumExample(Value.ONE_HUNDRED, "ONE_HUNDRED");
+
+    public static final List<EnumExample> VALUES =
+            Collections.unmodifiableList(Arrays.asList(ONE, TWO, ONE_HUNDRED));
 
     private final Value value;
 
@@ -75,8 +81,8 @@ public final class EnumExample {
         }
     }
 
-    public static EnumExample[] values() {
-        return new EnumExample[] {ONE, TWO, ONE_HUNDRED};
+    public static List<EnumExample> values() {
+        return VALUES;
     }
 
     public <T> T accept(Visitor<T> visitor) {
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/SimpleEnum.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/SimpleEnum.java
@@ -63,6 +63,10 @@ public final class SimpleEnum {
         }
     }
 
+    public static SimpleEnum[] values() {
+        return new SimpleEnum[] {VALUE};
+    }
+
     public <T> T accept(Visitor<T> visitor) {
         switch (value) {
             case VALUE:
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/SimpleEnum.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/SimpleEnum.java
@@ -3,6 +3,9 @@ package com.palantir.product;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.Preconditions;
+import java.util.Arrays;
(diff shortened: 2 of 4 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
```diff
@@ -28,6 +31,9 @@
     /** Value of 100. */
     public static final EnumExample ONE_HUNDRED = new EnumExample(Value.ONE_HUNDRED, "ONE_HUNDRED");
 
+    public static final List<EnumExample> VALUES =
```
carterkozak, line 34: Can we make this private?

conjure-java-core/src/integrationInput/java/com/palantir/product/SimpleEnum.java
```diff
@@ -21,6 +24,9 @@
 public final class SimpleEnum {
     public static final SimpleEnum VALUE = new SimpleEnum(Value.VALUE, "VALUE");
 
+    public static final List<SimpleEnum> VALUES =
```
carterkozak, line 27: do we need to defend against an enum value named `VALUES`?

conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
```diff
@@ -28,6 +31,9 @@
...
+            Collections.unmodifiableList(Arrays.asList(ONE, TWO, ONE_HUNDRED));
```
markelliot, line 35: this doesn't actually encompass all the possible values as it's possible for `EnumExample`, for instance, to contain an infinite number of other potential, unknown values.

markelliot:
It seems problematic that the values method can't possibly return all the possible values of the wrapper class because the wrapper class is capable of holding unknown values.

Re: leaky abstraction: we chose to make the underlying enum class available so that we would not need to deal with the uncanny valley of the wrapper behaving exactly like Java enums, but this PR seems to explicitly take us down this path.

conjure-java-core/src/integrationInput/java/com/palantir/product/SimpleEnum.java
ilyanep: Oof good point

conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
ilyanep: For most use cases that I can think of, I don't particularly care about any potential unknown values, since there won't really be anything I can do with them anyway and I'll be using the method on the server side, where I think we generally expect to have the newer version of a given type.

rubenfiszel: What about we rename this from `values` to `knownValues` so that we lower the expectations?

uschi2000:
@ilyanep can you please provide a use case / rationale for this change? I would prefer not opening this door if possible.

ilyanep:
@uschi2000 I have a couple of examples on the linked ticket #78 

Moreover, there are cases where I might want to generically loop over all the known values of a Conjure enum. For example, I have tests where I've divided the values of an enum into certain classes/sets and I want to write a unit test to make sure that the classes form a partition of the known values of the enum. Instead of rewriting that logic for every single Conjure-generated enum, I'd like to be able to write a generic method for it. 

In general, a lot of these tests would be much easier or unnecessary to write if I were using a native Java enum that I could control (for example, I can enforce a partition by having the class be a non-null member variable on the enum, and I can write test code that takes an `<E extends Enum<E>>`). I'm trying to regain at least a small amount of that usability between this PR and my other one.

I suspect @carterkozak has some other use cases in mind since he thumbsed up my ticket.

My other option is to constantly be translating between an internal enum that I control and the conjure enum at the API layer. It seems kind of clunky and wasteful to do that. I'd basically be encoding a lot of redundant data and also encoding related data in vastly different parts of my codebase.

The point of an enum is that the values are supposed to be known at compile-time. I understand that we want to force users to think about what to do in the `UNKNOWN` case. I also see from Mark's comment that we're okay with people reaching into the underlying `Value` enum, though Conjure doesn't currently allow me to do that in a generic fashion. 

I'm currently finding myself writing really fragile utility methods to get around this that have to reflectively find a method called `get` and a method called `valueOf` and a subclass called `Value` to get at the functionality I would like to have here. Some of that may be unavoidable, but by living in my codebase rather than here, I have no guarantee those won't ever break.

uschi2000:
Thanks for the examples, @ilyanep . My preference is to keep the generated union API surface as small as possible (i.e., the status quo) and in return accept the extra complexity in tests. I understand that this makes the testing code base a bit more cumbersome, but I believe this is the right trade.

PR #468, opened 2019-07-24
[Improvement] Add several interfaces that allow for compile-time type-checking for various Conjure enum operations

### LINKED ISSUE #79: Enum types in Java should implement some common interface

ilyanep:
Fixes the issues described in #78 and #79 in a slightly different way than originally envisioned.

## Before this PR
There is currently no way to tell that a type is a Conjure-generated enum (and thus that it will have methods like `values` or `valueOf`). This means that if I want to write a generic utility method, I tend to have to pass both the conjure enum class and the underlying java enum class, but have no type safety in doing so. 

For example, instead of:
```
    @Override
    public <T extends Enum<T>, U> U getEnum(PeerPropertyLookup remoteSystemId,
                                            NPPeerProperty prop,
                                            Class<T> enumClass,
                                            Class<U> returnClass) {
        Validate.isTrue(prop.getConjureEnumReturnType().equals(returnClass));
        // This validates that it's a valid entry in the enum.
        T enumReturn = getEnum(remoteSystemId, prop, enumClass);
        // This is a tricksy way to call #valueOf that doesn't rely on reflection but does rely on the fairly safe
        // assumption that Conjure will continue to make their enums Jackson-serializable in the future.
        try {
            return new ObjectMapper().readValue("\"" + enumReturn.name() + "\"", returnClass);
        } catch (IOException e) {
            throw new RuntimeException("Failed to parse enum for peer property " + prop + " into class " + returnClass.getName(),
                    e);
        }
    }
```

I can write:

```
    @Override
    public <E extends Enum<E> & WrappedConjureEnum<E, T>, T extends ConjureEnum<T, E>> T getEnum(PeerPropertyLookup remoteSystemId,
                                            NPPeerProperty prop,
                                            Class<T> conjureClass,
                                            Class<E> enumClass) {
        Validate.isTrue(prop.getReturnType().equals(conjureClass));
        E enumValue = Enum.valueOf(enumClass, getString(remoteSystemId, prop));
        return enumValue.toWrapper();
    }
```

## After this PR
==COMMIT_MSG==
The underlying value enum of all Conjure-generated enums now implement a common interface, `WrappedConjureEnum`, and Conjure-generated enums themselves now implement a common interface `ConjureEnum`, allowing for compile-time type-checking and generic code that deals with generated enums.
==COMMIT_MSG==

## Possible downsides?

The new interfaces are a little bit more complex than what most developers are used to working with when it comes to interfaces, however this is mitigated by the fact that users of the enums do not need to use the interfaces.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
@@ -21,7 +21,7 @@ import javax.annotation.Generated;
  * compile time.
  */
 @Generated("com.palantir.conjure.java.types.EnumGenerator")
-public final class EnumExample implements ConjureEnum<EnumExample> {
+public final class EnumExample implements ConjureEnum<EnumExample.Value> {
     public static final EnumExample ONE = new EnumExample(Value.ONE, "ONE");
 
     public static final EnumExample TWO = new EnumExample(Value.TWO, "TWO");
@@ -38,6 +38,7 @@ public final class EnumExample implements ConjureEnum<EnumExample> {
         this.string = string;
     }
 
+    @Override
     public Value get() {
         return this.value;
     }
@@ -74,10 +75,6 @@ public final class EnumExample implements ConjureEnum<EnumExample> {
             default:
                 return new EnumExample(Value.UNKNOWN, upperCasedValue);
         }
-    }
-
-    public static EnumExample[] values() {
-        return new EnumExample[] {ONE, TWO, ONE_HUNDRED};
     }
 
     public <T> T accept(Visitor<T> visitor) {
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
@@ -75,6 +75,10 @@ public final class EnumExample {
         }
     }
 
+    public static EnumExample[] values() {
+        return new EnumExample[] {ONE, TWO, ONE_HUNDRED};
+    }
+
     public <T> T accept(Visitor<T> visitor) {
         switch (value) {
             case ONE:
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
@@ -3,6 +3,7 @@ package com.palantir.product;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.conjure.java.lib.ConjureEnum;
+import com.palantir.conjure.java.lib.WrappedConjureEnum;
 import com.palantir.logsafe.Preconditions;
 import java.util.Locale;
 import javax.annotation.Generated;
@@ -21,7 +22,7 @@ import javax.annotation.Generated;
  * compile time.
  */
 @Generated("com.palantir.conjure.java.types.EnumGenerator")
-public final class EnumExample implements ConjureEnum<EnumExample.Value> {
+public final class EnumExample implements ConjureEnum<EnumExample, EnumExample.Value> {
     public static final EnumExample ONE = new EnumExample(Value.ONE, "ONE");
 
     public static final EnumExample TWO = new EnumExample(Value.TWO, "TWO");
@@ -91,7 +92,7 @@ public final class EnumExample implements ConjureEnum<EnumExample.Value> {
     }
 
     @Generated("com.palantir.conjure.java.types.EnumGenerator")
-    public enum Value {
+    public enum Value implements WrappedConjureEnum<Value, EnumExample> {
         ONE,
 
         TWO,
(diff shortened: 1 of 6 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
```diff
@@ -20,7 +21,7 @@
  * compile time.
  */
 @Generated("com.palantir.conjure.java.types.EnumGenerator")
-public final class EnumExample {
+public final class EnumExample implements ConjureEnum<EnumExample> {
```
markelliot: why is it necessary for this to satisfy the `ConjureEnum` interface?

ilyanep: This allows for the creation of generic utility code as given in the CR description. Right now, I have to write ugly, cludgy, or even fragile code to work with conjure enums generically.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/WrappedConjureEnum.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.lib;
+
+/**
+ * Represents a native Java enum that is the underlying value of a {@link ConjureEnum}.
+ */
+public interface WrappedConjureEnum<E extends Enum<E> & WrappedConjureEnum<E, T>, T extends ConjureEnum<T, E>> {
```
ilyanep, line 22: This is probably a bit overkill for most uses. If we're interested in actually getting this PR in, I think it might be simpler to do something more like `WrappedConjureEnum<E, T>` and `ConjureEnum<T, E extends WrappedConjureEnum<T>>` and then methods could have an `<E extends WrappedConjureEnum<E, T>, T extends ConjureEnum<T, E>` and only declare the `& Enum<E>` bit if they need that. Not sure if anyone has thoughts.

iamdanfox (changes requested):
I'm not sure the tradeoff here is worth it. The extra complexity and methods that will be available at production could lead to people getting confused about how to use the enum, and also increases the surface of possible ABI breaks. Given that this seems most beneficial in test code, I think sticking with the current approach (reaching inside to grab the values thing) is what I'd recommend.

PR #473, opened 2019-07-30
Undertow async result exceptions are logged with the correct traceId

carterkozak:
Also provides a tracing span showing asynchronous responses where
previously only the main action was traced.

## After this PR
==COMMIT_MSG==
Undertow async result exceptions are logged with the correct traceId
==COMMIT_MSG==


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -133,18 +134,41 @@
...
+                        serverExchange, exceptionSerializer, throwable), tracer));
             }
         }, DIRECT_EXECUTOR));
     }
+
+    private HttpHandler wrapCallback(HttpHandler action, DeferredTracer tracer) {
+        return new DeferredTracerHandler(new ConjureExceptionHandler(action, exceptionSerializer), tracer);
+    }
+
+    private static final class DeferredTracerHandler implements HttpHandler {
```
rubenfiszel: could also be shortened as:
```
private static final Function<DeferredTracer, HandlerWrapper> deferredTracerHandlerWrapper = tracer -> next -> tracer.withTrace(() -> {
                next.handleRequest(exchange);
                return null; })
```

rubenfiszel (approved):
LGTM

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -133,18 +134,41 @@
...
-                exchange.dispatch(new ConjureExceptionHandler(serverExchange -> ConjureExceptions.handle(
-                        serverExchange, exceptionSerializer, throwable), exceptionSerializer));
+                exchange.dispatch(wrapCallback(serverExchange -> ConjureExceptions.handle(
+                        serverExchange, exceptionSerializer, throwable), tracer));
             }
         }, DIRECT_EXECUTOR));
     }
+
+    private HttpHandler wrapCallback(HttpHandler action, DeferredTracer tracer) {
+        return new DeferredTracerHandler(new ConjureExceptionHandler(action, exceptionSerializer), tracer);
```
ferozco: Could we avoid the new class and return a lambda instead?
```suggestion
        return exchange -> tracer.withTrace(() -> {
            new ConjureExceptionHandler(action, exceptionSerializer).handleRequest(exchange);
            return null;
        });
```

carterkozak: Updated. I find classes easier to follow in stack traces, but I don't have much preference either way.

ferozco:
👍

ferozco:
poke

PR #480, opened 2019-08-02
conjure-undertow supports OPTIONS requests.

carterkozak:
## Before this PR
Previously they would result in 405 method not allowed responses.

## After this PR
==COMMIT_MSG==
conjure-undertow supports OPTIONS requests.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.java.undertow.runtime;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.Lists;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
@@ -24,6 +25,7 @@ import com.palantir.conjure.java.undertow.lib.Endpoint;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.tracing.undertow.TracedOperationHandler;
 import io.undertow.Handlers;
 import io.undertow.server.HttpHandler;
@@ -57,8 +59,7 @@ public final class ConjureHandler implements HttpHandler {
     }
 
     private void registerOptionsEndpoints(List<Endpoint> endpoints) {
-        ImmutableSetMultimap<String, HttpString> pathToMethod = endpoints.stream()
-                .filter(endpoint -> !Methods.OPTIONS.equals(endpoint.method()))
+        ImmutableSetMultimap<String, HttpString> pathToConfiguredMethods = endpoints.stream()
                 .collect(ImmutableSetMultimap.toImmutableSetMultimap(
                         endpoint -> normalizeTemplate(endpoint.template()),
                         Endpoint::method));
@@ -66,8 +67,10 @@ public final class ConjureHandler implements HttpHandler {
         // instead of using setInvalidMethodHandler.
         RoutingHandler optionalRoutingHandler = Handlers.routing()
                 .setFallbackHandler(ResponseCodeHandler.HANDLE_405);
-        pathToMethod.keySet().forEach(normalizedPath -> optionalRoutingHandler.add(Methods.OPTIONS, normalizedPath,
-                new WebSecurityHandler(new OptionsHandler(pathToMethod.get(normalizedPath)))));
+
+        pathToConfiguredMethods.asMap()
+                .forEach((normalizedPath, methods) -> optionalRoutingHandler.add(Methods.OPTIONS, normalizedPath,
+                        new WebSecurityHandler(new OptionsHandler(ImmutableSet.copyOf(methods)))));
         routingHandler.setInvalidMethodHandler(optionalRoutingHandler);
     }
 
@@ -89,6 +92,12 @@ public final class ConjureHandler implements HttpHandler {
 
     public static final class Builder {
 
+        private static final ImmutableSet<HttpString> ALLOWED_METHODS = ImmutableSet.of(
+                Methods.GET,
+                Methods.PUT,
+                Methods.POST,
+                Methods.DELETE);
+
         private final List<EndpointHandlerWrapper> wrappersJustBeforeBlocking = new ArrayList<>();
 
         private final List<Endpoint> endpoints = Lists.newArrayList();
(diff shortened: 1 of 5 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -51,6 +53,22 @@ private ConjureHandler(
             List<Endpoint> endpoints) {
         this.routingHandler = Handlers.routing().setFallbackHandler(fallback);
         endpoints.forEach(this::register);
+        registerOptionsEndpoints(endpoints);
+    }
+
+    private void registerOptionsEndpoints(List<Endpoint> endpoints) {
+        ImmutableSetMultimap<String, HttpString> pathToMethod = endpoints.stream()
+                .filter(endpoint -> !Methods.OPTIONS.equals(endpoint.method()))
```
markelliot: we probably ought to error if we ever find any of these?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -51,6 +53,22 @@ private ConjureHandler(
             List<Endpoint> endpoints) {
         this.routingHandler = Handlers.routing().setFallbackHandler(fallback);
         endpoints.forEach(this::register);
+        registerOptionsEndpoints(endpoints);
+    }
+
+    private void registerOptionsEndpoints(List<Endpoint> endpoints) {
+        ImmutableSetMultimap<String, HttpString> pathToMethod = endpoints.stream()
```
markelliot: nit: s/pathToMethod/pathToConfiguredMethods/`? (minimally should pluralize)

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -51,6 +53,22 @@ private ConjureHandler(
...
+                .filter(endpoint -> !Methods.OPTIONS.equals(endpoint.method()))
+                .collect(ImmutableSetMultimap.toImmutableSetMultimap(
+                        endpoint -> normalizeTemplate(endpoint.template()),
+                        Endpoint::method));
+        // Once UNDERTOW-1581 is fixed we should register OPTIONS handling directly with routingHandler
+        // instead of using setInvalidMethodHandler.
+        RoutingHandler optionalRoutingHandler = Handlers.routing()
+                .setFallbackHandler(ResponseCodeHandler.HANDLE_405);
+        pathToMethod.keySet().forEach(normalizedPath -> optionalRoutingHandler.add(Methods.OPTIONS, normalizedPath,
```
markelliot:
```suggestion
        pathToMethod.entrySet().forEach((normalizedPath, methods) -> optionalRoutingHandler.add(Methods.OPTIONS, normalizedPath,
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -51,6 +53,22 @@ private ConjureHandler(
...
+                new WebSecurityHandler(new OptionsHandler(pathToMethod.get(normalizedPath)))));
```
markelliot:
```suggestion
                new WebSecurityHandler(new OptionsHandler(methods))));
```

markelliot (commented):
few nits

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
carterkozak: Removed this filter in favor of validation on the builder to prevent any methods that aren't defined by conjure.

uschi2000:
Should we update the conjure spec to define the behavior for OPTIONS?

uschi2000:
(I'm also wondering, why exactly do we need support for OPTIONS?)

iamdanfox:
@uschi2000 dropwizard/jetty+jersey transparently support OPTIONS requests because browsers send them as part of a '[CORS preflight](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request)' request.  If your server doesn't respond to these, then you won't be able to hit these endpoints from a browser.

We discovered this when we were setting up tests where conjure-typescript hits the conjure-verification from a browser! https://github.com/palantir/conjure-verification/blob/d99bad8677a4655cdc77e99406e91a3343bc7145/verification-http-server/src/lib.rs

carterkozak:
Yep, @iamdanfox is correct, that is the rationale for this change.

> Should we update the conjure spec to define the behavior for OPTIONS?

Absolutely, I chatted with @markelliot about adding a note describing OPTIONS support but I neglected to file a ticket to follow up. Filed: https://github.com/palantir/conjure/issues/355

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -186,4 +221,8 @@ private void checkOverlappingPaths() {
             }
         }
     }
+
+    private static String normalizeTemplate(String template) {
+        return template.replaceAll("\\{.*?\\}", "{param}");
```
iamdanfox, line 226: Can we include a test to prove this works with multiple path parameters (e.g. `/foo/{owner}/{repo}/issues/{number}`? I _think_ that `.*?` is non-greedy, but I never trust myself to eyeball regexes and not miss something

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/OptionsHandler.java
```diff
@@ -0,0 +1,43 @@
...
+
+    OptionsHandler(Set<HttpString> methods) {
+        this.allowValue = Joiner.on(", ").join(methods);
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange exchange) {
+        Preconditions.checkState(Methods.OPTIONS.equals(exchange.getRequestMethod()), "Expected an OPTIONS request");
+        exchange.getResponseHeaders().put(Headers.ALLOW, allowValue);
+        exchange.setStatusCode(StatusCodes.NO_CONTENT);
```
iamdanfox, line 41: Do we want `Access-Control-Max-Age` too? https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request I assume this lets browsers cache this for a bit?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
carterkozak: 👍

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/OptionsHandler.java
carterkozak: If we want to set Access-Control headers, I think that belongs in [WebSecurityHandler](https://github.com/palantir/conjure-java/blob/4caf6a593d19597312f7895f41230aa2859589fe/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/WebSecurityHandler.java). I don't think there's anything specific to options requests for access control headers beyond standard browser use, but I'm not completely confident on that.

iamdanfox (approved):

PR #483, opened 2019-08-03
[fix] Add @Override annotations to visitor builder methods

aioobe:
## Before this PR
The recently introduced visitor builder for union type visitors does not include `@Override` annotations on generated builder methods. This sometimes causes compile errors in generated code.

## After this PR
==COMMIT_MSG==
[fix] Adds override annotations to union type visitor builder methods.
==COMMIT_MSG==

## Possible downsides?
None.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
@@ -87,24 +87,29 @@ public final class SingleUnion {
 
         private Function<String, T> unknownVisitor;
 
+        @Override
         public UnknownStageVisitorBuilder<T> foo(Function<String, T> fooVisitor) {
             Preconditions.checkNotNull(fooVisitor, "fooVisitor cannot be null");
             this.fooVisitor = fooVisitor;
             return this;
         }
 
+        @Override
         public CompletedStageVisitorBuilder<T> unknown(Function<String, T> unknownVisitor) {
             Preconditions.checkNotNull(unknownVisitor, "unknownVisitor cannot be null");
             this.unknownVisitor = unknownVisitor;
             return this;
         }
 
+        @Override
         public Visitor<T> build() {
             return new Visitor<T>() {
+                @Override
                 public T visitFoo(String value) {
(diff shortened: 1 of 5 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
```diff
@@ -87,24 +87,29 @@ public String toString() {
 
         private Function<String, T> unknownVisitor;
 
+        @Override
```
carterkozak, line 90: Mind adding a `final` modifier to the VisitorBuilder while we're here?

conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
```diff
@@ -87,24 +87,29 @@ public String toString() {
...
         public UnknownStageVisitorBuilder<T> foo(Function<String, T> fooVisitor) {
             Preconditions.checkNotNull(fooVisitor, "fooVisitor cannot be null");
             this.fooVisitor = fooVisitor;
             return this;
         }
 
+        @Override
         public CompletedStageVisitorBuilder<T> unknown(Function<String, T> unknownVisitor) {
             Preconditions.checkNotNull(unknownVisitor, "unknownVisitor cannot be null");
             this.unknownVisitor = unknownVisitor;
```
carterkozak, line 100: not necessarily for this PR, we can update this to the form:
```java
this.unknownVisitor = Preconditions.checkNotNull(unknownVisitor, "unknownVisitor cannot be null");
```

aioobe: Yes. This struck me too. Can't say I'm a fan of that pattern, but it could very well be that I'm just not used to it.

conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
aioobe: Done

carterkozak:
Thanks 👍

PR #485, opened 2019-08-04
[fix] Makes union type visitor handlers immutable in build method.

aioobe:
## Before this PR
By holding on to a visitor builder stage, a visitor handler can be updated after `build()` has been called.

## After this PR
==COMMIT_MSG==
[fix] Makes union type visitor handlers immutable in build method.
==COMMIT_MSG==

## Possible downsides?
None, unless someone thinks this is a feature (but it's not).


iamdanfox (approved):
Nice

Issue #47, opened 2018-07-18
Overloaded methods for optional query params may silently allow compilation in some unsafe cases


a-k-g:
With #21, we generate `@Deprecated` methods with the intention of avoiding compile breaks in cases where new query params are added. However, in some cases we might silently let compilation go through when it shouldn't. Specifically, if we add another optional query params with the same type as an existing one and in the api.yml declare it right before the existing one.

As an example, lets say we have the following API:

```
getTransactionsPage:
  http: POST /get-page
  args:
    query: string
    optionalEnd:
      type: optional<ResourceIdentifier>
      param-type: query
  returns: set<ResourceIdentifier>
-----
@POST
@Path("catalog/get-page")
Set<ResourceIdentifier> getTransactionsPage(
        @HeaderParam("Authorization") AuthHeader authHeader,
        @QueryParam("optionalEnd") Optional<ResourceIdentifier> optionalEnd,
        String query);

@Deprecated
default Set<ResourceIdentifier> getTransactionsPage(AuthHeader authHeader, String query) {
    return getTransactionsPage(authHeader, Optional.empty(), query);
}
```
and this is being used as follows in the client code:

```
Set<ResourceIdentifier> transactions = getTransactionsPage(authHeader, endTransactionRid, query);
```

Lets say the API was included to also have an `optionalStart` query param:
```
getTransactionsPage:
  http: POST /get-page
  args:
    query: string
    optionalStart:
      type: optional<ResourceIdentifier>
      param-type: query
    optionalEnd:
      type: optional<ResourceIdentifier>
      param-type: query
  returns: set<ResourceIdentifier>
-----
@POST
@Path("catalog/get-page")
Set<ResourceIdentifier> getTransactionsPage(
        @HeaderParam("Authorization") AuthHeader authHeader,
        @QueryParam("optionalStart") Optional<ResourceIdentifier> optionalStart,
        @QueryParam("optionalEnd") Optional<ResourceIdentifier> optionalEnd,
        String query);

@Deprecated
default Set<ResourceIdentifier> getTransactionsPage(AuthHeader authHeader, String query) {
    return getTransactionsPage(authHeader, Optional.empty(), Optional.empty(), query);
}

@Deprecated
default Set<ResourceIdentifier> getTransactionsPage(
        AuthHeader authHeader, Optional<ResourceIdentifier> optionalStart, String query) {
    return getTransactionsPage(authHeader, optionalStart, Optional.empty(), query);
}
```
[...]

PR #499, opened 2019-08-16
break: Match parameter order between jax-rs and retrofit clients

### LINKED ISSUE #339: Retrofit service parameters are not sorted
## What happened?
Retrofit service interface parameters are not sorted and do not provide backwards compatibility default methods with guarantees matching jersey.

Backwards compatibility guarantees for retrofit and jersey should match

robert3005:
## Before this PR
Retrofit generated interfaces didn't have their parameters sorted and didn't have the backcompat methods for optional arguments

## After this PR
==COMMIT_MSG==
Match jax-rs client generator behaviour in retrofit. Retrofit clients now have their arguments sorted the same way jax-rs does and have backcompat methods for optional arguments.
==COMMIT_MSG==

## Possible downsides?
The argument order of generated methods will change which will require fixing up by users

fixes #339

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/Retrofit2ServiceGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/Retrofit2ServiceGenerator.java
@@ -387,4 +387,5 @@ public final class Retrofit2ServiceGenerator implements ServiceGenerator {
         }
         throw new IllegalArgumentException("Unrecognized HTTP method: " + method);
     }
+
 }
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/Retrofit2ServiceGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/Retrofit2ServiceGenerator.java
@@ -28,32 +28,22 @@ import com.palantir.conjure.java.types.SpecializeBinaryClassNameVisitor;
 import com.palantir.conjure.java.types.TypeMapper;
 import com.palantir.conjure.java.util.Javadoc;
 import com.palantir.conjure.java.util.ParameterOrder;
-import com.palantir.conjure.java.visitor.DefaultTypeVisitor;
 import com.palantir.conjure.spec.ArgumentDefinition;
 import com.palantir.conjure.spec.ArgumentName;
 import com.palantir.conjure.spec.AuthType;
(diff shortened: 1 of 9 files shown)


dansanduleac:
@robert3005 can we feature flag this then? I'm worried that this can break existing `method(T, T)` invocations silently.

markelliot:
As an alternative to feature flagging, can generate a second interface with a new name so that consumers have a choice.

robert3005:
I don't think feature flagging this is any useful since it's not the consumers that set the flag but it's them who are affected. The reason I made this pr was to minimize differences between jax-rs and retrofit so reasonably anyone who would want to use retrofit will want to use it. @iamdanfox and I had a look at all internal usages and they were minimal compared to jax-rs so like in https://github.com/palantir/conjure-java/pull/21#pullrequestreview-132530436 it would be best to get that merged before there's more usage.

robert3005:
I do acknowledge there's an issue as described in https://github.com/palantir/conjure-java/issues/47 but like to think that the chance of it happening in practice is quite low

robert3005:
FWIW there's a chance currently to create invalid clients from POV of retrofit so unless we do this our clients will not necessarily work

conjure-java-core/src/main/java/com/palantir/conjure/java/services/Retrofit2ServiceGenerator.java
```diff
@@ -280,4 +401,69 @@ private static ClassName httpMethodToClassName(String method) {
...
+        @Override
+        public Boolean visitMap(MapType value) {
+            return true;
+        }
+
+        @Override
+        public Boolean visitDefault() {
+            return false;
+        }
+    };
```
iamdanfox: I don't really want two copies of an identical TYPE_DEFAULTABLE_PREDICATE if poss - separate private class?

iamdanfox:
Ok so given that:

a) retrofit clients for catalog do not currently work (apparently retrofit disallows query params before path params) and 
b) we looked in papaya and the vast majority of retrofit usage is by JB's atlas replacement service, which won't be affected by this change

I'm in favour of this change, thanks @robert3005!

iamdanfox (approved):
Cool thanks @robert3005, should make it much more convenient to switch between jaxrs and retrofit clients

dansanduleac:
👍

PR #505, opened 2019-08-20
Make verifier use random port

dansanduleac:
## Before this PR
develop is broken as `:conjure-java-server-verifier:test` and ` :conjure-java-undertow-runtime:test` both attempt to bind to ports 12345 12346.

## After this PR
==COMMIT_MSG==
Fix multiple test suites that run in parallel binding to ports 12345 and 12346.

conjure-java-server-verifier now instructs each of its dropwizard and undertow "servers-under-test" to bind to a random port.
==COMMIT_MSG==


ferozco:
👍

carterkozak:
Thanks Dan! 👍

PR #528, opened 2019-09-02
Take advantage of the fix for UNDERTOW-1581 for options handling

carterkozak:
==COMMIT_MSG==
Take advantage of the fix for UNDERTOW-1581 for cleaner options handling
==COMMIT_MSG==


ferozco:
👍

PR #538, opened 2019-09-09
Invariant typecasting on covariant optionals

### LINKED ISSUE #536: Generated java for optional<set<...>> types can fail to compile
Suppose a conjure object containing a field that is an optional set type: 

```
OptionalFoo:
        fields:
          item: optional<set<StringAliasExample>>
```
The generated objects java will contain a builder entry accepting an `Optional<? extends Set<StringAliasExample>>` type:

```
@JsonSetter(value = "otherItem", nulls = Nulls.SKIP)
        public Builder item(Optional<? extends Set<StringAliasExample>> item) {
            this.item =
                    (Optional<Set<StringAliasExample>>)
                            Preconditions.checkNotNull(item, "item cannot be null");
            return this;
        }
```

However, this will fail to compile on at least some java versions (ex: openjdk 11.0.2), because the `Optional<? extends Set<StringAliasExample>>` covariant type returned by the Preconditions check is not compatible with the invariant inner variable type (`Optional<Set<StringAliasExample>>`) it is being typecasted to. Example compilation error:

```
Preconditions.checkNotNull(item, "item cannot be null");
                                                      ^
  required: Optional<Set<StringAliasExample>>
  found:    Optional<CAP#1>
  where CAP#1 is a fresh type-variable:
    CAP#1 extends Set<StringAliasExample> from capture of ? extends Set<StringAliasExample>
```

verified on conjure-java 4.0.0

xfalco:
## Before this PR
Fixes https://github.com/palantir/conjure-java/issues/536

For optional fields, the generated builder objects contain setters accepting covariant inputs. A typecast is necessary to assign to the inner (invariant) variables, but this typecast generates a warning. 
## After this PR
==COMMIT_MSG==
No warning-generating typecast is present in the setter methods for optional fields.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
@@ -60,6 +60,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.lang.model.element.Modifier;
 import org.apache.commons.lang3.StringUtils;
@@ -336,16 +337,14 @@ public final class BeanBuilderGenerator {
                     .build();
         } else if (type.accept(TypeVisitor.IS_OPTIONAL)) {
             OptionalType optionalType = type.accept(TypeVisitor.OPTIONAL);
-            String parameterName = enriched.poetSpec().name;
             CodeBlock nullCheckedValue = Expressions.requireNonNull(
                     spec.name, enriched.fieldName().get() + " cannot be null");
 
             if (isWidenableContainedType(optionalType.getItemType())) {
-                // covariant optionals need to be narrowed to invariant type before assignment
-                Type innerType = optionalType.getItemType();
+                // we need to capture covariant type before assignment to invariant inner variable
                 return CodeBlock.builder()
-                        .addStatement("this.$1N = $2L.map(_$3L -> ($4T) _$3L)",
-                                spec.name, nullCheckedValue, parameterName, typeMapper.getClassName(innerType).box())
+                        .addStatement("this.$1N = $2L.map($3T.identity())",
+                                spec.name, nullCheckedValue, Function.class)
                         .build();
             } else {
                 return CodeBlocks.statement("this.$1L = $2L", spec.name, nullCheckedValue);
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/conjure-java](https://github.com/palantir/conjure-java), @xfalco! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -336,15 +336,16 @@ private CodeBlock typeAwareAssignment(EnrichedField enriched, Type type, boolean
...
                     spec.name, enriched.fieldName().get() + " cannot be null");
 
             if (isWidenableContainedType(optionalType.getItemType())) {
                 // covariant optionals need to be narrowed to invariant type before assignment
                 Type innerType = optionalType.getItemType();
                 return CodeBlock.builder()
-                        .addStatement("this.$1N = ($3T<$4T>) $2L",
-                                spec.name, nullCheckedValue, Optional.class, typeMapper.getClassName(innerType).box())
+                        .addStatement("this.$1N = $2L.map(_$3L -> ($4T) _$3L)",
+                                spec.name, nullCheckedValue, parameterName, typeMapper.getClassName(innerType).box())
```
markelliot: I think i'd use `_value` as a fixed name rather than `parameterName`.

xfalco: used `Function.identity()` generic to instead capture the covariant type (thanks for the offline recommendation)

ferozco:
👍

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -340,11 +341,10 @@ private CodeBlock typeAwareAssignment(EnrichedField enriched, Type type, boolean
                     spec.name, enriched.fieldName().get() + " cannot be null");
 
             if (isWidenableContainedType(optionalType.getItemType())) {
-                // covariant optionals need to be narrowed to invariant type before assignment
-                Type innerType = optionalType.getItemType();
+                // we need to capture covariant type before assignment to invariant inner variable
                 return CodeBlock.builder()
-                        .addStatement("this.$1N = ($3T<$4T>) $2L",
-                                spec.name, nullCheckedValue, Optional.class, typeMapper.getClassName(innerType).box())
+                        .addStatement("this.$1N = $2L.map($3T.identity())",
```
markelliot, line 347: could you also leave a comment here about why we want a `.map(Function.identity())`?

xfalco: Yeah - did you see my update to the comment above (https://github.com/palantir/conjure-java/pull/538/files#diff-3c6ac0d96f3bf3b56cf2702e890c6a96R344) ? Think that captures the reasoning, but happy to move it down to line 346

PR #555, opened 2019-09-13
Simplify conjure object toString implementation

carterkozak:
Avoid using unnecessary stringbuilders, in jdk8 javac generates
this code for us. Using a compiler target beyond jdk8 we can
take advantage of [jep 280](https://openjdk.java.net/jeps/280).

## After this PR
==COMMIT_MSG==
Simplify conjure object toString implementation
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
@@ -28,6 +28,7 @@ import com.squareup.javapoet.TypeSpec;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collector;
 import javax.lang.model.element.Modifier;
@@ -88,19 +89,15 @@ public final class MethodSpecs {
                 .collect(joining(CodeBlock.of(", "))));
     }
 
-    public static MethodSpec createToString(String thisClassName, Collection<FieldName> fieldNames) {
+    public static MethodSpec createToString(String thisClassName, List<FieldName> fieldNames) {
         CodeBlock returnStatement = fieldNames.isEmpty()
                 ? CodeBlock.builder().addStatement("return $S", thisClassName + "{}").build()
                 : CodeBlock.builder()
-                        .add("return new $T($S).append('{')\n", StringBuilder.class, thisClassName)
+                        .add("return $S\n", thisClassName + '{' + fieldNames.get(0).get() + ": ")
                         .indent()
                         .indent()
-                        .add(CodeBlocks.of(fieldNames.stream()
-                                .map(MethodSpecs::createAppendStatement)
-                                .collect(joining(CodeBlock.of(".append(\", \")")))))
-                        .unindent()
-                        .add(".append('}')\n")
-                        .addStatement(".toString()")
+                        .add(toStringConcatenation(fieldNames))
+                        .addStatement(" + '}'")
                         .unindent()
                         .build();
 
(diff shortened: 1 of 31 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
```diff
@@ -104,19 +105,15 @@ private static CodeBlock getHashInput(Collection<FieldSpec> fields) {
                 .collect(joining(CodeBlock.of(", "))));
     }
 
-    public static MethodSpec createToString(String thisClassName, Collection<FieldName> fieldNames) {
+    public static MethodSpec createToString(String thisClassName, List<FieldName> fieldNames) {
         CodeBlock returnStatement = fieldNames.isEmpty()
                 ? CodeBlock.builder().addStatement("return $S", thisClassName + "{}").build()
                 : CodeBlock.builder()
```
markelliot: it would be nicer for `toStringConcatenation` to return a CodeBlock than to require maintaining across two methods...and easier to write a test for it.

markelliot:
since the described JEP is to translate the bytecode generation to the thing we were producing anyway, is this change actually that beneficial?

carterkozak:
Before jep280 java generated bytecode matching what we generated. Newer versions implementing jep280 generate simpler bytecode which can be optimized beyond stringbuilder. I can post example bytecode when I’m near a keyboard (likely won’t make it today)

carterkozak:
Using IntegerExample for simplicity.

### Methodology

Clean and rebuild with an updated target version, and read bytecode using:
```
javap -c ./conjure-java-core/build/classes/java/integrationInput/com/palantir/product/IntegerExample.class
```

## Before this change (develop)

```
  public java.lang.String toString();
    Code:
       0: new           #7                  // class java/lang/StringBuilder
       3: dup
       4: ldc           #8                  // String IntegerExample
       6: invokespecial #9                  // Method java/lang/StringBuilder."<init>":(Ljava/lang/String;)V
       9: bipush        123
      11: invokevirtual #10                 // Method java/lang/StringBuilder.append:(C)Ljava/lang/StringBuilder;
      14: ldc           #11                 // String integer
      16: invokevirtual #12                 // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      19: ldc           #13                 // String :
      21: invokevirtual #12                 // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      24: aload_0
      25: getfield      #3                  // Field integer:I
      28: invokevirtual #14                 // Method java/lang/StringBuilder.append:(I)Ljava/lang/StringBuilder;
      31: bipush        125
      33: invokevirtual #10                 // Method java/lang/StringBuilder.append:(C)Ljava/lang/StringBuilder;
      36: invokevirtual #15                 // Method java/lang/StringBuilder.toString:()Ljava/lang/String;
      39: areturn
```

## With this change, targeting java 8

Here we see the old (pre jep280) compiler optimization resulting in the code we used to generate. Note this is a bit more compact than the old implementation due to this PR combining delimiter and field name strings.

```
  public java.lang.String toString();
    Code:
       0: new           #7                  // class java/lang/StringBuilder
       3: dup
       4: invokespecial #8                  // Method java/lang/StringBuilder."<init>":()V
       7: ldc           #9                  // String IntegerExample{integer:
       9: invokevirtual #10                 // Method java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
      12: aload_0
      13: getfield      #3                  // Field integer:I
      16: invokevirtual #11                 // Method java/lang/StringBuilder.append:(I)Ljava/lang/StringBuilder;
      19: bipush        125
      21: invokevirtual #12                 // Method java/lang/StringBuilder.append:(C)Ljava/lang/StringBuilder;
      24: invokevirtual #13                 // Method java/lang/StringBuilder.toString:()Ljava/lang/String;
      27: areturn
```

## With this change, targeting java 11

This is much simpler, we're giving the runtime a great deal more freedom to presize the resulting string rather than being prescriptive about the implementation details.

```
  public java.lang.String toString();
    Code:
       0: aload_0
       1: getfield      #2                  // Field integer:I
       4: invokedynamic #6,  0              // InvokeDynamic #0:makeConcatWithConstants:(I)Ljava/lang/String;
       9: areturn
```

markelliot:
Thanks. Should there be a corresponding errorprone and refaster rule?

carterkozak:
> Should there be a corresponding errorprone and refaster rule?

That's a good idea. Filed https://github.com/palantir/gradle-baseline/issues/831

ferozco (approved):

carterkozak:
@markelliot automated fix is here, if you're interested: https://github.com/palantir/gradle-baseline/pull/832

PR #589, opened 2019-10-14
Errors are logged from the ConjureExceptions class

carterkozak:
This is whitelisted for safe-logging, and includes trace/user
metadata where the server exception handler logs from a different scope.
This allows us to terminate partial responses when data has already
been sent.

## After this PR
==COMMIT_MSG==
Errors are logged from the ConjureExceptions class to provide more helpful metadata for debugging, and making the safe parts of the stack trace available in the log pipeline.
==COMMIT_MSG==

## Possible downsides?
none


ferozco (approved):

carterkozak:
bulldozer?

PR #593, opened 2019-10-14
Add endpoint that consumes set of aliases

JacekLach:
This breaks undertow generator because https://github.com/palantir/conjure-java/blob/4.4.0/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java#L739-L749 doesn't consider collections of aliases

## After this PR
==COMMIT_MSG==
Fix undertow generator for endpoints that refer to collections of aliased types.
==COMMIT_MSG==


JacekLach:
Actually this didn't repro eh

PR #599, opened 2019-10-16
Try palantir-java-format

iamdanfox:
We introduced a way to use the shiny new palantir-java-format here: https://github.com/palantir/gradle-baseline/pull/936. Just dogfooding

It seems like the formatter is occasionally still fighting checkstyle :/


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -366,26 +348,21 @@ private boolean isByteBuffer(Type type) {
         Type type = enriched.conjureDef().getType();
 
         if (type.accept(TypeVisitor.IS_LIST)) {
-            return ImmutableList.of(
-                    createCollectionSetter("addAll", enriched),
-                    createItemSetter(enriched, type.accept(TypeVisitor.LIST).getItemType()));
```
iamdanfox, line 371: @dansanduleac I think the number of parens in `createItemSetter(enriched, type.accept(TypeVisitor.LIST).getItemType())` makes it a bit subtle to read when it's re-flowed in the new 'dense' style.  Thoughts on having a heuristic that would value having opening and closing pairs of parens on the same line?

iamdanfox:
Closing out as we should use the official excavator one.

PR #615, opened 2019-10-29
Propagate 401s

pkoenig10:
See https://github.com/palantir/conjure-java-runtime/pull/1295.

## Before this PR
401 responses are not propagated.

## After this PR
401 responses are propagated.


carterkozak:
lgtm, thanks! 👍

PR #618, opened 2019-10-30
Produce a 4xx response when the request is not valid json

carterkozak:
## Before this PR
Most types of invalid requests result in 422 responses, but some result in a 500 response.

## After this PR
==COMMIT_MSG==
Produce a 4xx response when the request is not valid json
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-618.v2.yml
+++ b/changelog/@unreleased/pr-618.v2.yml
@@ -1,5 +1,5 @@
-type: improvement
-improvement:
+type: fix
+fix:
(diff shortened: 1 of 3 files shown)


iamdanfox (approved):
Sending back 422 definitely sounds like an improvement - my only question would be whether it could be confusing for operators if you can no longer distinguish between a string that just isn't valid JSON and some valid JSON that isn't the expected shape?

carterkozak:
Good point, 400 is more correct here. Update incoming.

PR #641, opened 2019-11-13
Break: Always generate ListenableFuture variant of retrofit interfaces

robert3005:
## Before this PR
There's 3 different variants of retrofit interfaces. Users don't know which one to use and likely end up choosing one that doesn't align with long term vision we have for conjure-java. 

## After this PR
==COMMIT_MSG==
Retrofit generated interfaces always return ListenableFuture and never Call nor CompletableFuture.
==COMMIT_MSG==

## Possible downsides?
There might be users relying on Call or CompletableFuture version of retrofit interfaces and this will force them to switch. Default variant returning calls is most likely to cause breakages since it allows you to inspect headers of response.


dansanduleac:
We should fix the binary generation as well to return `ListenableFuture<Optional<ResponseBody>>` before merging as otherwise we'd need to make two breaks.

Also, in order to fix existing users of Call, we probably need to write an errorprone check to migrate them from `call.execute().body()` to `Futures.getUnchecked(future)`

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -90,16 +90,6 @@ public void run() {
                 description = "Generate retrofit interfaces for streaming/async clients")
         private boolean generateRetrofit;
 
-        @CommandLine.Option(names = "--retrofitCompletableFutures",
-                defaultValue = "false",
-                description = "Generate retrofit services which return Java8 CompletableFuture instead of OkHttp Call")
-        private boolean retrofitCompletableFutures;
-
-        @CommandLine.Option(names = "--retrofitListenableFutures",
```
dansanduleac, line 98: keep this otherwise we break 200 consumers

carterkozak: I think we allow unknown arguments for this reason, removal shouldn't break anything

robert3005:
https://github.com/palantir/conjure-java/issues/182 suggests that we need some conjure-java-runtime changes to do the thing that @dansanduleac wants

robert3005:
Ok, the weird thing here is that retrofit doesn't let you override behaviour for builtin types and handles ResponseBody + Streaming itself. We just have to teach conjure to not delegate to jackson on responsebody return type

uschi2000:
Instead of all this, why not invest in dialogue?

robert3005:
It's a tactical change that mostly avoids confusion. I can't spend 8 hours on dialogue and ship it to production everywhere unfortunately

uschi2000:
I know ... it's more of a strategic question than a tactical one.

robert3005:
Sorry for being terse. I wanted to make this change to make some of the bad code I had seen in our repos to be impossible to write again. The thing that is now clear to me is that most of weirdness in conjure comes from trying to fit into existing frameworks instead of owning more of the rpc layer. I would love to have one version of the generator as well as simpler runtime. Overall I believe that something like dialogue is the right way. More concretely this is probably my last fix to the generators and runtime that doesn't get us closer to dialogue because existing framework is too hard to understand and change.

dansanduleac:
👍

PR #643, opened 2019-11-14
Break: Correctly generate optional<binary> fields as Optional<ResponseBody> return type in retrofit client

robert3005:
## Before this PR
`optional<binary>` conjure field ends up as ResponseBody type in the generated interface

## After this PR
==COMMIT_MSG==
Generate `optional<binary>` field as `Optional<ResponseBody>` in retrofit interfaces. Users don't have to worry about getting nullable ResponseBody value type.
==COMMIT_MSG==

## Possible downsides?
This requires runtime upgrade to support the decoding. Otherwise it will just throw at runtime with a cryptic jackson exception. Requires https://github.com/palantir/conjure-java-runtime/pull/1321


robert3005:
@dansanduleac this is the change you requested in https://github.com/palantir/conjure-java-runtime/pull/1321

conjure-java-core/src/test/resources/test/api/TestServiceRetrofit.java.retrofit
```diff
@@ -73,8 +73,7 @@ public interface TestServiceRetrofit {
         "hr-path-template: /catalog/datasets/{datasetRid}/raw-maybe",
         "Accept: application/octet-stream"
     })
-    @Streaming
```
robert3005, line 76: we should keep this annotation

robert3005: Even though it's only the builtin that respects it but avoids surprises

dansanduleac:
👍 
I've checked the remaining stragglers that haven't yet picked up https://github.com/palantir/conjure-java-runtime/pull/1321 and nobody else is using `optional<binary>`.

PR #644, opened 2019-11-15
add Strict Objects flag

ferozco:
## Before this PR
We generated bean builders with `@JsonIgnoreProperties(ignoreUnknown = true)` which caused jackson to always leniently deserialize beans, even when using the server object mapper

## After this PR
==COMMIT_MSG==
Add `StrictObjects` feature flag to generate objects that will fail to deserialize unknown properties. Users will still be able to ignore unknown properties by configuring their object mapper to do so
==COMMIT_MSG==

## Possible downsides?
N/A. This moves us towards the world where servers are actually strict


carterkozak (approved):
Thanks @ferozco

PR #656, opened 2019-11-22
Missing credentials result in 401 UNAUTHORIZED responses from conjure-undertow

carterkozak:
Previously we responded with a 400 bad request status, which isn't quite accurate.
Added custom `Conjure` namespaced ErrorTypes for better observability,
updated existing FrameworkExceptions to do the same.

## After this PR
==COMMIT_MSG==
Missing credentials result in 401 UNAUTHORIZED responses from conjure-undertow
Previously we responded with a 400 bad request status, which isn't as accurate as we would like.
==COMMIT_MSG==


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAuthorizationExtractor.java
```diff
@@ -38,6 +40,16 @@
...
 
+    private static final String USER_ID_KEY = "userId";
+    private static final String SESSION_ID_KEY = "sessionId";
+    private static final String TOKEN_ID_KEY = "tokenId";
+    private static Consumer<String> sessionIdSetter = sessionId -> MDC.put(SESSION_ID_KEY, sessionId);
+    private static Consumer<String> tokenIdSetter = tokenId -> MDC.put(TOKEN_ID_KEY, tokenId);
+    private static final ErrorType MISSING_CREDENTIAL_ERROR_TYPE = ErrorType.create(
+            ErrorType.Code.UNAUTHORIZED, "Conjure:MissingCredential");
+    private static final ErrorType MALFORMED_CREDENTIAL_ERROR_TYPE = ErrorType.create(
+            ErrorType.Code.UNAUTHORIZED, "Conjure:MalformedCredential");
```
dansanduleac: This might be a minor nit but I never heard this word being used in the singular in the context of authorization.
Thoughts on calling it "credentials"?

carterkozak: Good call. We could also use `Token` for conformity with https://github.com/palantir/auth-tokens

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AuthTest.java
```diff
@@ -55,7 +64,35 @@ public void testAuthHeaderMultipleValues() {
...
+        exchange.getRequestCookies().put(cookieName, new CookieImpl(cookieName, "token"));
+        assertThat(CONTEXT.auth().cookie(exchange, cookieName)).isEqualTo(expected);
+    }
+
+    @Test
+    public void testAuthCookieNotPresent() {
+        HttpServerExchange exchange = HttpServerExchanges.createStub();
+        assertThatThrownBy(() -> CONTEXT.auth().cookie(exchange, "any"))
+                .isInstanceOf(ServiceException.class)
+                .hasMessage("ServiceException: UNAUTHORIZED (Conjure:MissingCredential)");
```
dansanduleac: we should probably use the assertj matchers for these guys rather than assert things on strings.
https://github.com/palantir/conjure-java-runtime-api/blob/develop/test-utils/src/main/java/com/palantir/conjure/java/api/testing/ServiceExceptionAssert.java

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AuthTest.java
```diff
@@ -55,7 +64,35 @@ public void testAuthHeaderMultipleValues() {
         exchange.getRequestHeaders().add(Headers.AUTHORIZATION, "Bearer foo");
         exchange.getRequestHeaders().add(Headers.AUTHORIZATION, "Bearer bar");
         assertThatThrownBy(() -> CONTEXT.auth().header(exchange))
-                .isInstanceOf(SafeIllegalArgumentException.class)
-                .hasMessage("One Authorization header value is required");
+                .isInstanceOf(ServiceException.class)
+                .hasMessageContaining("ServiceException: UNAUTHORIZED (Conjure:MalformedCredential)");
+    }
+
+    //delete
```
dansanduleac: delete?

carterkozak: doh, I copy/pasted the tests to rename and add coverage for cookie auth. Marked was to differentiate methods with duplicate names while I renamed them. Update incoming.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/FrameworkException.java
```diff
@@ -25,23 +27,31 @@
...
         this.arguments = ImmutableList.copyOf(args);
         this.statusCode = statusCode;
+        this.errorType = errorType;
     }
 
-    static FrameworkException unprocessableEntity(String message, Throwable cause, Arg<?>... args) {
-        return new FrameworkException(message, StatusCodes.UNPROCESSABLE_ENTITY, cause, args);
+    static FrameworkException unprocessableEntity(
+            @CompileTimeConstant String message, Throwable cause, Arg<?>... args) {
+        return new FrameworkException(message, UNPROCESSABLE_ENTITY, StatusCodes.UNPROCESSABLE_ENTITY, cause, args);
```
dansanduleac, line 50: we should get the status code out of the `ErrorType.Code` directly rather than potentially creating a mismatch

carterkozak: FrameworkException is an internal detail of conjure-undertow to support failures which do not have a status code in ErrorType.Code.

dansanduleac: Never mind, I guess you can't really do that since that enum doesn't define a code for 412 nor 415 🙄

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/FrameworkException.java
```diff
@@ -58,6 +68,10 @@ int getStatusCode() {
         return statusCode;
     }
 
+    ErrorType getErrorType() {
```
uschi2000, line 71: `@VisibleForTesting`?

uschi2000: ah, nevermind this whole class has package visibility

dansanduleac:
:+1:

PR #674, opened 2019-12-10
Add metric for internal server error not caused by remote exceptions

parker89:
## Before this PR

Conjure-java equivalent of: https://github.com/palantir/conjure-java-runtime/pull/1336

## After this PR
==COMMIT_MSG==
Add `server.response.internalerror.all` metric for internal server errors not caused by remote exceptions
==COMMIT_MSG==


iamdanfox:
I'm a little less comfortable with this one - in WC we've managed to do all of the instrumentation in isolated layers, without having to plumb stuff straight into here. Additionally, since we have complete control here, I think it would be preferable not to force people to use the singleton tagged metric registry...

carterkozak:
On my queue to review this

parker89:
Reopening this.

@iamdanfox do you have a suggestion for how I might do this without using the shared registry?

What is the issue with using the shared registry anyway?

PR #676, opened 2019-12-11
Add NonNullCollection feature flag

ferozco:
## Before this PR
Conjure generated objects allowed nulls to appear inside of any collection, violating the wire spec.

## After this PR
==COMMIT_MSG==
Add `NonNullCollections` feature flag to generate objects that will fail to deserialize collections with null values.
==COMMIT_MSG==

We've expanded the scope of this change to also fix how we handled aliases of optionals

We'll also want to update the conjure-verifier to include tests to enforce that there are no nulls in collections

## Possible downsides?
Adopting this feature flag can be dangerous for some users depending on how they have used Conjure generated objects. In particular, users should be careful if they use Conjure objects in there storage layer since adopting the feature flag will make them unable to load data if any nulls had inadvertently been made it into the DB.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveOptionalsExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveOptionalsExample.java
@@ -213,7 +213,7 @@ public final class PrimitiveOptionalsExample {
             return this;
         }
 
-        @JsonSetter(value = "num", nulls = Nulls.SKIP)
+        @JsonSetter(value = "num", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
         public Builder num(OptionalDouble num) {
             this.num = Preconditions.checkNotNull(num, "num cannot be null");
             return this;
@@ -224,7 +224,7 @@ public final class PrimitiveOptionalsExample {
             return this;
         }
 
-        @JsonSetter(value = "bool", nulls = Nulls.SKIP)
+        @JsonSetter(value = "bool", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
         public Builder bool(Optional<Boolean> bool) {
             this.bool = Preconditions.checkNotNull(bool, "bool cannot be null");
             return this;
@@ -235,7 +235,7 @@ public final class PrimitiveOptionalsExample {
             return this;
         }
 
-        @JsonSetter(value = "integer", nulls = Nulls.SKIP)
+        @JsonSetter(value = "integer", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
         public Builder integer(OptionalInt integer) {
             this.integer = Preconditions.checkNotNull(integer, "integer cannot be null");
             return this;
@@ -246,7 +246,7 @@ public final class PrimitiveOptionalsExample {
             return this;
         }
 
-        @JsonSetter(value = "safelong", nulls = Nulls.SKIP)
+        @JsonSetter(value = "safelong", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
         public Builder safelong(Optional<SafeLong> safelong) {
             this.safelong = Preconditions.checkNotNull(safelong, "safelong cannot be null");
             return this;
@@ -258,7 +258,7 @@ public final class PrimitiveOptionalsExample {
             return this;
         }
 
-        @JsonSetter(value = "rid", nulls = Nulls.SKIP)
+        @JsonSetter(value = "rid", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
         public Builder rid(Optional<ResourceIdentifier> rid) {
             this.rid = Preconditions.checkNotNull(rid, "rid cannot be null");
             return this;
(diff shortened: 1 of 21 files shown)


carterkozak:
Mind adding a test for beans with a field of type `list<optional<?>>` for input `[null]`?

xfalco:
How does this behave for nested collections (ex: a `list<list<string>>` member) ? Is the non-nullity enforced only top-level, or for all members ?

ferozco:
@xfalco I've added a test case demonstrating the behaviour on nested lists

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -117,6 +117,11 @@ public void run() {
                 description = "Generate POJOs that by default will fail to deserialize unknown fields")
         private boolean strictObjects;
 
+        @CommandLine.Option(names = "--nonNullCollections",
+                defaultValue = "false",
+                description = "Generate POJOs that by default will fail to deserialize collections with null values")
+        private boolean nonNullCollections;
```
carterkozak, line 123: Mind adding this to the readme?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -249,6 +254,9 @@ private MethodSpec createSetter(EnrichedField enriched) {
                 .addMember("value", "$S", enriched.fieldName().get());
         if (isCollectionType(type)) {
             annotationBuilder.addMember("nulls", "$T.SKIP", Nulls.class);
+            if (featureFlags.contains(FeatureFlags.NonNullCollections) && !isOptionalInnerType(type)) {
```
carterkozak: A couple more edge cases (sorry!):
- I think inner type of `any` should be handled the same way as optional, optional.empty can go in, null is deserialized. Ideally we'd deserialize null (object) to empty optional, but I don't think jackson has a flag for that.
- Do we need to dealias `type`? I don't recall if `list<alias<optional<?>>>` is deserialized quite properly, if memory serves we get a null alias object instead of an alias of an empty optional.

ferozco: - The [spec](https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#51-built-in-types) says any is a primitive or an object with any fields which to me would suggest it be a non-nulllable value so I think we want to keep the current behaviour for any. 

- We currently do not correctly support aliases of optionals. I tried playing around with different flags to get it to work, but i wasn't successful. I'll push an additional test demonstrating to current behaviour

carterkozak: That's fair, any time an optional value is used for `any`, we should recommend using `optional<any>` instead. I got pushback on that from a product using the `any` type for backcompat where they didn't want to change the types in generated bindings for feat of breaking consumers. In that case they can defer adding the NonNullCollections flag until they've made the conjure definition change.

carterkozak:
👍

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -108,6 +109,10 @@ public static JavaFile generateAliasType(
                 .addStatement("return new $T(value)", thisClass)
                 .build());
 
+        if (typeDef.getAlias().accept(TypeVisitor.IS_OPTIONAL)) {
+            spec.addMethod(createDefaultConstructor(aliasTypeName));
+        }
```
carterkozak, line 115: Might be worth a comment explaining why we're adding a private no-arg constructor.

conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveOptionalsExample.java
```diff
@@ -213,7 +213,7 @@ public Builder from(PrimitiveOptionalsExample other) {
             return this;
         }
 
-        @JsonSetter(value = "num", nulls = Nulls.SKIP)
+        @JsonSetter(value = "num", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
         public Builder num(OptionalDouble num) {
```
carterkozak: What does `, contentNulls = Nulls.FAIL` do here? I thought that was for nested values in collections.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -249,6 +255,11 @@ private MethodSpec createSetter(EnrichedField enriched) {
                 .addMember("value", "$S", enriched.fieldName().get());
         if (isCollectionType(type)) {
```
carterkozak: isCollectionType applies to optional types as well as collections, I think we might need to take that into account below, and limit Optionals to applying nulls=SKIP

carterkozak:
Looks snazzy, great idea adding private no-arg constructors!

carterkozak:
I like it, thanks! 👍

PR #686, opened 2019-12-17
Fix Content-Type comparison to be case-insensitive

pkoenig10:
## Before this PR
The request `Content-Type` header comparison is case-sensitive. This results in errors when clients make request with `Content-Type` headers like `Application/json; charset=utf-8`.

## After this PR
The request `Content-Type` header comparison is case-insensitive. The HTTP spec states that media types values are case-insensitive:
https://tools.ietf.org/html/rfc7231#section-3.1.1.1

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Encoding.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Encoding.java
@@ -48,9 +48,9 @@ public interface Encoding {
 
     /**
      * Checks if a <pre>Content-Type</pre> or <pre>Accept</pre> value is supported by this encoding. This is not an
-     * exact match on {@link #getContentType()} because values may contain additional metadata, for
-     * example <pre>Content-Type: application/json; charset=utf-8</pre> may be supported by an {@link Encoding}
-     * which returns <pre>application/json</pre> from {@link #getContentType()}.
+     * exact match on {@link #getContentType()} because values are case-insensitive and may contain additional
+     * metadata, for example <pre>Content-Type: application/json; charset=utf-8</pre> may be supported by an
+     * {@link Encoding} which returns <pre>application/json</pre> from {@link #getContentType()}.
      */
     boolean supportsContentType(String contentType);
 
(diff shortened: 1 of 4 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Encodings.java
```diff
@@ -102,7 +103,7 @@ public boolean supportsContentType(String contentType) {
                 // TODO(ckozak): support wildcards? See javax.ws.rs.core.MediaType.isCompatible
                 return contentType != null
                         // Use startsWith to avoid failures due to charset
-                        && contentType.startsWith(CONTENT_TYPE);
+                        && contentType.toLowerCase(Locale.ROOT).startsWith(CONTENT_TYPE);
```
carterkozak: We may want to break out a utility function to loop over characters using `Characters.toLowerCase` and avoid potential re-allocation of the content-type string.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Encoding.java
```diff
@@ -48,9 +48,9 @@
 
     /**
      * Checks if a <pre>Content-Type</pre> or <pre>Accept</pre> value is supported by this encoding. This is not an
-     * exact match on {@link #getContentType()} because values may contain additional metadata, for
-     * example <pre>Content-Type: application/json; charset=utf-8</pre> may be supported by an {@link Encoding}
-     * which returns <pre>application/json</pre> from {@link #getContentType()}.
+     * exact match on {@link #getContentType()} because values are case-insensitive and may contain additional
+     * metadata, for example <pre>Content-Type: application/json; charset=utf-8</pre> may be supported by an
+     * {@link Encoding} which returns <pre>application/json</pre> from {@link #getContentType()}.
      */
```
carterkozak, line 55: last nit: would you mind adding the https://tools.ietf.org/html/rfc7231#section-3.1.1.1 link here?

carterkozak:
:+1:

PR #704, opened 2020-01-16
Upgrade tracing to trace the full request lifecycle

carterkozak:
## Before this PR
Only traced the blocking portion of the request.

## After this PR
==COMMIT_MSG==
Upgrade tracing to trace the full request lifecycle
==COMMIT_MSG==

## Possible downsides?
More tracing spans
slight overhead (was not measurable in benchmarks).


ferozco:
👍

PR #714, opened 2020-01-24
Indicate endpoint deprecation with a pseudo-standard header

markelliot:
## Before this PR
Clients only know about endpoint deprecation by taking API upgrades.

## After this PR
==COMMIT_MSG==
Clients may discover endpoint deprecation by observing return headers.
==COMMIT_MSG==

This unblocks our ability to localize deprecation warnings to the places where deprecated
endpoints are invoked.

## Possible downsides
None known.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/DeprecationReportingResponseHandler.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/DeprecationReportingResponseHandler.java
@@ -1,0 +1,44 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/DeprecationReportingResponseHandler.java
```diff
@@ -0,0 +1,44 @@
...
+package com.palantir.conjure.java.undertow.runtime;
+
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.util.HttpString;
+
+/**
+ * Adds HTTP response headers to indicate endpoint deprecation.
+ *
+ * <p>https://tools.ietf.org/id/draft-dalal-deprecation-header-01.html#rfc.section.2.1
```
carterkozak, line 26: Musing (no action requested): I wonder if it would be helpful to eventually include the deprecation docs in the response so old clients can log specific warning information. Vector of deprecation/version/endpoint allows us to find the same data without sending docs over the wire, but requires more lift to debug.

carterkozak:
I like this. We should document it in the conjure spec.

dansanduleac:
👍

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/DeprecationReportingResponseHandler.java
markelliot: I was a little worried about including the deprecation text as-is in the response in part because I think it might be surprising to existing API authors, but I'm open to either (a) communicating that that's what will happen in the future or (b) add another field to the definition explicitly for that purpose

carterkozak: That's fair, we should leave this as is and revisit if we find it's sub-optimal. The additional context wouldn't be meaningful for aggregate metrics anyhow.

PR #717, opened 2020-01-29
Convert featureFlags from enum to union

ferozco:
## Before this PR
We propogated feature flags as an enum which effectively restricted us to boolean properties.

## After this PR
Convert FeatureFlags from an enum to a union type allowing for more complex options (such as a package prefix) to be easily propogated

## Possible downsides?
We depend on another annotation processor and there was a bit of extra glue to make the union easy to use


iamdanfox:
I think derive4j is most useful when you want to do switches/pattern matching on variants of one 'thing', but I think we're not really benefitting from that functionality here, which means we're carrying a little more complexity than necessary.

What about replacing Set<FeatureFlag> with a bog standard value.immutable class that just encapsulates this info, maybe just called FeatureFlags?  This could internally store a boolean for the straightforward flags, and eventually will store a string for the shadowing?

ferozco:
Closing in favour of #718

PR #718, opened 2020-01-29
Convert feature flags set to Options object

ferozco:
## After this PR
Convert feature flags from set of enum to a POJO


iamdanfox (approved):
👌

PR #719, opened 2020-01-29
Add package prefix option

ferozco:
## Before this PR
users could not shade conjure generated classes

## After this PR
==COMMIT_MSG==
Add package prefix option to conjure-java
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/types/ObjectGeneratorTests.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/types/ObjectGeneratorTests.java
@@ -50,6 +50,16 @@ public final class ObjectGeneratorTests {
     }
 
     @Test
+    public void testObjectGenerator_allExamples_with_prefix() throws IOException {
+        ConjureDefinition def = Conjure.parse(ImmutableList.of(new File("src/test/resources/example-types.yml")));
+        List<Path> files = new ObjectGenerator(
+                        Options.builder().packagePrefix("test.prefix").build())
+                .emit(def, tempDir);
+
+        assertThatFilesAreTheSame(files, REFERENCE_FILES_FOLDER);
+    }
+
+    @Test
     public void testConjureImports() throws IOException {
         ConjureDefinition conjure = Conjure.parse(ImmutableList.of(
                 new File("src/test/resources/example-conjure-imports.yml"),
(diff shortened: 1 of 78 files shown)


ferozco:
For anyone that swings by this PR, the actual code change is quite small. We just check in a bunch of generated classes as a form of ete

conjure-java-core/src/test/java/com/palantir/conjure/java/types/ObjectGeneratorTests.java
```diff
@@ -49,6 +49,16 @@ public void testObjectGenerator_byteBufferCompatibility() throws IOException {
         assertThatFilesAreTheSame(files, REFERENCE_FILES_FOLDER);
     }
 
+    @Test
+    public void testObjectGenerator_allExamples_with_prefix() throws IOException {
+        ConjureDefinition def = Conjure.parse(ImmutableList.of(new File("src/test/resources/example-types.yml")));
+        List<Path> files = new ObjectGenerator(
+                        Options.builder().packagePrefix("test.prefix").build())
+                .emit(def, tempDir);
```
iamdanfox, line 57: I think there's a problem with these tests, which is that they don't really guarantee completeness. There are a lot of places in our generators where we need to combine packages & classnames, and if we forgot to use the prefix in one of those places, I think all our integrationTest generated classes would still compile because both the original `com.palantir.Foo` and `test.prefix.com.palantir.Foo` both exist on the same classpath, so if the `test.prefix` one happened to reach across and try and access the un-prefixed one, it would still work??

Options:
1. accept this risk and just steamroll onwards
2. put the generated code in some different sourceSet (`integrationTestInputPrefixed`) ?
3. make up a different input yml that we _only_ use for these prefix unit tests and not the others.

ferozco: @iamdanfox you're right that compilation does not demonstrate that all of references have been correctly prefixed. Happy to move them to a new SS if you'd like but I've verified the the prefixed classes do not have any imports to the original classes so there is no immediate risk

iamdanfox (approved):
Ok let's give it a whirl :)

conjure-java-core/src/test/java/com/palantir/conjure/java/types/ObjectGeneratorTests.java
carterkozak: +1, worst case things don't compile and an upgrade fails, no runtime risk.

PR #728, opened 2020-02-04
Support enum value, union type, and bean field deprecation

carterkozak:
## Before this PR
No support for per-field/value deprecation metadata.

## After this PR
==COMMIT_MSG==
Support enum value, union type, and bean field deprecation
==COMMIT_MSG==

https://github.com/palantir/conjure/pull/494


carterkozak:
This will require the conjure change to merge, I've verified tests pass when I use a local publication.

conjure-java-core/build.gradle
```diff
@@ -45,7 +45,9 @@ dependencies {
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-jdk8'
     compile 'com.google.googlejavaformat:google-java-format'
     compile 'com.google.guava:guava'
-    compile 'com.palantir.conjure:conjure-generator-common'
+    compile 'com.palantir.conjure:conjure-generator-common', {
+        exclude group: 'com.palantir.conjure.java', module: 'conjure-lib'
```
ferozco: why is this necessary?

carterkozak: Same reason this is duplicated on line 63, a recent change to conjure-generator-common must have added a new dependency on conjure-lib which breaks gcv write-locks picking between conjure-lib (5.3 from conjure project) and the local project link.

ferozco: Ok, I think we can make gradle's resolution strategy to prefer project dependencies https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ResolutionStrategy.html

ferozco:
👍 Thanks carter!

conjure-java-core/build.gradle
carterkozak: Ah fantastic, thanks!

PR #733, opened 2020-02-06
Fix empty unions

MarkRyanDev:
## Before this PR

If you try to define an empty union, the java codegen crashes.

## After this PR
==COMMIT_MSG==

Empty unions are generated as expected. This is not especially useful on its own, but can be valuable when the union will actually contain values in a later version of the api and you want to force clients to implement reasonable ways of handling the unknown case.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-733.v2.yml
+++ b/changelog/@unreleased/pr-733.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Defining an empty union no longer causes conjure-java to crash.
+  links:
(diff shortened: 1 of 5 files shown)


CRogers (approved):

ferozco:
@MarkRyanDev thanks for this fix! Looks like the changelog automation isn't working because you made the PR from the fork. Could you manually add a changelog in `changelog/@unreleased/` so we can get this in

ferozco:
dozer?

PR #758, opened 2020-02-13
[Stacked] Add dialogue verifier

ferozco:
## Before this PR
We did not test the generated client against the spec verifier

## After this PR
We do


conjure-java-client-verifier/src/test/java/com/palantir/conjure/java/compliance/AutoDeserializeTest.java
```diff
@@ -41,7 +42,9 @@
     public static final VerificationServerRule server = new VerificationServerRule();
 
     private static final Logger log = LoggerFactory.getLogger(AutoDeserializeTest.class);
-    private static final AutoDeserializeService testService = VerificationClients.autoDeserializeService(server);
+    private static final AutoDeserializeService jerseyTestService = VerificationClients.autoDeserializeService(server);
+    private static final BlockingAutoDeserializeService dialogueTestService =
```
iamdanfox, line 46: comment to justify that this probably gives us good enough coverage of the async one too?

conjure-java-client-verifier/src/test/java/com/palantir/conjure/java/compliance/SingleParamServicesTest.java
```diff
@@ -44,42 +44,61 @@
...
+                    VerificationClients.dialogueSinglePathParamService(server))
+            .putAll(
+                    "singleHeaderService",
+                    VerificationClients.singleHeaderService(server),
+                    VerificationClients.dialogueSingleHeaderService(server))
+            .putAll(
+                    "singleQueryParamService",
+                    VerificationClients.singleQueryParamService(server),
+                    VerificationClients.dialogueSingleQueryParamService(server))
+            .build();
```
iamdanfox, line 60: could we move this var into the `data()` func so it's clear we're only indexing into the multimap (which is a little gross) to build up the testcases?

conjure-java-client-verifier/src/test/java/com/palantir/conjure/java/compliance/SingleParamServicesTest.java
```diff
@@ -91,7 +110,7 @@ public void runTestCase() throws Exception {
 
         System.out.println(String.format("Invoking %s %s(%s)", serviceName, endpointName, jsonString));
 
-        Object service = servicesMaps.get(serviceName);
+        Object service = servicesMaps.get(serviceName).asList().get(serviceIndex);
```
iamdanfox, line 113: the .get() would change to just accessing one of the fields provided as a param

conjure-java-client-verifier/src/test/java/com/palantir/conjure/java/compliance/VerificationClients.java
```diff
@@ -18,16 +18,30 @@
...
 import com.palantir.conjure.verification.server.SinglePathParamService;
 import com.palantir.conjure.verification.server.SingleQueryParamService;
+import com.palantir.dialogue.JavaChannels;
+import com.palantir.tritium.metrics.registry.DefaultTaggedMetricRegistry;
 
 public final class VerificationClients {
     private VerificationClients() {}
 
+    private static final DefaultConjureRuntime DEFAULT_CONJURE_RUNTIME =
+            DefaultConjureRuntime.builder().build();
```
iamdanfox, line 43: I wonder if at some point we could dedupe the undertow & dialogue serializer registries?

gradle/verifier.gradle
```diff
@@ -85,6 +89,16 @@ task conjureJavaJersey(type: JavaExec) {
...
 
+task conjureJavaDialogue(type: JavaExec) {
+    main = "com.palantir.conjure.java.cli.ConjureJavaCli"
+    classpath = project(':conjure-java').sourceSets.main.runtimeClasspath
+    args 'generate',  "${-> configurations.verificationApi.singleFile}", 'src/generatedDialogue/java', '--dialogue'
+
+    inputs.file "${-> configurations.verificationApi.singleFile}"
+    outputs.dir "src/generatedDialogue/java"
+    doFirst { delete "src/generatedDialogue/java/*" }
+}
```
iamdanfox: can we hook this up to the first-class local codegen now? configuration containing the tgz file with a nice `builtBy` should work here?

iamdanfox (approved):

conjure-java-client-verifier/src/test/java/com/palantir/conjure/java/compliance/SingleParamServicesTest.java
ferozco: looks like we actually rely on having the full name: https://github.com/palantir/conjure-java/pull/758/files#diff-99ee67231f23433ff3738d89c722d01dR125 🤦‍♂

conjure-java-client-verifier/src/test/java/com/palantir/conjure/java/compliance/VerificationClients.java
carterkozak: I don't think we should, they cover similar concepts for separate uses so we would lose flexibility. I don't think they're large or complex enough for that to be worthwhile.

PR #768, opened 2020-02-18
On enum, generate values() similar to java enums.

hesselink:
I ran across the fact that conjure enums do no contain a `values` method like Java enums. I heard from others that ran into this in the past, so I figured I'd add it. Curious to hear what you think.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
@@ -3,6 +3,9 @@ package com.palantir.product;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.Preconditions;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Locale;
 import javax.annotation.Generated;
 import javax.annotation.Nonnull;
@@ -34,6 +37,9 @@ public final class EnumExample {
      */
     @Deprecated
     public static final EnumExample ONE_HUNDRED = new EnumExample(Value.ONE_HUNDRED, "ONE_HUNDRED");
+
+    private static final List<EnumExample> values =
+            Collections.unmodifiableList(Arrays.asList(ONE, TWO, ONE_HUNDRED));
 
     private final Value value;
 
@@ -95,6 +101,10 @@ public final class EnumExample {
             default:
                 return visitor.visitUnknown(string);
         }
+    }
+
+    public static List<EnumExample> values() {
+        return values;
     }
 
     @Generated("com.palantir.conjure.java.types.EnumGenerator")
--- a/changelog/@unreleased/pr-768.v2.yml
+++ b/changelog/@unreleased/pr-768.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: On enum, generate values() similar to java enums, which returns a list
+    containing all possible valid values.
+  links:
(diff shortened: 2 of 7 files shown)


iamdanfox:
Similar PRs have actually been proposed before ([example](https://github.com/palantir/conjure-java/pull/467)) and I think the crux of the problem is trying to make enums as forward compatible as possible, i.e. if an old client is talking to a new upstream server which starts returning new enum variants.

I think there's a _slight_ risk people would end up misusing the `.values()` method under the assumption that they know about all possible values.

What kind of use case do you have for this? are you also interested in using it in tests?

hesselink:
The use case here is an API advertising the capabilities of a backend. There's an enum listing all special fields (title, modified date, etc) for a certain backend, and the API wants to return all of them as e.g. sortable.

It's already possible to use `MyEnum.Values.values()`, which is what I'm using now. That has the same risk with mismatching client and server versions. In my case this risk doesn't exist, since I'm on the server talking about my own API.

hesselink:
Note that in the linked ticket, one of the use cases was also populating a dropdown, so it's not only test related.

hesselink:
@iamdanfox How do we move forward on this issue? I'm happy to discuss this in more depth on VTC if you want, or abandon if you say this is a hard no.

iamdanfox:
Given the number of requests for this and the relative obscurity of the downsides, I'd vote we just give it a go. Let's make sure we keep the list and unmodifiable stuff from https://github.com/palantir/conjure-java/pull/467/files

`./gradlew test -Drecreate=true` should fix up all the tests!

conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
```diff
@@ -2,7 +2,9 @@
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
+import com.google.common.collect.ImmutableList;
```
carterkozak: I think we avoid using guava in some configurations, so it's safer to use `Collections.unmodifiableList(Arrays.asList(<values>))`

conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
```diff
@@ -35,6 +37,8 @@
     @Deprecated
     public static final EnumExample ONE_HUNDRED = new EnumExample(Value.ONE_HUNDRED, "ONE_HUNDRED");
 
+    private static final List<EnumExample> VALUES = ImmutableList.of(ONE, TWO, ONE_HUNDRED);
```
carterkozak: What happens if there's an enum value with name `VALUES`? perhaps we should use `values` here instead?

hesselink: Ah, I checked and saw there was a dependency on guava, so preferred immutable over unmodifiable. I've switched now.

conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
hesselink: Ah, I checked and saw there was a dependency on guava, so preferred immutable over unmodifiable. I've switched now.

conjure-java-core/src/integrationInput/java/com/palantir/product/EnumExample.java
hesselink: Good call, done.

uschi2000: or `_VALUES`

carterkozak (approved):
Thanks!

PR #769, opened 2020-02-18
Add non-null annotation to all parameters with corresponding precondition

hesselink:
Conjure generates precondition nonnull checks, but doesn't annotate the function parameters as `Nonnull`. This means you can pass `Nullable` values without getting a warning, and get a runtime crash. This PR adds `Nonnull` annotations to all parameters with a nonnull precondition check.


iamdanfox (approved):
Thanks @hesselink, I like that this should give users even earlier feedback (IDE writing time > compile time > runtime)

PR #790, opened 2020-02-24
dialogue: FooServiceBlocking instead of BlockingFooService

iamdanfox:
## Before this PR

- **Blocking is here to stay:** The vast majority of clients use the blocking pattern, and I don't think we want to change this in the short-term as async code is a _lot_ harder to read & maintain, and would likely lead to a ton of broken error-handling/propagation code.

- **the prefixed naming convention is hard to scan**, as my brain ends up taking a very heavy emphasis on the first word 'Blocking' which is not the important differentiator when there are 10 clients constructed in a row.

- **it makes intellij prompt silly variable names** would prefer `catalogService` over `blockingCatalogService`.

Finally, I think it's discouraged to include the type/signature of a variable in it's name (e.g. we prefer `name` over `nameString` and `items` over `itemsSet`).  

## After this PR
==COMMIT_MSG==
The generated dialogue interfaces for a conjure FooService are now `FooServiceBlocking` and `FooServiceAsync` instead of `BlockingFooService` and `AsyncFooService`.
==COMMIT_MSG==

## Possible downsides?
- this is a break to the codegen... do we want to call it out as one??


carterkozak:
:+1:

uschi2000:
Is this pattern really so uncommon? Foo --> SimpleFoo, Store --> AltaStore/AtlasStore, FooGatekeeper. do you also still have DialogueFooService?

uschi2000:
Is this pattern really so uncommon? Foo --> SimpleFoo, Store --> AltaStore/AtlasStore, FooGatekeeper. do you also still have DialogueFooService?

uschi2000:
Is this pattern really so uncommon? Foo --> SimpleFoo, Store --> AltaStore/AtlasStore, FooGatekeeper. do you also still have DialogueFooService?

iamdanfox:
In the examples you mention, there's usually only one of them per repo, e.g. you try to have just one FooGatekeeper, and probably one AltaStore (and maybe a TestStore).  I think the clients are different because there are usually so many of them in each repo, which makes the visual scanning thing more of a concern.

iamdanfox:
Here's some snippets to compare if you're not convinced:
```groovy
        CompassService compass = clients.client(CompassService.class, "compass").get();
        PathService pathService = clients.client(PathService.class, "compass").get();
        OrganizationAndProjectService organizationAndProjectService =
                clients.client(OrganizationAndProjectService.class, "compass").get();
        BranchService branchService = clients.client(BranchService.class, "compass").get();
```

```groovy
        BlockingCompassService blockingCompass = clients.client(BlockingCompassService.class, "compass").get();
        BlockingPathService blockingPathService = clients.client(BlockingPathService.class, "compass").get();
        BlockingOrganizationAndProjectService blockingOrganizationAndProjectService =
                clients.client(BlockingOrganizationAndProjectService.class, "compass").get();
        BlockingBranchService blockingBranchService = clients.client(BlockingBranchService.class, "compass").get();
```

```groovy
        CompassServiceBlocking compass = clients.client(CompassServiceBlocking.class, "compass").get();
        PathServiceBlocking pathService = clients.client(PathServiceBlocking.class, "compass").get();
        OrganizationAndProjectServiceBlocking organizationAndProjectService =
                clients.client(OrganizationAndProjectServiceBlocking.class, "compass").get();
        BranchServiceBlocking branchService = clients.client(BranchServiceBlocking.class, "compass").get();
```

PR #812, opened 2020-03-02
conjure-undertow handles multi-value accept headers

carterkozak:
## Before this PR
We previously handled requests with multiple accept headers

## After this PR
now we also support accept headers with multiple comma-separated values.
==COMMIT_MSG==
conjure-undertow handles comma-separated multi-value accept headers
==COMMIT_MSG==

## Possible downsides?
a bit more complicated


ferozco:
👍

PR #819, opened 2020-03-04
Dialogue service generator avoids method references to the same type

carterkozak:
Reliance on interfaces to be functional will cause pain if new
fields are added in ways that are expected to be otherwise
compatible.

## After this PR
==COMMIT_MSG==
Dialogue service generator avoids method references to the same type
==COMMIT_MSG==


carterkozak:
Blocked on https://github.com/palantir/dialogue/pull/482

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/AsyncGenerator.java
```diff
@@ -182,8 +182,8 @@ private MethodSpec asyncClientImpl(ServiceDefinition serviceDefinition, Endpoint
                 REQUEST,
                 def.getReturns()
                         .filter(type -> type.accept(TypeVisitor.IS_BINARY))
-                        .map(type -> RUNTIME + ".bodySerDe()::deserializeInputStream")
-                        .orElseGet(() -> def.getEndpointName().get() + "Deserializer::deserialize"));
+                        .map(type -> RUNTIME + ".bodySerDe().inputStreamDeserializer()")
+                        .orElseGet(() -> def.getEndpointName().get() + "Deserializer"));
```
carterkozak, line 186: These two lines are the code change.

ferozco (approved):

PR #823, opened 2020-03-04
Update conjure-undertow default encodings

carterkozak:
The default configuration handles smile encoded data in addition
to cbor and json.

## After this PR
==COMMIT_MSG==
Update conjure-undertow default encodings
==COMMIT_MSG==


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -39,7 +39,9 @@
 
     private ConjureUndertowRuntime(Builder builder) {
         this.bodySerDe = new ConjureBodySerDe(
-                builder.encodings.isEmpty() ? ImmutableList.of(Encodings.json(), Encodings.cbor()) : builder.encodings);
+                builder.encodings.isEmpty()
+                        ? ImmutableList.of(Encodings.json(), Encodings.smile(), Encodings.cbor())
```
ferozco, line 43: Would be cool to wrap these in instrumented encodings so we can capture which encoding is being used when

ferozco (approved):

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
carterkozak: I like it, I'll implement something similar to the traced-encoding wrapper in a follow up

PR #824, opened 2020-03-05
Improve gitattributes

ferozco:
The current pattern was not matching all generated code


.gitattributes
```diff
@@ -1,2 +1,2 @@
 **/integrationInput/** linguist-generated=true
-**/src/test/resources/**.java* linguist-generated=true
+**/src/test/resources/**/*.java* linguist-generated=true
```
carterkozak, line 2: Ah nice!

carterkozak (approved):

PR #825, opened 2020-03-05
Fix annotating type rather than parameter with Nonnull

dansanduleac:
## Before this PR

#769 introduced a "Nonnull" annotation coming from `com.google.code.findbugs:jsr305` but applied it on the parameter type rather than on the parameter itself.

This led to a regression when defining such a parameter to a type that requires to be fully qualified (e.g. clashes with the currently defined class name) that would produce bean builder field setters like:
```java
package com.palantir.api;
class Foo {
  Builder fieldName(com.palantir.other.@Nonnull Foo value) { ... }
}
```

This is because we were annotating the TypeName, and according to https://github.com/square/javapoet/issues/431#issuecomment-195002442, when you do that javapoet assumes your annotation is a `TYPE_USE` annotation, which Nonnull is not.

## After this PR
==COMMIT_MSG==
Fix regression in 5.6.0 where conjure-java could generate bad Nonnull annotations when placed on a fully qualified parameter type.
==COMMIT_MSG==

```diff
 package com.palantir.api;
 class Foo {
-  Builder fieldName(com.palantir.other.@Nonnull Foo value) { ... }
+  Builder fieldName(@Nonnull com.palantir.other.Foo value) { ... }
 }
```

## Possible downsides?
No testing that this won't come back, but it's hard to catch this situation in all circumstances.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/Parameters.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/Parameters.java
@@ -1,0 +1,25 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ */
+
+package com.palantir.conjure.java.types;
+
+import com.squareup.javapoet.AnnotationSpec;
+import com.squareup.javapoet.ParameterSpec;
+import com.squareup.javapoet.TypeName;
+import javax.annotation.Nonnull;
+import javax.lang.model.element.Modifier;
+
+final class Parameters {
+
+    private Parameters() {}
+
+    static ParameterSpec nonnullParameter(TypeName typeName, String paramName, Modifier... modifiers) {
+        ParameterSpec.Builder builder = ParameterSpec.builder(typeName, paramName, modifiers);
+        if (typeName.isPrimitive()) {
+            return builder.build();
+        }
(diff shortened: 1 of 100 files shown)


dansanduleac:
@ferozco  @fawind not sure this will actually fix it, we are also using Nullable from jsr305 which _seems_ to work... so maybe the code needs to be changed to not reference the class directly, like https://github.com/palantir/conjure-java/blob/3f4ab405cf0c97164219f46c6191462920dcadb6/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java#L63 ?

dansanduleac:
~Actually this should be fine now.. tested on internal repo which encountered this issue and the broken link is now gone.~
Edit:  nope still broken... I'm not sure why but it keeps trying to reference the `NotNull` annotation too as if it's part of the package of the generated definition.

Edit: see updated description for the real bug that was causing this issue

conjure-java-core/src/main/java/com/palantir/conjure/java/types/Parameters.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
fawind, line 2: You forgot the apache license.

fawind (approved):

PR #830, opened 2020-03-06
Dialogue endpoints use version '0.0.0' over 'unknown' by default

carterkozak:
This prevents dialogue clients from spewing warnings on every
request.

## After this PR
==COMMIT_MSG==
Dialogue endpoints use version '0.0.0' over 'unknown' by default
==COMMIT_MSG==


ferozco:
👍

PR #840, opened 2020-03-18
Prevent ISE when rst_stream is received prior to async register-callback

carterkozak:
## Before this PR
```
java.lang.IllegalStateException: {throwable0_message}
	at io.undertow.server.HttpServerExchange.addExchangeCompleteListener(HttpServerExchange.java:941)
	at com.palantir.conjure.java.undertow.runtime.ConjureAsyncRequestProcessing.registerCallback(ConjureAsyncRequestProcessing.java:120)
	at com.palantir.conjure.java.undertow.runtime.ConjureAsyncRequestProcessing.register(ConjureAsyncRequestProcessing.java:98)
..etc
```

## After this PR
==COMMIT_MSG==
Prevent ISE when the exchange is completed prior to async register-callback
==COMMIT_MSG==

## Possible downsides?
Unfortunately I was not able to produce a test to reproduce this scenario, possibly due to okhttp h2-prior-knowledge shortcomings.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
@@ -116,8 +116,15 @@ final class ConjureAsyncRequestProcessing implements AsyncRequestProcessing {
             ListenableFuture<T> future, ReturnValueWriter<T> returnValueWriter, HttpServerExchange exchange) {
         // Attach data to the exchange in order to reuse a stateless listener
         exchange.putAttachment(FUTURE, future);
-        // Cancel the future if the exchange is completed before the future is done
-        exchange.addExchangeCompleteListener(COMPLETION_LISTENER);
+        // If the exchange is complete we still follow this path to provide standard logging.
+        if (exchange.isComplete()) {
+            // If the exchange has already completed (http/2 rst_stream while processing).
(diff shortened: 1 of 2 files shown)


iamdanfox (approved):

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAsyncRequestProcessing.java
```diff
@@ -116,8 +116,15 @@
             ListenableFuture<T> future, ReturnValueWriter<T> returnValueWriter, HttpServerExchange exchange) {
         // Attach data to the exchange in order to reuse a stateless listener
         exchange.putAttachment(FUTURE, future);
-        // Cancel the future if the exchange is completed before the future is done
-        exchange.addExchangeCompleteListener(COMPLETION_LISTENER);
+        // If the exchange is complete we still follow this path to provide standard logging.
+        if (exchange.isComplete()) {
```
jkozlowski, line 120: Is this not racy in any way? Is it not possible that exchange would complete in between you calling exchange.isComplete and exchange.addExchangeCompleteListener?

carterkozak: It shouldn't be since the data both read from is non-volatile.

jkozlowski: Sweet, I see that now.

PR #842, opened 2020-03-18
dialogue: suppress deprecation

iamdanfox:
## Before this PR

Dialogue produces a couple of classes, FooBlocking and FooAsync, where FooBlocking delegates to FooAsync. However, if a method on FooAsync is deprecated, then the caller in FooBlocking gets flagged by `-Xlint:deprecation`.

We discovered this on one of @pkoenig10's projects, https://internal-github/foundry/provenance/pull/1567

## After this PR
==COMMIT_MSG==
dialogue now adds a `@SuppressWarnings("deprecated")` to its own code when it needs to call out to a known deprecated method.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/BlockingGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/BlockingGenerator.java
@@ -21,6 +21,7 @@ import com.palantir.conjure.spec.EndpointDefinition;
 import com.palantir.conjure.spec.ServiceDefinition;
 import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.ConjureRuntime;
+import com.squareup.javapoet.AnnotationSpec;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.MethodSpec;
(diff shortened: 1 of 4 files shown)


ferozco:
👍

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/BlockingGenerator.java
```diff
@@ -74,6 +75,13 @@ private MethodSpec blockingClientImpl(EndpointDefinition def) {
                 .addModifiers(Modifier.PUBLIC)
                 .addParameters(params)
                 .addAnnotation(Override.class);
+
+        if (def.getDeprecated().isPresent()) {
+            methodBuilder.addAnnotation(AnnotationSpec.builder(SuppressWarnings.class)
+                    .addMember("value", "$S", "deprecated")
```
carterkozak, line 81: I think this needs s/deprecated/deprecation

iamdanfox: Argh 🤦‍♂

PR #856, opened 2020-03-30
add prefix to dialogue defined variables

### LINKED ISSUE #587: Colliding variable names ("channel") in dialogue api code-gen
## What happened?
If the api definition uses certain variable names, it causes the code-gen to have compilation errors.
`incompatible types: String cannot be converted to Channel`

Example:
"channel" is an arg in a twofamanager endpoint

The code-gen should generate a different name if "channel" or "runtime" are already being used.

ferozco:
Closes https://github.com/palantir/dialogue/issues/587

## Before this PR
User defined variables could conflict with internal variables

## After this PR
==COMMIT_MSG==
Add prefix to internal variable names to avoid collisions with user variables
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/changelog/@unreleased/pr-856.v2.yml
+++ b/changelog/@unreleased/pr-856.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Add prefix to internal variable names to avoid collisions with user
+    variables
+  links:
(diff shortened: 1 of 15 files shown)


carterkozak (approved):

iamdanfox:
🐌

PR #859, opened 2020-03-30
Defer expensive ser/de initialization for faster startup

carterkozak:
## After this PR
==COMMIT_MSG==
Defer expensive ser/de initialization for faster startup
==COMMIT_MSG==

Same as https://github.com/palantir/dialogue/pull/599


ferozco:
👍

PR #860, opened 2020-03-30
Fix dialogue auth header serialization

carterkozak:
This provides the expected `Bearer ` prefix. Note that this does
not use the plainSerDe methods, instead relying on the AuthHeader
toString implementation.

## After this PR
==COMMIT_MSG==
Fix dialogue auth header serialization by providing the expected `Bearer ` prefix.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-860.v2.yml
+++ b/changelog/@unreleased/pr-860.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Fix dialogue auth header serialization by providing the expected `Bearer
+    ` prefix.
+  links:
+  - https://github.com/palantir/conjure-java/pull/860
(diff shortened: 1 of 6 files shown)


carterkozak:
We've discussed removing sensitive data from AuthHeader.toString in the past, this would push us away from that goal, but it's not likely something we'll target any time soon.

ferozco:
> We've discussed removing sensitive data from AuthHeader.toString in the past, this would push us away from that goal, but it's not likely something we'll target any time soon.

We could codegen `"Bearer " + auth.getBearerToken()` if you would like to avoid relying on AuthHeader.toString

carterkozak:
I don't think that helps us since BearerToken.toString would be impacted as well if we wanted to move away from sensitive data in a toString method.

ferozco:
👍

carterkozak:
A specialized serializer method on the runtime object would allow us to decouple that decision, but probably not worth the squeeze for the time being. We already rely on toString pretty heavily for everything that's jaxrs-based, no reason we couldn't add the specialized pieces later on.

PR #861, opened 2020-03-30
Fix Dialogue cookie auth

ferozco:
## Before this PR
We would generate code that wouldn't compile for services with cookie auth

## After this PR
==COMMIT_MSG==
Dialogue generates valid code for services with cookie auth
==COMMIT_MSG==

## Possible downsides?
N/A


conjure-java-core/src/test/resources/test/api/CookieServiceAsync.java.dialogue
```diff
@@ -28,7 +28,7 @@ public interface CookieServiceAsync {
             @Override
             public ListenableFuture<Void> eatCookies(BearerToken token) {
                 Request.Builder _request = Request.builder();
-                _request.putHeaderParam("Cookie", "PALANTIR_TOKEN=" + _plainSerDe.serializeBearerToken(token));
```
carterkozak, line 31: Can we generate this in integration input for compilation?

carterkozak (approved):

PR #874, opened 2020-04-08
Dialogue ignores markers by design, no need to print 1k lines about it

carterkozak:
## Before this PR
```
conjure-dialogue does not support markers. In particular, it ignores @Safe annotations
conjure-dialogue does not support markers. In particular, it ignores @Safe annotations
conjure-dialogue does not support markers. In particular, it ignores @Safe annotations
conjure-dialogue does not support markers. In particular, it ignores @Safe annotations
...
```

## After this PR
==COMMIT_MSG==
Dialogue ignores markers by design, it's neither necessary nor helpful to print a warning to the console for ignored markers, especially when there's no identifying information describing the maker or the endpoint where it's defined.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-874.v2.yml
+++ b/changelog/@unreleased/pr-874.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix misleading console logging in dialogue code gen
+  links:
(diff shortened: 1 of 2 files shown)


ferozco:
👍

carterkozak:
:robot: :popcorn:

PR #879, opened 2020-04-23
Dialogue clients no longer hard-code "0.0.0" (optional --apiVersion param)

iamdanfox:
## Before this PR

Right now, every request made by a dialogue client has a user agent like `... FooServiceBlocking/0.0.0 ...`. This is suboptimal (and a regression from c-j-r) because server authors/operators can no longer assess how up-to-date their callers are.

## After this PR
==COMMIT_MSG==
When generating dialogue clients, conjure-java now allows users to specify an `--apiVersion` flag.
==COMMIT_MSG==

This exists to support the local codegen workflow, where you explicitly _don't_ want to just use the `Foo.class.getPackage().getImplementationVersion()`.

### DIFF
--- a/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
+++ b/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
@@ -157,9 +157,10 @@ public final class ConjureJavaCli implements Runnable {
         @Nullable
         private String packagePrefix;
 
-        @CommandLine.Option(names = "--apiVersion", description = "A optional version number (e.g. '1.2.3') to be "
-                + "embedded in the generated Java code. If omitted, the version will be read from the Jar Manifest's "
-                + "Implementation-Version.")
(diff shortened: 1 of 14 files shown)


conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -175,7 +181,7 @@ public void run() {
                 ServiceGenerator jerseyGenerator = new JerseyServiceGenerator(config.options());
                 ServiceGenerator retrofitGenerator = new Retrofit2ServiceGenerator(config.options());
                 ServiceGenerator undertowGenerator = new UndertowServiceGenerator(config.options());
-                ServiceGenerator dialogueServiceGenerator = new DialogueServiceGenerator(config.options(), "0.0.0");
```
iamdanfox, line 178: this made me chuckle.

carterkozak (approved):
This has fewer todos ^.^

PR #883, opened 2020-05-01
Annotate undertow service methods

lycarter:
## Before this PR
It's not possible to determine whether a server-side implementation contains a specific annotation.

## After this PR
==COMMIT_MSG==
Undertow service methods are now annotated
==COMMIT_MSG==


lycarter:
discussed with @carterkozak offline, we'll be pursuing a different approach

PR #886, opened 2020-05-01
Dialogue client bindings now have useful toString methods

iamdanfox:
## Before this PR

Currently a dialogue client has a toString that looks something like this:

```
com.palantir.dialogue.example.SampleServiceBlocking$1@5c080ef3
```

This is a little annoying because it's hard to eyeball whether internals are actually being re-used or now.

## After this PR
==COMMIT_MSG==
Dialogue blocking and async implementations now have a human-readable `toString` method
==COMMIT_MSG==

## Possible downsides?
- this could make it slightly harder to distinguish if you have tons and tons of clients (as previously you'd be able to see the `5c080ef3` change each time.)

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/CookieServiceBlocking.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/CookieServiceBlocking.java
@@ -24,7 +24,7 @@ public interface CookieServiceBlocking {
 
             @Override
             public String toString() {
-                return "CookieService{channel=" + _channel + ", runtime=" + _runtime + '}';
+                return "CookieServiceBlocking{channel=" + _channel + ", runtime=" + _runtime + '}';
             }
         };
     }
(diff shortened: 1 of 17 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/CookieServiceBlocking.java
```diff
@@ -20,6 +21,11 @@ static CookieServiceBlocking of(Channel _channel, ConjureRuntime _runtime) {
             public void eatCookies(BearerToken token) {
                 _runtime.clients().block(delegate.eatCookies(token));
             }
+
+            @Override
+            public String toString() {
```
carterkozak, line 26: Blocking?

carterkozak (commented):
Very nice! I think we should use the actual class name, otherwise lgtm

carterkozak (approved):

PR #892, opened 2020-05-11
Dialogue blocking clients require binary responses to be closed

carterkozak:
## Before this PR
Easy to leak resources.

## After this PR
==COMMIT_MSG==
Dialogue blocking clients require binary responses to be closed using error-prone annotations
==COMMIT_MSG==

## Possible downsides?
This adds an error-prone annotation dependency to conjure-lib.

### DIFF
--- a/conjure-lib/build.gradle
+++ b/conjure-lib/build.gradle
@@ -24,6 +24,7 @@ dependencies {
     compile 'com.palantir.conjure.java.api:errors'
     compile 'jakarta.annotation:jakarta.annotation-api'
     compile 'jakarta.ws.rs:jakarta.ws.rs-api'
+    compile 'com.google.errorprone:error_prone_annotations'
 
     testCompile 'org.assertj:assertj-core'
     testCompile 'org.mockito:mockito-core'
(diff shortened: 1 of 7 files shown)


iamdanfox:
Can we validate this for alias of binary too?

(I guess there's nothing we can do for `optional<binary>`?)

carterkozak:
Ya, I went for the simplest possible case here, but we can use more complex logic instead.

carterkozak:
I ended up checking the resolved return type against InputStream instead of attempting to duplicate the more complex dealiasing. Thoughts?

conjure-lib/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
     compile 'com.palantir.conjure.java.api:errors'
     compile 'jakarta.annotation:jakarta.annotation-api'
     compile 'jakarta.ws.rs:jakarta.ws.rs-api'
+    compile 'com.google.errorprone:error_prone_annotations'
```
carterkozak, line 27: note: This adds error-prone annotations to conjure-lib

iamdanfox (approved):

iamdanfox:
🐙

PR #893, opened 2020-05-12
Fix compilation failure when union types are called `completed`

carterkozak:
## Before this PR
Compilation failures.

## After this PR
==COMMIT_MSG==
Fix compilation failure when union types are called `completed`
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
@@ -68,7 +68,7 @@ public final class UnionGenerator {
     private static final String VALUE_FIELD_NAME = "value";
     private static final String UNKNOWN_WRAPPER_CLASS_NAME = "UnknownWrapper";
     private static final String VISIT_UNKNOWN_METHOD_NAME = "visitUnknown";
-    private static final String COMPLETED = "completed";
+    private static final String COMPLETED = "completed_";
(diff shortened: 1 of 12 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -68,7 +68,7 @@
     private static final String VALUE_FIELD_NAME = "value";
     private static final String UNKNOWN_WRAPPER_CLASS_NAME = "UnknownWrapper";
     private static final String VISIT_UNKNOWN_METHOD_NAME = "visitUnknown";
-    private static final String COMPLETED = "completed";
+    private static final String COMPLETED = "completed_";
```
carterkozak, line 71: code change

conjure-java-core/src/test/resources/example-types.yml
```diff
@@ -176,6 +176,7 @@ types:
           if: integer # some 'bad' member names!
           new: integer
           interface: integer
+          completed: integer
```
carterkozak, line 179: added a case that previously failed

iamdanfox (approved):

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
schlosna: Thanks @carterkozak this fixes my case. Resulting generated class name will be `Completed_StageVisitorBuilder`

PR #910, opened 2020-05-19
Dialogue codegen uses new 'bind' method to build EndpointChannels upfront

iamdanfox:
## Before this PR

Similar to what we already do in undertow-land, I've got a performance optimization in dialogue here https://github.com/palantir/dialogue/pull/764, which needs some new codegen.

## After this PR
==COMMIT_MSG==
conjure-java generates more efficient code for dialogue Async clients, by using the `bind` method to build EndpointChannels upfront.
==COMMIT_MSG==

`do not merge` until we actually release the corresponding dialogue PR.

## Possible downsides?
If people pick up this codegen but don't have the necessary `dialogue-target` version, then conjure-java will generate code that doesn't compile.


versions.props
```diff
@@ -9,7 +9,7 @@ com.palantir.conjure.java.api:* = 2.13.0
 com.palantir.conjure.java.runtime:* = 5.9.1
 com.palantir.conjure.verification:* = 0.18.5
 com.palantir.conjure:* = 4.8.0
-com.palantir.dialogue:* = 1.44.0
+com.palantir.dialogue:* = 9.9.9
```
ferozco: revert

build.gradle
```diff
@@ -52,6 +52,7 @@ allprojects {
         maven {
             url 'https://dl.bintray.com/palantir/releases/'
         }
+        mavenLocal()
```
ferozco: revert

ferozco (approved):
Code change looks good, just blocking on the dialogue change.

I think we have a default version of dialogue in gradle-conjure that we might want to upgrade too

carterkozak (approved):

iamdanfox:
Good shout - added the version to gradle-conjure here https://github.com/palantir/gradle-conjure/pull/420

PR #921, opened 2020-05-21
Generate endpoint markers as annotations on Dialogue interfaces

ferozco:
## Before this PR
We would completely ignore endpoint markers in Dialogue generated interfaces. Ideally we would replace markers with a more universal structure that every language can take care of. Doing so would require making changes to the IR which we currently do not have the bandwidth to do, so in lieu of first class support we mirror other implementations and add markers as annotations.

cc @akroy 

## After this PR
==COMMIT_MSG==
Generate endpoint markers as annotations on Dialogue interfaces
==COMMIT_MSG==

## Possible downsides?
Further couples us to endpoint markers which long term we would like to remove.

### DIFF
--- a/changelog/@unreleased/pr-921.v2.yml
+++ b/changelog/@unreleased/pr-921.v2.yml
@@ -1,0 +1,5 @@
+type: feature
+feature:
+  description: Generate endpoint markers as annotations on Dialogue interfaces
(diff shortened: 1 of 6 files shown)


carterkozak (approved):
Thanks!

PR #949, opened 2020-06-11
improve union visitor generation

berler:
## Before this PR
The Union accept method is not very efficient because it checks the value type for being an instance of each possible value wrapper. This is up to N instanceof checks where N is the number of distinct union value types.

## After this PR
The logic of which visitor method to call is delegated to each ValueWrapper class. The efficiency of calling the accept method on unions will now be the same no matter how many distinct value types there are.

==COMMIT_MSG==
Improves the union generator to avoid relying on instanceof checks to
pick the correct visit method. This makes using visitors with union
types more efficient when there are many possible value types, as it
avoids repeated checks for each possible value type.
==COMMIT_MSG==

## Possible downsides?

Introduces asymmetry between the way unions and enums are implemented. We might want to add a similar fix for enums, as they have the exact same problem when using visitors with many values.

### DIFF
--- a/changelog/@unreleased/pr-949.v2.yml
+++ b/changelog/@unreleased/pr-949.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Improve generated union classes to avoid instanceof checks.
+  links:
+  - https://github.com/palantir/conjure-java/pull/949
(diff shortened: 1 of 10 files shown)


carterkozak:
> We might want to add a similar fix for enums, as they have the exact same problem when using visitors with many values.

Enums could take advantage of a switch on the underlying `Value` enum rather than something like this.

carterkozak (approved):
Looks slick. Thanks, Berler!

iamdanfox (approved):

berler:
Ah yes you're right about enums. I missed that it was switching based on the internal enum, so that shouldn't need to change. The valueOf method for enums switches based on the string value (and uses that as the JsonCreator) so not sure if that can be improved or if switches on strings is already efficient.

Issue #960, opened 2020-06-17
Unexpected discrepencies between FooService and FooServiceBlocking when introducing new API versions


Alexis-D:
## What happened?

If we start/release a `FooService` defined as such:

```yaml
services:
  FooService:
    name: Foo Service
    package: com.palantir.foo
    base-path: /foo
    default-auth: header
    endpoints:
      foo:
        http: GET /foo
        args:
          arg:
            type: string
            param-type: query
        returns: Foo
```

And then update it to look like this (introducing a new optional query param `newArg`):

```yaml
services:
  FooService:
    name: Foo Service
    package: com.palantir.foo
    base-path: /foo
    default-auth: header
    endpoints:
      foo:
        http: GET /foo
        args:
          arg:
            type: string
            param-type: query
          newArg:
            type: optional<string>
            param-type: query
        returns: Foo
```

Then `FooService` will generated with a method like this:

```java
@Deprecated
default Foo foo(AuthHeader authHeader, String arg) {
    return foo(authHeader, arg, Optional.empty());
}
```

Which means that clients using `FooService` can upgrade their dependency without incurring any dev breaks. `FooServiceBlocking` however doesn't generate those methods, which makes releasing wire-safe API changes trigger dev breaks for `FooServiceBlocking` users.

## What did you want to happen?

`FooServiceBlocking` should generate deprecated methods in order to avoid dev breaks. There's a similar argument to be made with `FooServiceAsync`.

carterkozak:
Adding context:

We originally chose not to implement the backcompat methods because our goal was to use local codegen for dialogue. We later decided not to use local codegen because it became more complicated for some scenarios involving interactions with supporting libraries and external imports so it was never fully rolled out.
Revisiting our assumptions I think it's reasonable to generate the backcompat method stubs.

PR #967, opened 2020-06-22
Add backcompat stubs for optional header params

gsheasby:
## Before this PR
Similar to #960 - when an optional Header param is added to a method, no `@Deprecated` method without the header param is generated. This results in dev breaks, requiring any consumers to add either `Optional.empty()` or the header, as in [this commit](https://github.palantir.build/foundry/foundry/pull/6202/commits/7d3bd7819dcabed470afdfb32e70d24e53afdbbd).

## After this PR
==COMMIT_MSG==
When optional header params are added, Retrofit and Jersey generated code now includes deprecated methods without these headers - similar to those generated when optional query params are added.
==COMMIT_MSG==

## Possible downsides?
Still not added for Dialogue - may split this into another PR.

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/TestBase.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/TestBase.java
@@ -56,7 +56,7 @@ public abstract class TestBase {
                 Files.deleteIfExists(output);
                 Files.copy(file, output);
             }
-            assertThat(readFromFile(file)).isEqualTo(readFromFile(output));
+            assertThat(readFromFile(file)).describedAs(output.toString()).isEqualTo(readFromFile(output));
         }
     }
 }
--- a/conjure-java-core/src/test/resources/example-service.yml
+++ b/conjure-java-core/src/test/resources/example-service.yml
@@ -224,6 +224,21 @@ services:
             type: optional<rid>
             param-type: query
 
+      testOptionalHeaderParam:
+        http: POST /test-optional-header-param
+        args:
+          query: string
+          maybeHeader:
+            type: optional<string>
+            param-type: header
+            param-id: MaybeHeader
+          implicit:
+            type: rid
+            param-type: query
+          maybeQuery:
+            type: optional<rid>
+            param-type: query
+
       testBoolean:
         http: GET /boolean
         returns: boolean
--- a/changelog/@unreleased/pr-967.v2.yml
+++ b/changelog/@unreleased/pr-967.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: When optional header params are added, Retrofit and Jersey generated
+    code now includes deprecated methods without these headers - similar to those
+    generated when optional query params are added.
+  links:
+  - https://github.com/palantir/conjure-java/pull/967
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
@@ -200,6 +200,39 @@ public final class JerseyServiceGenerator implements ServiceGenerator {
             }
         }
 
+        if (!headerArgs.isEmpty()) {
+            // ensure the correct ordering of parameters by creating the complete sorted parameter list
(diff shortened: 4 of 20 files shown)


carterkozak:
I'm worried that defining something along these lines may not compile due to multiple equivalent method signatures:

```yaml
testOptionalHeaderParam:
  http: POST /test-optional
  args:
    maybeHeaderOne:
      type: optional<string>
      param-type: header
      param-id: MaybeHeaderOne
    maybeHeaderTwo:
      type: optional<string>
      param-type: header
      param-id: MaybeHeaderTwo
    maybeQueryOne:
      type: optional<string>
      param-type: query
    maybeQueryTwo:
      type: optional<string>
      param-type: query
```

conjure-java-core/src/test/java/com/palantir/conjure/java/TestBase.java
```diff
@@ -56,7 +56,7 @@ protected static void validateGeneratorOutput(List<Path> files, Path outputDir,
                 Files.deleteIfExists(output);
                 Files.copy(file, output);
             }
-            assertThat(readFromFile(file)).isEqualTo(readFromFile(output));
+            assertThat(readFromFile(file)).describedAs(output.toString()).isEqualTo(readFromFile(output));
```
carterkozak, line 59: Unsure if you're aware, we have a helper function to update the expected output `./gradlew test -Drecreate=true` documented [here](https://github.com/palantir/conjure-java/blob/develop/CONTRIBUTING.md#local-development)

conjure-java-core/src/test/resources/example-service.yml
```diff
@@ -224,6 +224,21 @@ services:
...
+          maybeHeader:
+            type: optional<string>
+            param-type: header
+            param-id: MaybeHeader
+          implicit:
+            type: rid
+            param-type: query
+          maybeQuery:
+            type: optional<rid>
+            param-type: query
```
carterkozak, line 240: This example is generated into the suffixed files, but those are never compiled -- it would be helpful to define this somewhere that outputs into a source set that's also compiled for additional verification, otherwise duplicate methods with the same signature won't be caught by CI.

gsheasby: do you have a place in mind where I could put this example?

I think the cases where type clashes may cause the headers to not compile would be exactly when the _first_ query param is optional and has the same type as _any_ header param. This could be detected in code and worked around by not generating backcompat methods for header params if they would clash. It's likely a corner case in practice, but I appreciate that we need something that always compiles.

carterkozak: Adding it to a new service defined in ete-service.yml should provide compilation verification. Using a new service instead of an existing one means you won't need to update any resource implementations to stub the method.

Is there danger in upgrading from an endpoint with an optional header to an endpoint with an optional header and optional query parameter that the same parameter index and type becomes used for another field?

gsheasby: I guess to be safe you'd want there to always be a non-optional param (of any kind) directly after the last optional header param.

gsheasby:
Caught up with @carterkozak via DM. Given the complexities involved in determining whether it's safe to have two dimensions of compatibility backfill methods (headers and query params), we won't be merging this.

iamdanfox:
@gsheasby is it one specific library that you're concerned about here? Could we explore a local codegen & relocation workflow to make it invincible to any kind of java ABI/API change?

a-k-g:
@carterkozak - would something like this work, to avoid the concern of equivalent method signatures:

1. Parameters in the generated methods would be in the same order as declared in the conjure definition. AuthHeader should always be the first, and any request object should be the last param. So for the example you gave, we should generate 
```
testOptionalHeaderParam(request);
testOptionalHeaderParam(maybeHeaderOne, request);
testOptionalHeaderParam(maybeHeaderOne, maybeHeaderTwo, request);
testOptionalHeaderParam(request, maybeHeaderOne, maybeHeaderTwo, maybeQueryOne, request);
testOptionalHeaderParam(request, maybeHeaderOne, maybeHeaderTwo, maybeQueryOne, maybeQueryTwo, request);
```
2. Any time new parameters are added, we enforce (using conjure-backcompat) that they have to be added at the end. `checkConjureBackCompat` should fail if there is an attempt to add a new parameter in between existing ones.

PR #975, opened 2020-06-24
Generated undertow service interfaces extend UndertowService now

fawind:
## Before this PR

Users had to wrap the implementation of their generated service interface with a service handler that collects the endpoints.
If they forget this and directly register their service, it would only fail at runtime.

```java
UndertowExampleService service = ...

// Current usage
server.api(ExampleEndpoints.of(service));

// Only fails at runtime as it uses the overloaded '#api(Object object)' method
// instead of `#api(UndertowService service)` 
server.api(service);
```

## After this PR

* The generated undertow service interface now extends `UndertowService` and provides a default `#endpoints(UndertowRuntime runtime)` method that delegates to the endpoints method of the generated service endpoints class.

```diff
UndertowExampleService service = ...

- server.api(ExampleEndpoints.of(service));
+ server.api(service) // This uses `#api(UndertowService service)` now
```

==COMMIT_MSG==
Generated undertow service interfaces extend UndertowService now
==COMMIT_MSG==


conjure-java-server-verifier/src/test/java/com/palantir/conjure/java/verification/server/undertest/UndertowServerUnderTestExtension.java
```diff
@@ -37,8 +36,7 @@
 
     @Override
     public void beforeAll(ExtensionContext _context) {
-        UndertowAutoDeserializeService autoDeserialize =
-                Reflection.newProxy(UndertowAutoDeserializeService.class, new EchoResourceInvocationHandler());
+        UndertowAutoDeserializeService autoDeserialize = new EchoUndertowAutoDeserializeService();
```
fawind: Using a proxy that calls a default method on a non-private-accessible interface is quite tricky as you have to use a different workaround depending on your jvm version which seems quite hard to maintain ([reference](https://dzone.com/articles/correct-reflective-access-to-interface-default-methods)). Based on that, I manually generated the implementation that mirrors the behavior the previously used proxy.

carterkozak:
As written I think there are a couple issues:
1. This will break folks who already use undertow services properly via `server.api(FooEndpoints.of(resource));`
2. It becomes more difficult to read the generated interface. The legibility concern doesn't really impact dialogue client interfaces because those aren't meant for humans to implement, but I think we should tread carefully around the undertow server interfaces which are meant to be implemented.

I wonder if we can solve both of these concerns by leaving the *Endpoints classes, and generating a single default method that delegates to `return *Endpoints.of(this)`.

There's also a corner case around the diamond problem for multiple inheritance when multiple undertow interfaces are implemented that we should reason through.

fawind:
> 1. This will break folks who already use undertow services properly via `server.api(FooEndpoints.of(resource));`

Can you expand on how this would break? In its current version `FooEndpoints.of(resource)` would delegate the endpoints call to the default method of  `resource#endpoints` which should be the same behavior just moved somewhere else?

carterkozak:
Ah sorry, I skimmed through and thought the *Endpoints classes were removed entirely, sorry!

markelliot:
wondering if we should leave the *Endpoints classes and delegate the calls the other way because:
* the Endpoint defintion classes will stay private
* the interfaces will stay more concise

Over time, we'd deprecate/migrate the use of *Endpoints classes and move them to become package private, only invoked as a delegate from the interface.

carterkozak:
+1, @markelliot articulated the proposal I was trying to make much better than I did.

fawind:
That sounds good to me! Also think the legibility concerns are very valid. Will iterate on this!

> There's also a corner case around the diamond problem for multiple inheritance when multiple undertow interfaces are implemented that we should reason through.

You mean both the *Endpoints classes and service interfaces implementing `UndertowService`? I might be missing context here, but I don't see a case where this would be a big problem?

carterkozak:
> You mean both the *Endpoints classes and service interfaces implementing UndertowService? I might be missing context here, but I don't see a case where this would be a big problem?

I think there are a few places where we have code matching this pattern:

```java
final class MyResource implements UndertowFooService, UndertowBarService {
     @Override
     public void fooMethod() {}

     @Override
     public void barMethod() {}
}
```
and
```java
MyResource resource = new MyResource();
wc.api(FooServiceEndpoints.of(resource));
wc.api(BarServiceEndpoints.of(resource));
```

carterkozak:
To clarify, I don't think this is a common pattern so I'm not requesting that it works without human intervention, just that we have a path forward for services that may be impacted.

fawind:
Added your suggestions:

* Reverted the changes to *Enpoints classes (but marked `*Endpoints.of(delegate)` as deprecated)
* The generated service interface still extends `UndertowService` but the default `endpoints` method now delegates to the respective method on the *Endpoints class.

---

> I think there are a few places where we have code matching this pattern:

Good catch! I found a handful of places searching for `implements\sUndertow.*Service\,\sUndertow.*Service` in sourcegraph.

I think we can resolve this case using the following pattern:

```java
final class MyResource implements UndertowFooService, UndertowBarService {
    @Override
    public List<Endpoint> endpoints(UndertowRuntime runtime) {
        return ImmutableList.<Endpoint>builder()
                .addAll(FooServiceEndpoints.of(this).endpoints(runtime))
                .addAll(BarServiceEndpoints.of(this).endpoints(runtime))
                .build();
     }

     @Override
     public void fooMethod() {}

     @Override
     public void barMethod() {}
}
```

Would this work?

carterkozak:
That should do it.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -151,6 +151,11 @@ public JavaFile generateServiceHandler(
                 .addField(FieldSpec.builder(serviceClass, DELEGATE_VAR_NAME, Modifier.PRIVATE, Modifier.FINAL)
                         .build())
                 .addMethod(MethodSpec.methodBuilder("of")
+                        .addAnnotation(Deprecated.class)
+                        .addJavadoc(
+                                "@Deprecated: You can now use {@link $1T} directly as it implements {@link $2T}.",
+                                serviceClass,
+                                UndertowService.class)
```
carterkozak: I don't think we should deprecate this if our recommendation for the diamond problem is to call it directly.

conjure-java-server-verifier/src/test/java/com/palantir/conjure/java/verification/server/undertest/EchoUndertowAutoDeserializeService.java
```diff
@@ -0,0 +1,500 @@
...
+import com.palantir.conjure.java.com.palantir.conjure.verification.types.StringExample;
+import com.palantir.conjure.java.com.palantir.conjure.verification.types.UuidAliasExample;
+import com.palantir.conjure.java.com.palantir.conjure.verification.types.UuidExample;
+
+/**
+ * Simple implementation of {@link UndertowAutoDeserializeService}, mirroring the behavior of a proxy using
+ * {@link EchoResourceInvocationHandler}. This implementation is used instead of a proxy, due to difficulties calling
+ * default methods on the non-private-accessible interface {@link UndertowAutoDeserializeService#endpoints}.
+ */
+final class EchoUndertowAutoDeserializeService implements UndertowAutoDeserializeService {
```
carterkozak: Builds will fail when conjure-verifier is upgraded since the methods on AutoDeserializeService may change. I have an internal link with an example of default method invocations that works on all supported jres, see github.p.b/g/conjure-java-service-adapter/pull/242

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
fawind: Good point, removed the deprecation flag.

conjure-java-server-verifier/src/test/java/com/palantir/conjure/java/verification/server/undertest/EchoUndertowAutoDeserializeService.java
fawind: That's a good solution. Updated the InvocationHandler proxy to handle default methods and removed the `EchoUndertowAutoDeserializeService`.

iamdanfox (approved):
This looks great, I think it'll really help people avoid surprising 404s when converting from jersey -> undertow :D

carterkozak:
Note that this will impact instrumentation on these interfaces, `wc.api(Tritium.instrument(MyService.class, impl, metrics));` won't actually produce any instrumentation because we instrument the default method which references the delegate.

PR #979, opened 2020-06-26
Include request line in generated dialogue/undertow services

Alexis-D:
## Before this PR

Generated dialogue/undertow services do not include the http request line.

## After this PR
==COMMIT_MSG==
Include request line in generated dialogue/undertow services
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/ServiceGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/ServiceGenerator.java
@@ -31,16 +31,16 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-public interface ServiceGenerator {
+public abstract class ServiceGenerator {
 
     /** Returns the set of Java files generated from the service definitions in the given conjure specification. */
-    Set<JavaFile> generate(ConjureDefinition conjureDefinition);
+    public abstract Set<JavaFile> generate(ConjureDefinition conjureDefinition);
 
     /**
      * Generates and emits to the given output directory all services and types of the given conjure definition, using
      * the instance's service and type generators.
      */
-    default List<Path> emit(ConjureDefinition conjureDefinition, File outputDir) {
+    public List<Path> emit(ConjureDefinition conjureDefinition, File outputDir) {
         List<Path> emittedPaths = new ArrayList<>();
         generate(conjureDefinition).forEach(f -> {
             Path emittedPath = Goethe.formatAndEmit(f, outputDir.toPath());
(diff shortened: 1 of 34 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/util/Javadoc.java
```diff
@@ -80,6 +82,10 @@ public static String render(Documentation documentation) {
                         + Javadoc.render(argument.getDocs().get()));
     }
 
+    public static String getRequestLine(HttpMethod httpMethod, HttpPath httpPath) {
+        return String.format("%s %s\n", httpMethod, httpPath);
```
carterkozak: I wonder if we should take advantage of the relatively new `@apiNote` javadoc tag?
`@apiNote Endpoint {@code GET /foo/{bar}/baz}`

I don't think a trailing newline should be necessary

conjure-java-core/src/main/java/com/palantir/conjure/java/services/ServiceGenerator.java
```diff
@@ -49,17 +49,24 @@
         return emittedPaths;
     }
 
-    static Optional<String> getJavaDoc(EndpointDefinition endpointDef) {
+    static Optional<String> getJavaDoc(EndpointDefinition endpointDef, boolean includeRequestLine) {
```
carterkozak: nit: boolean args aren't the easiest to read at call sites, thoughts on making this private and exposing `getJavaDoc` and `getJavaDocWithRequestLine`?

conjure-java-core/src/main/java/com/palantir/conjure/java/util/Javadoc.java
Alexis-D: Ah great I didn't know about `@apiNote`, I was wondering if there was a good tag to use, will update.

The newline is necessary (everything else gets one thanks to https://github.com/palantir/conjure-java/blob/24e6758e902f1b2c90a6e246b389ac6255be7c40/conjure-java-core/src/main/java/com/palantir/conjure/java/util/Javadoc.java#L61).

carterkozak: Makes sense, thanks for pointing it out :-)

conjure-java-core/src/test/resources/test/api/TestServiceBlocking.java.dialogue
```diff
@@ -30,46 +30,80 @@ import javax.annotation.Nonnull;
...
+     */
     InputStream getAliasedRawData(AuthHeader authHeader, ResourceIdentifier datasetRid);
 
+    /**
+     * @apiNote GET /catalog/datasets/{datasetRid}/raw-maybe
+     */
     Optional<InputStream> maybeGetRawData(AuthHeader authHeader, ResourceIdentifier datasetRid);
 
+    /**
+     * @apiNote GET /catalog/datasets/{datasetRid}/string-aliased
```
carterkozak: I don't think this is valid javadoc since the curly braces aren't escaped.
```suggestion
     * @apiNote {@code GET /catalog/datasets/{datasetRid}/string-aliased}
```

Alexis-D: From what I can tell it's fine and boils down to whether we're using an `@` sign, i.e. `{thisisfine}` v `{@thisiswrong}`, e.g.:

image

Generates the java below:

image

(when using ` -tag "apiNote:a:API Note:"`)

carterkozak: :+1:

Alexis-D: (actually `@code` gives us fixed-width fonts in the javadoc, so I'll update :))

carterkozak (approved):

conjure-java-core/src/test/resources/test/api/TestServiceBlocking.java.dialogue
carterkozak: Thanks :-)

PR #981, opened 2020-06-29
Expose framework exception

ferozco:
## Before this PR
We had defined a `FrameworkException` to work around `ServiceException` only supporting status codes defined in the Conjure wire spec. `FrameworkException` was package private preventing our server library from properly handling these exceptions.

## After this PR
==COMMIT_MSG==
Change `FrameworkException`'s visibility to public
==COMMIT_MSG==

## Possible downsides?
This doesn't solve underlying grossness of having two similar but slightly different structured exceptions. Ideally we would unify on a single type which supports arbitrary status codes. We would still expect consistency across the Conjure ecosystem since the language and code-gen would limit you to Spec's status codes


carterkozak:
I'd like to look at the current implementation and see if there are better options.

ferozco:
Sure, open to suggestions. We could play whack-a-mole and add non-spec status codes to [`ErrorType#Code`](https://github.com/palantir/conjure-java-runtime-api/compare/fo/add-custom-error-codes?expand=1) or we could replace `FrameworkException` with a new exception type that lives in conjure-java-runtime-api like we did `QosException`

PR #999, opened 2020-07-08
SafeExchangeCompletionListener also catches failures from `next.proceed()`

carterkozak:
==COMMIT_MSG==
SafeExchangeCompletionListener also catches failures from `next.proceed()`
==COMMIT_MSG==


iamdanfox (approved):

iamdanfox:
But what if log.error throws ?!!?! 

(just kidding)

carterkozak:
I like where your head's at! :-P

PR #1006, opened 2020-07-13
Fix for large values w/ double alias

raiju:
## Before this PR

For a conjure alias of double, whole numbers larger than 2^31 would fail to deserialize (see `testValueOf_largeDouble_long`). This is because there was a `@JsonCreator` for int and one for double, but not for long.

Relevant internal tickets: PDS-126734, PDS-126642

## After this PR
==COMMIT_MSG==
Fix for bug with alias'd doubles with whole numbers larger than 2^31
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
@@ -2,6 +2,7 @@ package com.palantir.product;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
+import com.palantir.conjure.java.lib.SafeLong;
 import javax.annotation.Generated;
 
 @Generated("com.palantir.conjure.java.types.AliasGenerator")
@@ -44,7 +45,8 @@ public final class DoubleAliasExample {
 
     @JsonCreator
     public static DoubleAliasExample of(long value) {
-        return new DoubleAliasExample((double) value);
+        long safeValue = SafeLong.of(value).longValue();
+        return new DoubleAliasExample((double) safeValue);
     }
 
     @JsonCreator
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
@@ -43,12 +43,12 @@ public final class DoubleAliasExample {
     }
 
     @JsonCreator
-    public static DoubleAliasExample of(int value) {
+    public static DoubleAliasExample of(long value) {
         return new DoubleAliasExample((double) value);
     }
 
     @JsonCreator
-    public static DoubleAliasExample of(long value) {
+    public static DoubleAliasExample of(int value) {
         return new DoubleAliasExample((double) value);
     }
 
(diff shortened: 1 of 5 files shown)


ferozco:
👍 Thanks @raiju! Does the same problem occur with safelong?

conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
```diff
@@ -47,6 +47,11 @@ public static DoubleAliasExample of(int value) {
         return new DoubleAliasExample((double) value);
     }
 
+    @JsonCreator
+    public static DoubleAliasExample of(long value) {
+        return new DoubleAliasExample((double) value);
```
markelliot, line 52: I think we probably need to check if the cast would end up imprecise on this

raiju: Maybe? If this is a double value that got encoded as an exact value it should be precisely representable. I can add an epsilon comparison if you want though.

conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
```diff
@@ -42,6 +42,11 @@ public static DoubleAliasExample of(double value) {
         return new DoubleAliasExample(value);
     }
 
+    @JsonCreator
+    public static DoubleAliasExample of(long value) {
+        return new DoubleAliasExample((double) value);
```
markelliot: I think we probably need to check if the cast would end up imprecise on this

raiju: I updated the code to only support numbers up to 2^53. That said, I don't think this is quite right. If I pass in a large double, jackson will happily deserialize lossily. I'm not sure why the behavior should be different when the `.` character is omitted.

markelliot: oof, we probably ought to figure out what Jackson settings we need to set to at least be able to log on floating point values that double can't represent

raiju: How would we detect that? We'd have to supply our own json decoder.

raiju: Can we merge this fix for now? Happy to own making the next change

conjure-java-core/src/test/java/com/palantir/conjure/java/types/AliasTests.java
```diff
@@ -43,4 +48,25 @@ public void testValueOf_external() {
     public void testValueOf_externalNested() {
         assertThat(ExternalLongAliasTwo.valueOf("3")).isEqualTo(ExternalLongAliasTwo.of(ExternalLongAliasOne.of(3L)));
     }
+
+    @Test
+    public void testValueOf_largeDouble_lossyDouble() throws JsonProcessingException {
+        // So large it doesn't fit in a double; jackson silently drops precision
+        assertThat(TEST_MAPPER.readValue("9007199254740992.234", DoubleAliasExample.class))
+                .isEqualTo(DoubleAliasExample.of(9007199254740992.0));
```
raiju, line 56: To clarify, this is what I meant when I said Jackson is lossy; I'm not sure why we'd accept `9007199254740993.0` but reject `9007199254740993` (the number cannot be represented in `double`)

raiju:
Marked as "merge when ready" as this fixes the issue without accepting any new lossy inputs (it expands it to +- 2^53; all losslessly converted numbers).

Let's have a separate discussion on what we want the behavior of doubles to be in Conjure on the conjure repo (the lossiness discussion isn't really a java-specific issue).

PR #1010, opened 2020-07-16
Improve the ConjureHandler builder ergonomics and composition

carterkozak:
Refactor exception handling onto the UndertowRuntime

This allows UndertowService objects to be registered with the
ConjureHandler much more easily, and reuses the same non-static
object to handle exceptions in both blocking and async request
processing.

This could be used to expand the capabilities provided by the
exception handler.

==COMMIT_MSG==
Improve the ConjureHandler builder ergonomics and composition

This allows UndertowService objects to be registered with the
ConjureHandler much more easily, and reuses the same non-static
object to handle exceptions in both blocking and async request
processing.
==COMMIT_MSG==

## Possible downsides?
The ExceptionHandler mechanism is not used by generated code, only consumed by the handler, which doesn't quite match the other functionality provided by the UndertowRuntime.
However, our ConjureHandler wraps each endpoint in code that could be generated, and there's no reason the functionality shouldn't be able to integrate with runtime components, so this is a reasonable place for exception handling.


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureHandler.java
```diff
@@ -141,8 +166,20 @@ public Builder addWrapperBeforeBlocking(EndpointHandlerWrapper wrapper) {
...
+            this.runtime = Preconditions.checkNotNull(value, "UndertowRuntime");
+            return this;
+        }
+
         public HttpHandler build() {
-            checkOverlappingPaths();
+            ImmutableList<Endpoint> allEndpoints = Streams.concat(
+                            endpoints.stream(), services.stream().flatMap(service -> service.endpoints(runtime).stream()
+                                    .map(Builder::checkEndpoint)))
+                    .collect(ImmutableList.toImmutableList());
```
ferozco: nit: find it clearer to have a separate variable for endpoints from a service
```suggestion
            Stream<Endpoint> serviceEndpoints = services.stream()
                    .flatMap(service -> service.endpoints(runtime).stream().map(Builder::checkEndpoint));
            ImmutableList<Endpoint> allEndpoints =
                    Streams.concat(endpoints.stream(), serviceEndpoints).collect(ImmutableList.toImmutableList());
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandler.java
```diff
@@ -16,37 +16,14 @@
...
 
     @Override
-    public void handleRequest(HttpServerExchange exchange) {
-        try {
-            delegate.handleRequest(exchange);
-        } catch (Throwable throwable) {
-            ConjureExceptions.handle(exchange, serializer, throwable);
-        }
+    public void handle(HttpServerExchange exchange, Throwable throwable) {
+        ConjureExceptions.handle(exchange, throwable);
```
ferozco: Any reason to keep the exception handling logic in a separate class? To augment the logic we should be able to define our own `ExceptionHandler` that takes this a delegate

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandler.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
ferozco: nit: I believe our stance is to avoid updating copyright date.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandler.java
```diff
@@ -16,37 +16,14 @@
...
-    ConjureExceptionHandler(HttpHandler delegate) {
-        this(delegate, ConjureExceptions.serializer());
-    }
-
-    // Constructor allows new exception handlers to be created without creating new serializer instances.
-    ConjureExceptionHandler(HttpHandler delegate, Serializer<SerializableError> serializer) {
-        this.delegate = delegate;
-        this.serializer = serializer;
-    }
+enum ConjureExceptionHandler implements ExceptionHandler {
```
ferozco: I think it will be a bit annoying for us to re-use this logic internally if we keep it package private

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -76,6 +77,11 @@ public AsyncRequestProcessing async() {
         return async;
     }
 
+    @Override
+    public ExceptionHandler exceptions() {
```
ferozco: I think we'll want to expose this on the builder

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/EndpointService.java
```diff
@@ -0,0 +1,42 @@
...
+import com.palantir.conjure.java.undertow.lib.UndertowRuntime;
+import com.palantir.conjure.java.undertow.lib.UndertowService;
+import java.util.List;
+
+/** Test utility to migrate away from deprecated methods. */
+final class EndpointService implements UndertowService {
+
+    private final Endpoint endpoint;
+
+    static UndertowService of(Endpoint endpoint) {
```
ferozco: might clean up the test code to let this take varargs of endpoints

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandler.java
carterkozak: Ya, I'm not too worried about opening visibility since it won't be useful/tempting to anyone else.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandler.java
carterkozak: The diff is a little misleading, the old version moved to ConjureExceptionHandlerHttpHandler which still has a 2018 header, and I reused this name for the Implementation of ExceptionHandler. Handler is overloaded between exchange handlers (HttpHandler) and exception handlers.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
carterkozak: :+1:

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandler.java
carterkozak: Yep, I've updated ConjureExceptions to an enum singleton which implements ExceptionHandler in order to avoid changing our most heavily used logger origin.

ferozco (approved):
Looks great! thanks @carterkozak

PR #1015, opened 2020-07-17
Dialogue bindings have a first-class of(EndpointChannelFactory, ConjureRuntime) method

iamdanfox:
## Before this PR

We made an optimization in dialogue where we can do some work upfront (in a `bind` method) rather than re-doing it on every single request.

To avoid breaking the world, we originally implemented this with a sneaky little `instanceof`, but this has caused a bunch of confusion so far, cos you get subtly different behaviour depending on what _subtype_ of channel you pass in.

## After this PR
==COMMIT_MSG==
All dialogue bindings now have a first-class factory method that takes the preferred `EndpointChannelFactory` type, which allows better performance than the plain Channel signature.
==COMMIT_MSG==


conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueInterfaceGenerator.java
```diff
@@ -74,7 +80,44 @@ private JavaFile generate(
...
+                .addParameter(ConjureRuntime.class, StaticFactoryMethodGenerator.RUNTIME)
+                .addCode(CodeBlock.builder()
+                        .add(
+                                "if ($L instanceof $T) { return $L(($T) $L, $L); }\n",
+                                StaticFactoryMethodGenerator.CHANNEL,
+                                EndpointChannelFactory.class,
+                                staticFactoryMethod.name,
+                                EndpointChannelFactory.class,
+                                StaticFactoryMethodGenerator.CHANNEL,
+                                StaticFactoryMethodGenerator.RUNTIME)
```
carterkozak, line 102: This isn't likely to produce issues like #908 because the factory is called once and the resulting EndpointChannel is reused for the lifespan of the client. :+1:

conjure-java-core/src/integrationInput/java/com/palantir/product/CookieServiceAsync.java
```diff
@@ -46,4 +47,21 @@ public String toString() {
             }
         };
     }
+
+    /**
+     * Creates an asynchronous/non-blocking client for a CookieService service.
+     */
+    static CookieServiceAsync of(Channel _channel, ConjureRuntime _runtime) {
```
iamdanfox, line 56: I thought about marking this method as deprecated, but it's previously hindered the rollout of conjure-java...

conjure-java-core/src/integrationInput/java/com/palantir/product/CookieServiceBlocking.java
```diff
@@ -31,4 +34,21 @@ public String toString() {
             }
```
carterkozak, line 35: nit: channel name is a little wonky, perhaps we use `return "CookieServiceBlocking{" + delegate  + '}';` and update `channel` naming in the async implementation. Not a blocker for this PR, toStrings can be updated separately.

iamdanfox: :+1: fixed

carterkozak (approved):

PR #1018, opened 2020-07-20
Tweak double equality around NaN, negative zero

iamdanfox:
## Before this PR

@padamowicz raised a question in #dev-foundry-infra about doubles, and it was noted that in [RFC 005](https://palantir.github.io/conjure/#/docs/rfc/005-handling-double-nan-and-infinity), we acknowledged that:

> java has inconsistent behaviour here between unboxed and boxed variants - we want the boxed behaviour.

```
// Java handles zero equality differently for unboxed and boxed values!
System.out.println(0.0d == -0.0d); // prints 'true'
System.out.println(new Double(0.0d).equals(new Double(-0.0d))); // prints 'false'
```

However our beans and aliases still use the surprising `==`, NOT the more sane boxed behaviour.

## After this PR
==COMMIT_MSG==
Conjure types involving `double` now have more sane behaviour:
- for negative zero: `0.0d` is no longer considered equal to `-0.0d` (previously it was)
- `NaN` values produced in different ways are now considered to be equal (e.g. `Math.sqrt(-1)` and `0d/0d`)

==COMMIT_MSG==

Open question
 - ~do we need to do this for the hashcode too?~ Edit: no, seems like Double.hashCode does the right thing already:
```java
    public static int hashCode(double value) {
        long bits = doubleToLongBits(value);
        return (int)(bits ^ (bits >>> 32));
    }
```
 - do we need to major rev for this?

(Just in case you're interested:

```
        double d1 = 0.0d / 0.0d;
        double d2 = Math.sqrt(-1);
        System.out.println(d1 + " " + Double.doubleToRawLongBits(d1) + " " + Double.doubleToLongBits(d1));
        System.out.println(d2 + " " + Double.doubleToRawLongBits(d2) + " " + Double.doubleToLongBits(d2));
```
Prints:
```
NaN 9221120237041090560 9221120237041090560
NaN -2251799813685248 9221120237041090560
```

### DIFF
--- a/conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/DoubleAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/DoubleAliasExample.java
@@ -26,7 +26,9 @@ public final class DoubleAliasExample {
     @Override
     public boolean equals(Object other) {
         return this == other
-                || (other instanceof DoubleAliasExample && this.value == ((DoubleAliasExample) other).value);
+                || (other instanceof DoubleAliasExample
+                        && Double.doubleToLongBits(this.value)
+                                == Double.doubleToLongBits(((DoubleAliasExample) other).value));
     }
 
     @Override
--- a/changelog/@unreleased/pr-1018.v2.yml
+++ b/changelog/@unreleased/pr-1018.v2.yml
@@ -1,0 +1,8 @@
+type: fix
+fix:
+  description: |-
+    Conjure types involving `double` now have more sane behaviour:
+    - for negative zero: `0.0d` is no longer considered equal to `-0.0d` (previously it was)
+    - `NaN` values produced in different ways are now considered to be equal (e.g. `Math.sqrt(-1)` and `0d/0d`)
+  links:
+  - https://github.com/palantir/conjure-java/pull/1018
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleAliasExample.java
@@ -26,7 +26,9 @@ public final class DoubleAliasExample {
     @Override
     public boolean equals(Object other) {
         return this == other
-                || (other instanceof DoubleAliasExample && this.value == ((DoubleAliasExample) other).value);
+                || (other instanceof DoubleAliasExample
+                        && Double.doubleToLongBits(this.value)
+                                == Double.doubleToLongBits(((DoubleAliasExample) other).value));
     }
 
     @Override
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/DoubleExample.java
@@ -29,7 +29,7 @@ public final class DoubleExample {
     }
 
     private boolean equalTo(DoubleExample other) {
-        return this.doubleValue == other.doubleValue;
+        return Double.doubleToLongBits(this.doubleValue) == Double.doubleToLongBits(other.doubleValue);
     }
 
     @Override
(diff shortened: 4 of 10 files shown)


conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/DoubleAliasExample.java
```diff
@@ -26,7 +26,9 @@ public String toString() {
     @Override
     public boolean equals(Object other) {
         return this == other
-                || (other instanceof DoubleAliasExample && this.value == ((DoubleAliasExample) other).value);
+                || (other instanceof DoubleAliasExample
+                        && Double.doubleToLongBits(this.value)
+                                == Double.doubleToLongBits(((DoubleAliasExample) other).value));
```
iamdanfox, line 31: This is what the boxed `Double.equals` does under the hood:

http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/lang/Double.java#l797

ferozco: Could we use `Double.compare()`?

iamdanfox: We could, but not sure what the benefit would be since it's just using this `doubleToLongBits` method under the hood anyway?

```java
    public static int compare(double d1, double d2) {
        if (d1 < d2)
            return -1;           // Neither val is NaN, thisVal is smaller
        if (d1 > d2)
            return 1;            // Neither val is NaN, thisVal is larger

        // Cannot use doubleToRawLongBits because of possibility of NaNs.
        long thisBits    = Double.doubleToLongBits(d1);
        long anotherBits = Double.doubleToLongBits(d2);

        return (thisBits == anotherBits ?  0 : // Values are equal
                (thisBits < anotherBits ? -1 : // (-0.0, 0.0) or (!NaN, NaN)
                 1));                          // (0.0, -0.0) or (NaN, !NaN)
    }
```

ferozco: you get to avoid the more expensive `Double.doubleToLongBits` on the common path where values are different and we get to simplify generated code

iamdanfox: @ferozco We could do some benchmarking to figure out whether the branching of checking d1 < d2 is faster than just running the `doubleToLongBits` every time, but given that in the jdk's Double.equals method they decided to use:
```java
    public boolean equals(Object obj) {
        return (obj instanceof Double)
               && (doubleToLongBits(((Double)obj).value) ==
                      doubleToLongBits(value));
    }
```

I'm kinda inclined to just do the same thing they picked? (My hunch is that because `doubleToRawLongBits` is literally just the [in memory representation](https://github.com/openjdk/jdk/blob/6bab0f539fba8fb441697846347597b4a0ade428/src/hotspot/cpu/x86/templateInterpreterGenerator_x86_32.cpp#L291-L312), it's fast enough)

ferozco (approved):

conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/DoubleAliasExample.java
carterkozak: Performance should be equivalent outside of tight loops -- cost of doubleToRawLongBits is near zero.

carterkozak (approved):

carterkozak:
Actually, there's an interesting question about NAN in conjure sets... on the producing side we may have two distinct nans, but the consumer will receive two identical nan values.

iamdanfox:
Yeah I think NaNs are weird enough that I'm sure we'll find more edge cases beyond the one fixed in this PR. Maybe we'll end up making in our own specialized DoubleSet at some point...

PR #1022, opened 2020-07-21
conjure-undertow drains request bytes prior to writing responses

carterkozak:
Note that we already effectively provide this behavior when the
response is sufficiently small (fits into some combination of the
web server and system buffers) where the request is drained after
the response has been sent. This change allows us to provide the
same behavior in the presence of large responses.

==COMMIT_MSG==
conjure-undertow drains request bytes prior to writing responses
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/EteBinaryResource.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/EteBinaryResource.java
@@ -43,7 +43,7 @@ final class EteBinaryResource implements EteBinaryService {
             @NotNull AuthHeader authHeader, int bytesToRead, @NotNull InputStream body) {
         if (bytesToRead > 0) {
             try {
-                ByteStreams.copy(ByteStreams.limit(body, bytesToRead), ByteStreams.nullOutputStream());
+                ByteStreams.exhaust(ByteStreams.limit(body, bytesToRead));
             } catch (IOException e) {
                 throw new RuntimeException(e);
             }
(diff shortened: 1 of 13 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -428,6 +431,36 @@ public void testBinaryServerSideFailureAfterFewBytesSent() {
                 .isInstanceOf(RemoteException.class);
     }
 
+    @Test
+    @Timeout(20)
+    public void testBinaryServerSideFailureAfterFewBytesReceived() {
+        byte[] data = new byte[1024];
+        ThreadLocalRandom.current().nextBytes(data);
+        assertThatThrownBy(() -> {
```
carterkozak, line 439: retrofit hangs until it reaches the client write timeout without this change.

conjure-java-core/src/test/java/com/palantir/conjure/java/EteBinaryResource.java
```diff
@@ -32,6 +38,20 @@ public StreamingOutput postBinary(AuthHeader _authHeader, InputStream body) {
         return output -> output.write(ByteStreams.toByteArray(body));
     }
 
+    @Override
+    public StreamingOutput postBinaryThrows(
+            @NotNull AuthHeader authHeader, int bytesToRead, @NotNull InputStream body) {
+        if (bytesToRead > 0) {
+            try {
+                ByteStreams.copy(ByteStreams.limit(body, bytesToRead), ByteStreams.nullOutputStream());
```
iamdanfox: Is this just the same as `ByteStreams.exhaust(ByteStreams.limit())` ?

carterkozak: Yep, I wasn't aware of exhaust.

iamdanfox (approved):

PR #1034, opened 2020-07-31
improvement: dialogue interfaces ignore undertow async marker

ferozco:
## Before this PR
We would generate dialogue interfaces with a non-existant annotation if users chose to use the experimentalUndertowAsyncMarker

## After this PR

==COMMIT_MSG==
Dialogue interfaces ignore undertow async marker
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/changelog/@unreleased/pr-1034.v2.yml
+++ b/changelog/@unreleased/pr-1034.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Dialogue interfaces ignore undertow async marker
+  links:
+  - https://github.com/palantir/conjure-java/pull/1034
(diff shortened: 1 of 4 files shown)


carterkozak:
I think we want Undertow interfaces to ignore them as well. We use the marker as a hint to the generator, the return value tells us whether or not methods are async at compile and runtime!

ferozco:
I was under the impression that undertow already omitted the annotation, will fix

ferozco:
@carterkozak it doesn't actually look like markers are used by undertow besides checking for the async flag so I think this is ready to go

carterkozak (approved):

PR #1038, opened 2020-08-04
Generated objects memoized hash code is not volatile

carterkozak:
The Java memory model does not allow word tearing and we don't
expect concurrent hashing across threads. In cases where that
occurs there is already a race that allows multiple computations,
which are relatively inexpensive.

See [String.java](https://github.com/openjdk/jdk/blob/1f63603288670e6a01ae52930fc7daeec5b15ffc/src/java.base/share/classes/java/lang/String.java#L164-L165) for a similar example. Note that they use a second field to track whether the hash code result was zero to prevent adversarial inputs from producing poor performance (denial of service), which isn't important in our case.

## After this PR
==COMMIT_MSG==
Generated objects memoized hash code is not volatile
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1038.v2.yml
+++ b/changelog/@unreleased/pr-1038.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Generated objects memoized hash code is not volatile
+  links:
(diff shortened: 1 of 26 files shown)


ferozco (approved):

carterkozak:
:robot:

PR #1051, opened 2020-08-18
fix #1050: Undertow endpoints disallow request body collections with null values

### LINKED ISSUE #1050: Ensure collection request bodies don't contain null values
## What happened?
A request argument that is a collection type can contain `null` as a value. This leads to either `NullPointerException` (500) or requires defensive null checking.

Deserializing a collection value should fail with a 400/422 if any of the values in the collection are `null`. 

There shouldn't be any backwards compatibility concerns since this would only apply to request bodies that are collections.

carterkozak:
==COMMIT_MSG==
Undertow endpoints disallow request body collections with null values when `--nonNullCollections` is set
==COMMIT_MSG==

## Possible downsides?
Some consumers may rely on collections including null values.

### DIFF
--- a/changelog/@unreleased/pr-1051.v2.yml
+++ b/changelog/@unreleased/pr-1051.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Undertow endpoints disallow request body collections with null values
+    when `--nonNullCollections` is set
+  links:
+  - https://github.com/palantir/conjure-java/pull/1051
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -350,7 +350,7 @@ final class UndertowServiceHandlerGenerator {
     private TypeName immutableCollection(TypeName input) {
         // Note that only the outermost collection is considered for replacement to avoid
         // generic type incompatibilities.
-        if (options.nonNullCollections() && input instanceof ParameterizedTypeName) {
+        if (options.nonNullTopLevelCollectionValues() && input instanceof ParameterizedTypeName) {
             ParameterizedTypeName parameterized = (ParameterizedTypeName) input;
             if (LIST_NAME.equals(parameterized.rawType)) {
                 return ParameterizedTypeName.get(
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -348,6 +348,8 @@ final class UndertowServiceHandlerGenerator {
     private static final ClassName IMMUTABLE_SET_NAME = ClassName.get(ImmutableSet.class);
 
     private TypeName immutableCollection(TypeName input) {
+        // Note that only the outermost collection is considered for replacement to avoid
+        // generic type incompatibilities.
         if (options.nonNullCollections() && input instanceof ParameterizedTypeName) {
             ParameterizedTypeName parameterized = (ParameterizedTypeName) input;
             if (LIST_NAME.equals(parameterized.rawType)) {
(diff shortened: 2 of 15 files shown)


carterkozak:
I've validated compilation in a large internal project.

conjure-java-core/src/test/resources/ete-service.yml
```diff
@@ -223,3 +223,13 @@ services:
...
+
+      receiveListOfOptionals:
+        http: PUT /list/optionals
+        args:
+          value: list<optional<string>>
+
+      receiveSetOfOptionals:
+        http: PUT /set/optionals
+        args:
+          value: set<optional<string>>
```
iamdanfox, line 235: is it worth also validating the case where you receive a raw map? like `map<string,string>`, I guess we probably want to reject `{"hello": null}`?

carterkozak: I think so, but I didn't include it in this change because immutablemap will additionally validate against duplicate keys. I'd prefer to implement that separately.

iamdanfox:
Ok so the question of how to roll this out seems a bit interesting - a quick sourcegraph search shows that 263 repos have opted in to nonNullCollections, although it's harder to get a sense of how many of those repos actually have _raw_ lists/sets in a request body. I expect there to be just a _few_ services that this will affect, including the one that @pkoenig10 and @tusharnarayan mention in https://github.com/palantir/conjure-java/issues/1050.

There's a _slight_ chance this could end up causing a P0 somewhere, but if it does I think the solution will be to just blocklist and revert whichever server started rejecting nulls that were important for whatever reason and reassess.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -339,6 +342,28 @@ private TypeSpec generateEndpointHandler(
...
 
+    private static final ClassName LIST_NAME = ClassName.get(List.class);
+    private static final ClassName IMMUTABLE_LIST_NAME = ClassName.get(ImmutableList.class);
+    private static final ClassName SET_NAME = ClassName.get(Set.class);
+    private static final ClassName IMMUTABLE_SET_NAME = ClassName.get(ImmutableSet.class);
+
+    private TypeName immutableCollection(TypeName input) {
+        // Note that only the outermost collection is considered for replacement to avoid
+        // generic type incompatibilities.
+        if (options.nonNullCollections() && input instanceof ParameterizedTypeName) {
```
iamdanfox: @tusharnarayan can you remind me why your service (f/c ?) doesn't currently use nonNullCollections?

pkoenig10: We persist Conjure objects with collections in a number of places. It's difficult for us to be confident that nulls have not been persisted.

pkoenig10: The path forward is for us to replace Conjure objects in our storage layer with out own objects that allow nulls, but this is a non-trivial undertaking.

pkoenig10: For context, see f/c/pull/9962.

tusharnarayan: We'll just pick up the work to audit and deal with any stored Conjure objects, and then will turn on nonNullCollections - once this PR is merged, should solve the case I outlined in https://github.com/palantir/conjure-java/issues/1050

carterkozak:
Added a `--nonNullTopLevelCollectionValues` flag to opt into this without modifying objects. Existing consumers with null collection validation will still get this by default.

changelog/@unreleased/pr-1051.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: Undertow endpoints disallow request body collections with null values
+    when `--nonNullCollections` is set
```
pkoenig10, line 4: Need to update this changelog

carterkozak: I don't really want anyone except you to know about the new flag.

iamdanfox (approved):

PR #1086, opened 2020-09-14
Annotate enums with error-prone Immutable

### LINKED ISSUE #700: Enums should be annotated with @Immutable
## What happened?

```
warning: [ImmutableEnumChecker] enums should be immutable: 'VectorFieldType' has field 'conjureType' of type 'com.palantir.tiles2.api.config.ConjureVectorFieldType', the declaration of type 'com.palantir.tiles2.api.config.ConjureVectorFieldType' is not annotated with @com.google.errorprone.annotations.Immutable
```

When I define a type:

```java
public enum VectorFieldType {
  Number(ConjureVectorFieldType.NUMBER);

  private final ConjureVectorFieldType conjureType;

  VectorFieldType(ConjureVectorFieldType conjureType) {
    this.conjureType = conjureType;
  }
}
```

The conjure generated enums should be annotated with `@com.google.errorprone.annotations.Immutable` if they truly are (they are).

iamdanfox:
## Before this PR

fixes https://github.com/palantir/conjure/issues/700

## After this PR
==COMMIT_MSG==
The java classes for conjure enums are now annotated with `@com.google.errorprone.annotations.Immutable`, which ensures that if they're used inside hand-rolled enums, Google's [ImmutableEnumChecker](https://errorprone.info/bugpattern/ImmutableEnumChecker) won't complain.
==COMMIT_MSG==

We already have the necessary jar on the `conjure-lib` classpath, so this should be literally zero downside... https://github.com/palantir/conjure-java/blob/5130885768f0a147e7a69dfd29e3ef06cfb8a138/conjure-lib/build.gradle#L27

Only open question is should we just do this for beans/unions too??

### DIFF
--- a/changelog/@unreleased/pr-1086.v2.yml
+++ b/changelog/@unreleased/pr-1086.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: The java classes for conjure enums are now annotated with `@com.google.errorprone.annotations.Immutable`,
+    which ensures that if they're used inside hand-rolled enums, Google's [ImmutableEnumChecker](https://errorprone.info/bugpattern/ImmutableEnumChecker)
+    won't complain.
+  links:
+  - https://github.com/palantir/conjure-java/pull/1086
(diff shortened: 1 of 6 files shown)


carterkozak:
I'm generally in favor of taking advantage of error-prone functionality, I think we may want to update the gradle plugin to add error-prone annotations as an implementation dependency before this is released.

ferozco (approved):

iamdanfox:
Just rescuing this from stalebot - I'm pretty sure this will compile due to the [transitive provided error-prone annotations](https://github.com/palantir/conjure-java/blob/5130885768f0a147e7a69dfd29e3ef06cfb8a138/conjure-lib/build.gradle#L27) on conjure-lib, but will FLUP to add it to gradle-conjure anyway.

PR #1092, opened 2020-09-22
Dialogue async services toString uses the correct class name

carterkozak:
==COMMIT_MSG==
Dialogue async service toString uses the correct class name
==COMMIT_MSG==


iamdanfox (approved):
Oops, thanks @carterkozak!

PR #1102, opened 2020-10-02
fix #1099: conjure-undertow correctly dealiases binary responses

### LINKED ISSUE #1099: Incorrect content-type header for binary alias return type
## What happened?
When using the Undertow service generator, using a return type that is an alias of binary type results in an incorrect content-type header being added to the response. In this situation we'd expect to receive the content-type of `application/octet-stream` but instead receive `application/json` which results in errors in the conjure client like the following:
```
Unsupported Content-Type: {received=application/json, supportedEncodings=[EncodingDeserializerContainer{encoding=BinaryEncoding{application/octet-stream}, deserializer=InputStreamDeserializer{}}]}
com.palantir.logsafe.exceptions.SafeRuntimeException: Unsupported Content-Type: {received=application/json, supportedEncodings=[EncodingDeserializerContainer{encoding=BinaryEncoding{application/octet-stream}, deserializer=InputStreamDeserializer{}}]}
	at com.palantir.conjure.java.dialogue.serde.ConjureBodySerDe$EncodingDeserializerRegistry$1.deserialize(ConjureBodySerDe.java:298)
	at com.palantir.conjure.java.dialogue.serde.ConjureBodySerDe$EncodingDeserializerRegistry.deserialize(ConjureBodySerDe.java:257)
	at com.palantir.dialogue.futures.DialogueDirectTransformationFuture.onSuccess(DialogueDirectTransformationFuture.java:103)
```

For other service generators, like Dialogue and Jersey, it looks like binary alias return types are de-aliased during interface generation so that we don't end up using the binary alias type at all for return types and don't run into this issue. However, for the Undertow interfaces, we end up using the binary alias return type and I'm guessing we end up falling back to the default `application/json` in https://github.com/palantir/conjure-java/blob/develop/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java#L119-L137.

In this situation, I'd want/expect the content-type to be appropriately set to `application/octet-stream` when using an endpoint return type of a binary alias.

Just thinking out loud, there are probably a few different ways to go about this. We would de-alias binary alias return types in the Undertow interfaces to be consistent with the other generators, though that could cause compile breaks in consumers. Additionally, I'm guessing we could add additional logic into the `ConjureBodySerDe` class to detect and handle this situation.

carterkozak:
==COMMIT_MSG==
conjure-undertow correctly dealiases binary response types
==COMMIT_MSG==


versions.lock
```diff
@@ -85,7 +85,7 @@ ch.qos.logback:logback-access:1.2.3 (1 constraints: b41148e2)
 ch.qos.logback:logback-classic:1.2.3 (4 constraints: d232d9c3)
 ch.qos.logback:logback-core:1.2.3 (3 constraints: 1a28577f)
 com.fasterxml:classmate:1.3.4 (1 constraints: 9b122713)
-com.fasterxml.jackson.datatype:jackson-datatype-joda:2.11.2 (1 constraints: af0e575e)
```
carterkozak: I didn't touch deps, but builds failed locally unless I updated the locks file.

conjure-java-core/src/test/resources/test/api/TestServiceEndpoints.java.undertow
```diff
@@ -277,12 +277,12 @@ public final class TestServiceEndpoints implements UndertowService {
...
         private final TestService delegate;
 
-        private final Serializer<NestedAliasedBinary> serializer;
+        private final Serializer<BinaryResponseBody> serializer;
 
         GetAliasedRawDataEndpoint(UndertowRuntime runtime, TestService delegate) {
             this.runtime = runtime;
             this.delegate = delegate;
-            this.serializer = runtime.bodySerDe().serializer(new TypeMarker<NestedAliasedBinary>() {});
+            this.serializer = runtime.bodySerDe().serializer(new TypeMarker<BinaryResponseBody>() {});
```
carterkozak: bug: Serializer should not be created

carterkozak: fixed

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SpecializeBinaryClassNameVisitor.java
```diff
@@ -16,60 +16,96 @@
...
 public final class SpecializeBinaryClassNameVisitor implements ClassNameVisitor {
 
     private final ClassNameVisitor delegate;
-    private final TypeName binaryClassName;
+    private final List<TypeDefinition> types;
+    private final ClassName binaryClassName;
+    private final TypeName optionalBinaryTypeName;
 
-    public SpecializeBinaryClassNameVisitor(ClassNameVisitor delegate, TypeName binaryClassName) {
+    public SpecializeBinaryClassNameVisitor(
```
ferozco, line 41: nit: telescope constructors instead of duplicating field initialization

conjure-java-core/src/main/java/com/palantir/conjure/java/util/TypeFunctions.java
```diff
@@ -0,0 +1,208 @@
...
+import com.palantir.conjure.visitor.TypeDefinitionVisitor;
+import com.palantir.conjure.visitor.TypeVisitor;
+import com.palantir.logsafe.SafeArg;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+public final class TypeFunctions {
+
+    public static boolean isAliasType(Type type) {
```
ferozco: this doesn't seem quite right, an reference can refer to any non-primitive type not just an alias

conjure-java-core/src/main/java/com/palantir/conjure/java/util/TypeFunctions.java
```diff
@@ -0,0 +1,208 @@
...
+                return value.getItemType().accept(TypeVisitor.IS_BINARY);
+            }
+        });
+    }
+
+    public static boolean isBinaryOrOptionalBinary(Type type) {
+        return type.accept(TypeVisitor.IS_BINARY) || isOptionalBinary(type);
+    }
+
+    public static boolean isCollectionType(Type type) {
```
ferozco: this method name is a bit misleading since we also refer to Maps (and sometimes optionals) as collections. Can we be a bit more explicit?

conjure-java-core/src/main/java/com/palantir/conjure/java/util/TypeFunctions.java
```diff
@@ -0,0 +1,208 @@
...
+                .get();
+    }
+
+    public static Optional<Type> getAliasedType(
+            com.palantir.conjure.spec.TypeName typeName, Collection<TypeDefinition> typeDefinitions) {
+        // return type definition for the provided alias type
+        TypeDefinition typeDefinition = typeDefinitions.stream()
+                .filter(typeDef -> {
+                    com.palantir.conjure.spec.TypeName currName = typeDef.accept(TypeDefinitionVisitor.TYPE_NAME);
+                    String currClassName = currName.getPackage() + "." + currName.getName();
```
ferozco: can we compare `TypeName` instead of comparing a string constructed from them?

conjure-java-core/src/main/java/com/palantir/conjure/java/util/TypeFunctions.java
```diff
@@ -0,0 +1,208 @@
...
+                .get();
+
+        if (typeDefinition.accept(TypeDefinitionVisitor.IS_ALIAS)) {
+            AliasDefinition aliasDefinition = typeDefinition.accept(TypeDefinitionVisitor.ALIAS);
+            return Optional.of(aliasDefinition.getAlias());
+        }
+        return Optional.empty();
+    }
+
+    private static final ImmutableMap<PrimitiveType.Value, String> PRIMITIVE_TO_TYPE_NAME = new ImmutableMap.Builder<
```
ferozco: nit: can you put static fields at the top of the class?

conjure-java-core/src/main/java/com/palantir/conjure/java/util/TypeFunctions.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    public static boolean isCollectionType(Type type) {
+        return type.accept(TypeVisitor.IS_LIST) || type.accept(TypeVisitor.IS_SET);
+    }
+
+    // Returns the type that the given alias type refers to. For example, if the input type is defined as
+    // "alias: integer", the returned type will be the type for "integer". The provided type must be an alias
+    // (reference) type.
+    public static Type getAliasedType(Type type, List<TypeDefinition> typeDefinitions) {
```
ferozco: we probably want to pass around `Map<TypeName, TypeDefinition>` to avoid having to itterate of the entire set of conjure types over and over again

conjure-java-core/src/main/java/com/palantir/conjure/java/util/TypeFunctions.java
carterkozak: These are moved verbatim from the old `UndertowTypeFunctions` utility since they're not specific to the undertow generator, and I needed some for the general purpose dealias function, I think only the `isBinaryOrOptionalBinary` utility function is new.

conjure-java-core/src/main/java/com/palantir/conjure/java/util/TypeFunctions.java
carterkozak: :+1:

conjure-java-core/src/main/java/com/palantir/conjure/java/util/TypeFunctions.java
carterkozak: :+1:
I think this has been copied around a lot, that should result in the same behavior without string munging

ferozco (approved):

carterkozak:
:robot:

PR #1111, opened 2020-10-09
Pin jackson to 2.11.1

carterkozak:
==COMMIT_MSG==
Pin jackson to 2.11.1
==COMMIT_MSG==


versions.props
```diff
@@ -1,12 +1,12 @@
 ch.qos.logback:* = 1.2.3
 com.atlassian.commonmark:* = 0.12.1
-com.fasterxml.jackson.*:jackson-* = 2.11.3
+com.fasterxml.jackson.*:jackson-* = 2.11.1
```
iamdanfox: thoughts on just moving this into the #upgrader:OFF section? we've done something similar on tracing-java https://github.com/palantir/tracing-java/blob/0ad75535d954dad5a95bafdde6296bff5928f4c5/versions.props#L2-L3

carterkozak: Eh, we've also removed jackson from our managed artifacts

iamdanfox: Fair enough

iamdanfox (approved):

PR #1112, opened 2020-10-12
publish conjure-java jar

rzpt:
## Before this PR

conjure-java was only published as a dist so it could be used as a cli.

## After this PR

==COMMIT_MSG==
Publish conjure-java as a jar so that it can be invoked in process in gradle so it does not unnecessarily consume extra resources.
==COMMIT_MSG==

## Possible downsides?

No downsides

### DIFF
--- a/changelog/@unreleased/pr-1112.v2.yml
+++ b/changelog/@unreleased/pr-1112.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: The `conjure-java` CLI is now published as a jar so that it can be invoked in process by custom gradle plugins. Note that gradle-conjure does not use this approach to ensure that classpaths remain isolated and that the conjure-java generator can be upgraded independently from the gradle-conjure plugin itself.
+  links:
+  - https://github.com/palantir/conjure-java/pull/1112
--- a/conjure-java/build.gradle
+++ b/conjure-java/build.gradle
@@ -15,6 +15,7 @@
  */
 
 apply from: "$rootDir/gradle/publish-dist.gradle"
+apply from: "$rootDir/gradle/publish-jar.gradle"
 
 mainClassName = 'com.palantir.conjure.java.cli.ConjureJavaCli'
 


carterkozak:
@iamdanfox had strong feelings about this, unsure if they're still valid.

Since nobody has implemented the ideas to make standard use faster, and none of those would result in performance as good as running the code directly as a library, I'm inclined to merge this unless there are strong reservations with intent to merge an alternative quickly.

iamdanfox:
(Argh commented on the wrong PR).

Thanks for the tag, BLUF is yes I'd still prefer to not publish jars of generators. I don't like the precedent and I think even in the best-case it's a sub-optimal optimization. Would also need changes to the main conjure CLI and possibly conjure-python too.

Have put together an alternative here: https://github.com/palantir/gradle-conjure/pull/618

rzpt:
So what's the plan?

iamdanfox (approved):
Ok so I haven't been able to get the [isolated in-process classloader approach](https://github.com/palantir/gradle-conjure/pull/618) to work within a reasonable timeframe, and I think the gradle-graal approach will be pretty time consuming to make work in a platform independent way.

Still a bit gutted we couldn't get to an optimization that could benefit both foundry and gotham (because I _really_ don't want gradle-conjure to start using this), but maybe something is better than nothing :shrug:

iamdanfox:
Not sure why changelog-app isn't behaving here - any chance you could push one manually @rzpt?

rzpt:
Manually added a changelog.

iamdanfox:
@ferozco / @carterkozak any last concerns or shall we just pull the trigger?

carterkozak (approved):

carterkozak:
It looks like this breaks resolution of the distribution package, I'm going to temporarily revert so everyone can upgrade the conjure generator.
I'll take a look at publishing in a way that doesn't break distribution resolution,.

ferozco:
I don't think reverting will solve the problem given the artifact has already been produced. Also why did we publish the CLI package? It seems like what you really want is the conjure-java-core.

carterkozak:
> I don't think reverting will solve the problem given the artifact has already been produced

I think it solves resolution searching for a particular version, no?

carterkozak:
revert here: https://github.com/palantir/conjure-java/pull/1179

If the revert wouldn't solve it, wouldn't we expect projects using old conjure-java versions to fail as well?

ferozco:
I guess it will work once everyone upgrades to the latest release

PR #1113, opened 2020-10-12
fix flag counting logic

rzpt:
## Before this PR

An odd number of flags was accepted when only one was allowed.

## After this PR

==COMMIT_MSG==
Fix the flag counting logic so that specifying only one flag would be allowed.
==COMMIT_MSG==

## Possible downsides?

No downsides.


carterkozak (approved):
hah, nice catch!

carterkozak:
Not sure what's up with the robot, I'm going to add the no-changelog label here and merge.

PR #1114, opened 2020-10-12
add environment variable to control formatting

rzpt:
to avoid formatting, set CONJURE_JAVA_NO_FORMAT=true

## Before this PR

There was no way to control whether or not formatting was allowed.

## After this PR

==COMMIT_MSG==
add environment variable to control formatting
==COMMIT_MSG==

## Possible downsides?

No downsides. I'm not sure this is the best way to implement this, but we don't want it to run in CI.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/util/Goethe.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/util/Goethe.java
@@ -49,12 +49,16 @@ public final class Goethe {
             Path outputFile = getFilePath(file, baseDir);
             StringBuilder code = new StringBuilder();
             file.writeTo(code);
-
             CharSink sink = com.google.common.io.Files.asCharSink(outputFile.toFile(), StandardCharsets.UTF_8);
-            try {
-                JAVA_FORMATTER.formatSource(CharSource.wrap(code), sink);
-            } catch (FormatterException e) {
-                throw new RuntimeException(generateMessage(file, code.toString(), e.diagnostics()), e);
+            boolean noFormat = Boolean.parseBoolean(System.getenv("CONJURE_JAVA_NO_FORMAT"));
+            if (noFormat) {
+                sink.write(code);
+            } else {
+                try {
+                    JAVA_FORMATTER.formatSource(CharSource.wrap(code), sink);
+                } catch (FormatterException e) {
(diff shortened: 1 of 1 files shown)


iamdanfox:
Thanks for the tag - indeed I'd really prefer to avoid publishing jars for conjure-java, mainly because I don't want people to suddenly start using these on server backends, considering deep internal classes to be API without our knowledge.  In addition, I want to keep the API between build tools as clear and well defined as possible, as we've previously ended up writing features that can only work for java and are unusable for other languages.  Finally, if we're going to spend time trying to optimize this stuff, I'd really prefer if regular gradle-conjure users can benefit from it, not just gotham.

To that end, I've put together a quick proof-of-concept alternative which I think achieves the same thing: https://github.com/palantir/gradle-conjure/pull/618

iamdanfox:
Hey @rzpt can you flesh out why you consider "no way to control whether or not formatting was allowed" to be bad/broken or worth changing now?

(The intention of the PR template is really to get reviewers and PR authors on the same page, while that sentence is a true statement about the world, it doesn't really help me understand your PR.)

```
<!-- What's wrong with the current state of the world and why change it now? -->
```

Also in general we try to avoid using environment variables for configuration, as they are not very testable or discoverable.

rzpt:
Did you take a look at the document I posted internally? By not formatting, it saves me 12 seconds of build time in the project I work on. The intent of the environment variable is so that it would not accidentally be applied in CI, and it would be applied only in dev environments.

If not an environment variable, what would you suggest? Extra flag to gradle which toggle adding a flag to conjure-java? Wouldn't that flag also be not discoverable?

iamdanfox:
Hmmm 12 second speedup is pretty decent - what does the output actually look like when the formatter is off? is it still readable to humans?

I'd be kinda interested to see what you can get from just parallelizing the formatter too.

Something to consider here is that I think making the CI and local behaviour different would affect build-cache hits on local machines.

rzpt:
The output is still readable. It's mostly whitespace changes it seems.

I thought about parallelizing the formatter. I don't think it is a good idea because when you run a gradle build, each task gets its own thread. If a task is trying to spawn more threads, then that creates more contention.

I am not too familiar with how build caching works. If I happen to not hit the build cache, then I'd get the unformatted code generated locally, but if I do hit the build cache, then I'd download some formatted code instead. This sounds fine to me, but maybe there are other consequences that I'm not aware of.

PR #1140, opened 2020-10-27
Undertow Endpoints provide tags when defined

carterkozak:
==COMMIT_MSG==
Undertow Endpoints provide tags when defined
==COMMIT_MSG==

These can allow us to set up more specific rate limiting based on the definition using simple filters around registration.
Unclear if we would want to do something similar with parameter tags.

### DIFF
--- a/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
+++ b/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
@@ -16,10 +16,13 @@
 
 package com.palantir.conjure.java.undertow.lib;
 
+import com.google.common.collect.ImmutableList;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.palantir.logsafe.Preconditions;
 import io.undertow.server.HttpHandler;
 import io.undertow.util.HttpString;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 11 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
```diff
@@ -111,6 +120,12 @@ public Builder deprecated(Optional<String> value) {
             return this;
         }
 
+        @CanIgnoreReturnValue
+        public Builder deprecated(List<String> value) {
```
ferozco: s/deprecated/tags/

carterkozak: Good catch. Should take an `Iterable` as well!

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
```diff
@@ -111,6 +120,12 @@ public Builder deprecated(Optional<String> value) {
...
+            tags = ImmutableList.copyOf(Preconditions.checkNotNull(value, "tags is required"));
```
carterkozak:
```suggestion
            tags = ImmutableList.copyOf(Preconditions.checkNotNull(value, "tags are required"));
```

carterkozak:
Split into #1144 and #1145

PR #1144, opened 2020-10-28
Support conjure tags in place of cumbersome java annotation-based markers

carterkozak:
==COMMIT_MSG==
Support conjure tags in place of cumbersome java annotation-based markers. The `safe` and `unsafe` tags are supported by the Jersey generator. The Undertow generator passe through the simpler value to the underlying ParamMarker which has internally been updated to support `safe` in addition to `com.palantir.logsafe.Safe`. The undertow `Async` annotation can be replaced by the `server-async` tag.
==COMMIT_MSG==


conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -278,7 +279,10 @@ private ParameterSpec createServiceMethodParameterArg(
                 typeMapper.getClassName(def.getType()), def.getArgName().get());
         if (withAnnotations) {
             getParamTypeAnnotation(def).ifPresent(param::addAnnotation);
-            param.addAnnotations(createMarkers(typeMapper, def.getMarkers()));
+            param.addAnnotations(ImmutableSet.<AnnotationSpec>builder()
+                    .addAll(createMarkers(typeMapper, def.getMarkers()))
```
ferozco, line 283: should we warn/error if a user specifys the tag and the marker?

conjure-java-core/src/test/resources/test/api/AsyncMarkers.java.undertow
```diff
@@ -1,13 +1,19 @@
...
 import javax.annotation.Generated;
 
 @Generated("com.palantir.conjure.java.services.UndertowServiceInterfaceGenerator")
 public interface AsyncMarkers {
     /**
-     * @apiNote {@code GET /async}
+     * @apiNote {@code GET /async/marker}
      */
-    String async();
+    String asyncMarker();
```
ferozco, line 11: shouldn't this also be async?

carterkozak: Nope, this is from a test case without `experimentalUndertowAsyncMarkers`. The `.undertow.async` output reflects `experimentalUndertowAsyncMarkers(true)`. Note that the tag variation does not rely on the experimental markers flag.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
carterkozak: That's probably for the best, otherwise we'd see the value twice in request logs in the bad case.

ferozco (approved):

carterkozak:
:robot:

PR #1145, opened 2020-10-28
Undertow Endpoints provide tags when defined

carterkozak:
## Before this PR
No way to discover tag information while registering endpoints.

==COMMIT_MSG==
Undertow Endpoints provide tags when defined
==COMMIT_MSG==

This opens the door for prototyping work outside of the conjure-java project, for example to allow individual endpoints to opt into rate limiting based on a shared key for all endpoints of a similar type.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -357,13 +357,12 @@ final class UndertowServiceHandlerGenerator {
             CodeBlock arrayValues = CodeBlock.join(
                     Collections2.transform(endpointDefinition.getTags(), value -> CodeBlock.of("$S", value)), ", ");
             endpointBuilder.addField(FieldSpec.builder(
-                            ParameterizedTypeName.get(List.class, String.class),
+                            ParameterizedTypeName.get(ImmutableList.class, String.class),
                             "TAGS",
                             Modifier.PRIVATE,
                             Modifier.STATIC,
                             Modifier.FINAL)
-                    .initializer(CodeBlock.of(
-                            "$T.unmodifiableList($T.asList(" + arrayValues + "))", Collections.class, Arrays.class))
+                    .initializer(CodeBlock.of("$T.of($L)", ImmutableList.class, arrayValues))
                     .build());
             endpointBuilder.addMethod(MethodSpec.methodBuilder("tags")
                     .addModifiers(Modifier.PUBLIC)
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -78,9 +78,7 @@ import io.undertow.util.StatusCodes;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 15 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
```diff
@@ -60,6 +63,11 @@
         return Optional.empty();
     }
 
+    /** Endpoint tags corresponding to the endpoint definition. */
+    default List<String> tags() {
```
carterkozak: I chose List over Set here to make it clear that order is retained from the definition. In many cases I would expect values not to work as simple lookups e.g. `collection.contains(value)`, as there may be a known prefix with a dynamic suffix. Thoughts?

carterkozak:
~~This will result in a semantic merge conflict with #1144, must be rebased prior to merging.~~
rebased.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -349,6 +352,28 @@ private TypeSpec generateEndpointHandler(
         return endpointBuilder.build();
     }
 
+    private static void addTags(EndpointDefinition endpointDefinition, TypeSpec.Builder endpointBuilder) {
+        if (!endpointDefinition.getTags().isEmpty()) {
+            CodeBlock arrayValues = CodeBlock.join(
+                    Collections2.transform(endpointDefinition.getTags(), value -> CodeBlock.of("$S", value)), ", ");
```
carterkozak, line 352: Hmm, we already use guava ListenableFutures in generated code, we might as well use ImmutableList here.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
ferozco: hmm aren't the tags sets into the conjure definition? I think we should keep the uniqueness guarantee

carterkozak: ArgumentDefinition tags are a list, EndpointDefinition tags are a set. imo lists are easier to deal with at runtime, and certainly more concise in memory. Lookups are faster in small lists than small sets.

carterkozak: We provide the uniqueness constraint at the conjure schema level, moving the validation cost from execution time to build time.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
carterkozak: Updated the endpoints list as well, I'm a lot happier with how it looks :-)

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Endpoint.java
ferozco: hmm ok we should consolidate on one at the IR level, thats my bad. I'm more concerned about the information the types give (that you should expect unique values) rather than the memory/lookup cost of lists vs. sets

carterkozak: I've been thinking about these carrying composite data more than constant string values, where the set doesn't provide any help since we already need to handle deduplication by a common component. Either way the collection type shouldn't matter much since I expect it will only be used in setup, not on a per-request basis.

ferozco: hmm, i'm a bit concerned about the composite data. the original design of tags was string because we explicitly wanted to limit the complexity of the information in the tag.

carterkozak: Are we comfortable adding more complex data directly to the IR?

carterkozak: When we discussed the design, we concluded that the single string was sufficient because it could handle both cases. It's possible I missed a subsequent discussion.

ferozco: I think we would need to discuss alternatives like tags being `map<string, string>` or `map<string, any>`. unclear if we want that much complexity in the IR. Happy to hop on a call and discuss

ferozco:
👍

PR #1150, opened 2020-10-30
Apply `@Incubating` annotations based on the `incubating` endpoint tag

carterkozak:
Quick draft of what we discussed, the guava `@Beta` annotation turns invocations yellow in idea based on [UnstableApiUsageInspection.kt#35](https://github.com/JetBrains/intellij-community/blob/5d726bf894d19afe62da7a5b42350e9f021c70fd/jvm/jvm-analysis-impl/src/com/intellij/codeInspection/UnstableApiUsageInspection.kt#L35)

==COMMIT_MSG==
Apply guava `@Beta` annotations based on the `unstable` endpoint tag
==COMMIT_MSG==

Considerations:

1. This implementation uses `unstable`. Would we prefer `incubating`, or something different?
2. This implementation adds the gauva `@Beta` annotation. We can define our own in `conjure-lib` and update our idea configuration to apply it to the UnstableApiUsageInspection list (greater complexity, must work for both idea gradle integration and the idea ipr).

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
@@ -16,8 +16,8 @@
 
 package com.palantir.conjure.java;
 
-import com.google.common.annotations.Beta;
 import com.google.common.collect.ImmutableList;
+import com.palantir.conjure.java.lib.internal.Incubating;
 import com.palantir.conjure.spec.Documentation;
 import com.palantir.conjure.spec.EndpointDefinition;
 import com.squareup.javapoet.AnnotationSpec;
@@ -38,9 +38,9 @@ public final class ConjureAnnotations {
                 : ImmutableList.of();
     }
 
-    public static ImmutableList<AnnotationSpec> unstable(EndpointDefinition definition) {
-        if (definition.getTags().contains("unstable")) {
-            return ImmutableList.of(AnnotationSpec.builder(Beta.class).build());
+    public static ImmutableList<AnnotationSpec> incubating(EndpointDefinition definition) {
+        if (definition.getTags().contains("incubating")) {
+            return ImmutableList.of(AnnotationSpec.builder(Incubating.class).build());
         }
         return ImmutableList.of();
     }
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
@@ -16,8 +16,10 @@
 
 package com.palantir.conjure.java;
 
+import com.google.common.annotations.Beta;
 import com.google.common.collect.ImmutableList;
 import com.palantir.conjure.spec.Documentation;
+import com.palantir.conjure.spec.EndpointDefinition;
 import com.squareup.javapoet.AnnotationSpec;
 import com.squareup.javapoet.ClassName;
 import java.util.Optional;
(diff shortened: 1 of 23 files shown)


conjure-java-core/src/test/resources/test/api/TestService.java.undertow
```diff
@@ -24,6 +25,7 @@ public interface TestService {
      * Returns a mapping from file system id to backing file system configuration.
      * @apiNote {@code GET /catalog/fileSystems}
      */
+    @Beta
```
lycarter: Would it make sense to add a comment to this to explain that this corresponds to "Unstable"? Otherwise, I'd be fine with calling the APIs "Beta". I'm also curious what options we have for typescript, to ensure that they all make sense together.

lycarter (approved):
Overall this looks like exactly what we're looking for. Just a couple questions about how to make sure it ties together well for other generators

conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
```diff
@@ -36,5 +38,12 @@ public static AnnotationSpec getConjureGeneratedAnnotation(Class<?> clazz) {
                 : ImmutableList.of();
     }
 
+    public static ImmutableList<AnnotationSpec> unstable(EndpointDefinition definition) {
+        if (definition.getTags().contains("unstable")) {
```
iamdanfox: - seems like 'incubating' has connotations of a phase you go through at the beginning, but then don't return to after you've become stable.
- `unstable` seems a bit more ambiguous, and could lead to people thinking they should flip endpoints to 'unstable' when they contain bugs?

iamdanfox: even just calling it 'beta' also seems kinda aligned with the 'incubating' connotations

iamdanfox:
Using the existing guava one is definitely neat, but I think it does come with the tradeoff that since so many (well used) guava APIs are already marked 'beta', it might be hard for a project to 'lock in' this check, if they wanted to have a stamp of approval that says "I don't use _any_ beta endpoints"?

A little bit like when people are evaluating a rust crate, they're often interested to know if it contains any 'unsafe', as this might be a source of spicy bugs in the future

conjure-java-core/src/test/resources/test/api/TestService.java.undertow
carterkozak: I'm not sure how this will work in typescript, but I imagine there are similar extension points. I don't think endpoint tags have been implemented in any in conjure-typescript yet.

conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
carterkozak: We can go with `incubating` across the board. All the names in this draft are placeholders :-)

conjure-java-core/src/test/resources/example-service.yml
```diff
@@ -54,7 +54,9 @@ services:
     endpoints:
       getFileSystems:
         markers:
-          - Nonnull # requires FeatureFlags.DangerousGothamMethodMarkers
```
carterkozak, line 57: I've always wanted to order "DangerousGotham" shirts.

conjure-java-core/src/test/resources/test/api/TestService.java.undertow
lycarter: deprecated recently got added as a jsdoc: https://github.com/palantir/conjure-typescript/pull/149

I went through the list of jsdoc tags (https://jsdoc.app/), but didn't see any that looked quite right. I'm not familiar with the ecosystem to know if there are other standards that could be leveraged besides jsdoc, but the support for jsdoc's `@Deprecated` only arrived in vscode in August (https://code.visualstudio.com/updates/v1_49#_deprecated-tag-support-for-javascript-and-typescript), so I don't have a lot of confidence there are good fits available.

carterkozak:
Related: https://github.com/palantir/gradle-baseline/pull/1529

conjure-java-core/src/test/resources/test/api/TestService.java.undertow
carterkozak: Custom vscode plugins might be an option

iamdanfox (approved):

PR #1154, opened 2020-11-06
Fix union deserialization of collections with null or omitted values

carterkozak:
## Before this PR
This is allowed based on the wire specification but without this
fix failed deserialization:
https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#561-coercing-json-null--absent-to-conjure-types

## After this PR
==COMMIT_MSG==
Fix union deserialization of collections with null or omitted values
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSetter;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeName;
@@ -207,7 +208,7 @@ public final class Union {
         private final String value;
 
         @JsonCreator
-        private FooWrapper(@JsonProperty("foo") @Nonnull String value) {
+        private FooWrapper(@JsonSetter("foo") @Nonnull String value) {
             Preconditions.checkNotNull(value, "foo cannot be null");
             this.value = value;
         }
@@ -247,7 +248,7 @@ public final class Union {
         private final int value;
 
         @JsonCreator
-        private BarWrapper(@JsonProperty("bar") @Nonnull int value) {
+        private BarWrapper(@JsonSetter("bar") @Nonnull int value) {
             Preconditions.checkNotNull(value, "bar cannot be null");
             this.value = value;
         }
@@ -288,7 +289,7 @@ public final class Union {
         private final long value;
 
         @JsonCreator
-        private BazWrapper(@JsonProperty("baz") @Nonnull long value) {
+        private BazWrapper(@JsonSetter("baz") @Nonnull long value) {
             Preconditions.checkNotNull(value, "baz cannot be null");
             this.value = value;
         }
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
@@ -23,7 +23,7 @@ import javax.annotation.Nonnull;
 public final class Union {
     private final Base value;
 
-    @JsonCreator
+    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
     private Union(Base value) {
         this.value = value;
     }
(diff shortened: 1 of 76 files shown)


carterkozak:
removed merge-when-ready to ensure we can merge #1155 into this branch before merging+releasing this branch.

conjure-java-core/src/test/resources/example-types.yml
```diff
@@ -178,6 +178,9 @@ types:
           interface: integer
           completed: integer
           unknown: integer
+          optional: optional<string>
+          list: list<string>
+          map: map<string, string>
```
iamdanfox, line 183: could we just re-order this to put the `set` on an adjacent line, so we can see all the container types are in one place?

Also we should probably put these tests in thee conjure-verification server!

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -336,6 +341,54 @@ public void testUnionType_noType() {
...
+    public void testUnionType_nullSetValue() throws IOException {
+        assertThat(mapper.readValue("{\"type\":\"set\",\"set\":null}", UnionTypeExample.class))
+                .isEqualTo(UnionTypeExample.set(ImmutableSet.of()));
+    }
+
+    @Test
+    public void testUnionType_excludedListValue() throws IOException {
+        assertThat(mapper.readValue("{\"type\":\"list\"}", UnionTypeExample.class))
+                .isEqualTo(UnionTypeExample.list(ImmutableList.of()));
+    }
```
iamdanfox, line 372: nice

conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
```diff
@@ -207,7 +208,7 @@ public T visitUnknown(String value) {
         private final String value;
 
         @JsonCreator
-        private FooWrapper(@JsonProperty("foo") @Nonnull String value) {
+        private FooWrapper(@JsonSetter("foo") @Nonnull String value) {
```
iamdanfox, line 211: ~what effect does the JsonProperty -> JsonSetter change have?~ nvm it has the nice `, nulls = Nulls.AS_EMPTY` param

conjure-java-core/src/test/resources/example-types.yml
carterkozak: I'm on my way out the door for a weekend away, would you mind making the reorder change? Otherwise if there are other issues, I can fix on Monday.
This is burning some folks working on config.

conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
carterkozak: Yep, they split the JsonProperty annotation into JsonGetter and JsonSetter for specificity and additional feature support.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -574,6 +579,19 @@ private static TypeSpec generateBase(
...
+        AnnotationSpec.Builder builder = AnnotationSpec.builder(JsonSetter.class)
+                .addMember("value", "$S", field.getFieldName().get());
+        if (isListSetOrMap(field.getType())) {
+            builder.addMember("nulls", "$T.AS_EMPTY", Nulls.class);
+        }
+        return builder.build();
+    }
+
+    private static boolean isListSetOrMap(Type type) {
+        return type.accept(TypeVisitor.IS_LIST) || type.accept(TypeVisitor.IS_SET) || type.accept(TypeVisitor.IS_MAP);
```
iamdanfox, line 590: teeny bit surprising that this includes 3/4 container types, i.e. doesn't include optionals??

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -542,7 +542,7 @@ private static TypeSpec generateBase(
                                     .addAnnotation(AnnotationSpec.builder(JsonCreator.class)
                                             .build())
                                     .addParameter(ParameterSpec.builder(memberType, VALUE_FIELD_NAME)
-                                            .addAnnotation(jsonPropertyAnnotation)
+                                            .addAnnotation(wrapperConstructorParameterAnnotation(memberTypeDef))
```
iamdanfox, line 543: does this JsonSetter approach also work for aliases? like if someone made a union variant which is an Alias of `List<String>` e.g. a Poem where each string is a line?

```yaml
Performance:
  union
    poem: Poem
    dance: Dance
```

iamdanfox:
:+1: because I think this is a strict improvement over the current world.  For completeness, I think we should tackle aliases too, but no need to block this fix on that.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
carterkozak: No need to include optional here because null is already deserialized into an empty optional. I added test coverage for the null and omitted cases in this PR :-)

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
carterkozak: Unfortunately not, that gets a little more complicated. The difficulty is hinting that we want an alias of null rather than a null alias.
In the optional case we generate a default no-arg constructor, we would need to do something similar for the collection types.

carterkozak:
Thanks, Dan!

PR #1155, opened 2020-11-06
Conjure types avoid jackson heuristics and ambiguity around JsonCreator

carterkozak:
## Before this PR
JsonCreator may attempt to use properties rather than delegation.
No reason to do additional heuristic work, and we certainly don't
want unexpected inputs to be deserialized successfully.

## After this PR
==COMMIT_MSG==
Conjure types avoid Jackson heuristics and ambiguity around JsonCreator
==COMMIT_MSG==

## Possible downsides?
Reference an additional jackson class from generated code.


carterkozak:
Ah right, merge-when-ready gets excited around non-develop targets...

ferozco:
@carterkozak we could do the work to deploy label-maker against open source repos to avoid these kinds of issues

PR #1156, opened 2020-11-09
Support alias of wrapper null coercion

carterkozak:
This is similar to an earlier change, but applies to aliases.

## Before this PR
The new WireFormatTests fail without this change.
https://github.com/palantir/conjure-java/compare/ckozak/alias_default_ctor?expand=1#diff-7af5d62c67a59ff3e03f1af5202ec0270b5d9d048b5634e6ae4d87fe715b2266

## After this PR

==COMMIT_MSG==
Support alias of wrapper null coercion
==COMMIT_MSG==

## Possible downsides?
Note that this does allow `{}` to coerce into an empty value, however
it also allows null to work the same between aliases and dealiased
collections. It's more important to correctly support required
parts of the spec than to guard against other edge cases.

Also note that the tests expect null to result in a failure, however
null may be coerced into an empty collection per the conjure spec.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
@@ -238,6 +238,8 @@ public final class AliasGenerator {
 
         @Override
         public Optional<MethodSpec> visitReference(com.palantir.conjure.spec.TypeName value) {
+            // TODO(ckozak): Ideally this would support recursive aliases, for example an alias of alias of list
+            // should have a default constructor.
             return Optional.empty();
         }
 
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
@@ -16,7 +16,6 @@
 
 package com.palantir.conjure.java.types;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.conjure.java.ConjureAnnotations;
 import com.palantir.conjure.java.Options;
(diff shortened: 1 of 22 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -190,13 +190,66 @@ private static boolean isAliasOfDouble(AliasDefinition typeDef) {
...
+            return Optional.of(MethodSpec.constructorBuilder()
+                    .addModifiers(Modifier.PRIVATE)
+                    .addStatement("this($T.emptyMap())", Collections.class)
+                    .build());
+        }
+
+        @Override
+        public Optional<MethodSpec> visitReference(com.palantir.conjure.spec.TypeName value) {
+            return Optional.empty();
+        }
```
iamdanfox, line 244: maybe it's too far-fetched to bother supporting, but in theory should we be recursing here? like if some defines an `Alias` of an `Alias` of a `List<String>`? (like conjure's DealiasingTypeVisitor)

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -577,17 +588,22 @@ private static TypeSpec generateBase(
...
         return builder.build();
     }
 
-    private static boolean isListSetOrMap(Type type) {
-        return type.accept(TypeVisitor.IS_LIST) || type.accept(TypeVisitor.IS_SET) || type.accept(TypeVisitor.IS_MAP);
+    private static boolean supportsNullCoercion(Type type) {
+        return type.accept(TypeVisitor.IS_LIST)
+                || type.accept(TypeVisitor.IS_SET)
+                || type.accept(TypeVisitor.IS_MAP)
+                || type.accept(TypeVisitor.IS_OPTIONAL);
```
iamdanfox: interestingly I think `DefaultableTypeVisitor` implements the same logic (true for all container types, false otherwise)... maybe that class deserves a reename though... `IsContainerTypeVisitor`?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
carterkozak: Ya, I didn't want to bother in this change since it's a much smaller subset, but it might make sense in a follow-up.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
carterkozak: good call, I think `DefaultableTypeVisitor` name makes sense since this is for types with a default value as well.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
iamdanfox: maybe worth dropping a comment or a TODO in there so that people know we're aware this isn't the 100% solution?

iamdanfox (approved):

PR #1166, opened 2020-11-17
Fix alias-of-wrapper null coercion deserialization

carterkozak:
## Before this PR
https://github.com/palantir/conjure-java-runtime/issues/851

## After this PR
"it just works"
==COMMIT_MSG==
Fix alias-of-wrapper null coercion deserialization
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
@@ -28,7 +28,9 @@ import com.palantir.conjure.java.lib.internal.ConjureCollections;
 import com.palantir.conjure.java.types.BeanGenerator.EnrichedField;
 import com.palantir.conjure.java.util.JavaNameSanitizer;
 import com.palantir.conjure.java.util.Javadoc;
+import com.palantir.conjure.java.util.TypeFunctions;
 import com.palantir.conjure.java.visitor.DefaultTypeVisitor;
+import com.palantir.conjure.java.visitor.DefaultableTypeVisitor;
 import com.palantir.conjure.spec.ExternalReference;
 import com.palantir.conjure.spec.FieldDefinition;
 import com.palantir.conjure.spec.FieldName;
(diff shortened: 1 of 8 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -254,6 +262,11 @@ private MethodSpec createSetter(EnrichedField enriched) {
             } else if (options.nonNullCollections()) {
                 annotationBuilder.addMember("contentNulls", "$T.FAIL", Nulls.class);
             }
+        } else if (type.accept(TypeVisitor.IS_REFERENCE)) {
+            Type dealiased = TypeFunctions.toConjureTypeWithoutAliases(type, typesMap);
```
iamdanfox, line 266: YES! This reads really nicely, and matches the pseudo-code that we originally intended in the spec

conjure-java-core/src/test/resources/example-types.yml
```diff
@@ -47,6 +47,9 @@ types:
         fields:
           item: optional<any>
           setItem: optional<set<StringAliasExample>>
+      OptionalAliasExample:
+        fields:
+          optionalAlias: OptionalAlias
```
iamdanfox, line 52: I'm pretty sure the `toConjureTypeWithoutAliases` already does the right thing, but maybe no harm in having a more deeply nested one to prove we're not just unwrapping the top level of aliasing? e.g. `AliasOfAliasOfOptional`?

iamdanfox (approved):
Would be great to include a case in https://github.com/palantir/conjure-verification/blob/develop/master-test-cases.yml#L254 (think it would also need your new type to be defined in https://github.com/palantir/conjure-verification/blob/develop/example-types.conjure.yml).  Then all language generators will be forced to confront this case

conjure-java-core/src/test/resources/example-types.yml
carterkozak: I'm less confident that we already generate empty constructors for nested aliases, there's definitely some room for improvement in that regard. It's a tad more difficult because we must construct an empty instance of the delegate in generated constructor, and multidimensional aliases are somewhat less common.

We could add a `public static MyAlias empty()` factory which returns a singleton. Rather than building all the nested layers of empty constructor arguments we could simply call the delegate, which in turn handles its delegate. Either way I'd prefer to use a separate PR.

ferozco (approved):

PR #1176, opened 2020-11-25
Generate unions compatible with Jackson 2.12

carterkozak:
Jackson does not consider it legal to override JsonTypeInfo
annotations, so we use the existing property feature and add
additional properties to the known union variants.

==COMMIT_MSG==
Generate unions compatible with Jackson 2.12 without attempting to override JsonTypeInfo.
==COMMIT_MSG==


conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -488,6 +488,7 @@ private static TypeSpec generateBase(
                 .addModifiers(Modifier.PRIVATE)
                 .addAnnotation(AnnotationSpec.builder(JsonTypeInfo.class)
                         .addMember("use", "JsonTypeInfo.Id.NAME")
+                        .addMember("include", "JsonTypeInfo.As.EXISTING_PROPERTY")
```
iamdanfox, line 491: Can't say I fully understand the implications of this javadoc 🙈
```
        /**
         * Inclusion mechanism similar to <code>PROPERTY</code> with respect
         * to deserialization; but one that is produced by a "regular" accessible
         * property during serialization. This means that <code>TypeSerializer</code>
         * will do nothing, and expects a property with defined name to be output
         * using some other mechanism (like default POJO property serialization, or
         * custom serializer).
         *<p>
         * Note that this behavior is quite similar to that of using {@link JsonTypeId}
         * annotation;
         * except that here <code>TypeSerializer</code> is basically suppressed;
         * whereas with {@link JsonTypeId}, output of regular property is suppressed.
         * This mostly matters with respect to output order; this choice is the only
         * way to ensure specific placement of type id during serialization.
         * 
         * @since 2.3.0 but databind <b>only since 2.5.0</b>.
         */
        EXISTING_PROPERTY
```

iamdanfox: But i guess if it works it works!

carterkozak: EXISTING_PROPERTY means that _we_ must build objects which produce the `type` field. The UnknownWrapper was already using EXISTING_PROPERTY but Jackson didn't intend to allow subtypes to override JsonTypeInfo from a base type.
Now we apply that to all wrappers, and the known variants provide the type field as a standard property.

carterkozak: Basically, it tells jackson not to write the type identifier because we define it ourselves, otherwise we wouldn't be able to support unknowns.

iamdanfox (approved):

PR #108, opened 2018-11-16
Publish conjure-java-core

uschi2000:

PR #1179, opened 2020-11-26
Revert "publish conjure-java jar (#1112)"

carterkozak:
This reverts commit 25cec998c6c9e0cc250be08746d3f85cb3ed7fae.

## Before this PR
Publishing a jar breaks resolution of the tgz distribiution that
our gradle plugins consume. We'll need to publish differently if
we don't want to break compatibility with the gradle plugin,
or roll out a gradle plugin change separately.

## After this PR

==COMMIT_MSG==
Revert publishing the conjure-java jar to avoid breaking conjure generator distribution resolution with our standard plugins.
==COMMIT_MSG==

## Possible downsides?
Unfortunately this is a step back away from getting good generator performance.

### DIFF
--- a/changelog/@unreleased/pr-1179.v2.yml
+++ b/changelog/@unreleased/pr-1179.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Revert publishing the conjure-java jar to avoid breaking conjure generator
+    distribution resolution with our standard plugins.
+  links:
+  - https://github.com/palantir/conjure-java/pull/1179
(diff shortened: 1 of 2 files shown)


iamdanfox:
I _think_ we might be able to use some gradle attributes on the gradle-conjure side to specify that we want a tgz not a jar

ferozco:
Even if you worked around it within gradle, I think it was wrong to publish the CLI as a JAR. I think the thing you really want is conjure-java-core since it actually houses the generator implementation

ferozco:
👍

iamdanfox:
Huh. Turns out that's already published https://github.com/palantir/conjure-java/blob/6cc280fcbc7fb36cb064c515c4804defae29adcc/conjure-java-core/build.gradle#L17

carterkozak:
via https://github.com/palantir/conjure-java/pull/108

PR #1180, opened 2020-11-26
Drop pjf version to the last java8 compatible release

carterkozak:
## Before this PR
Conjure codegen regressed dropping support for java 8

## After this PR
==COMMIT_MSG==
Fix conjure code generation using a java 8 JRE
==COMMIT_MSG==

## Possible downsides?
We don't get the newest formatting innovations.


iamdanfox (commented):
LGTM

PR #1192, opened 2020-12-08
Move cli to conjure-java-core

rzpt:
## Before this PR

Publishing the jar didn't work. Instead, move the cli logic into conjure-java-core which is published as a jar.

## After this PR

==COMMIT_MSG==
Move cli to conjure-java-core
==COMMIT_MSG==

## Possible downsides?

None.


rzpt:
@carterkozak @ferozco @iamdanfox Here is my new attempt at improving performance. What do you think?

PR #1197, opened 2020-12-10
Improvement: Alias of double can be constructed from BigDecimal

### LINKED ISSUE #1196: alias:double must be generated with an additional internal BigDecimal creator
## What happened?

2.11 added support for BigInteger/BigDecimal which can result in deserialization failure in some cases.
Worth looking into a code fix in jackson if it's possible -- I have to imagine it's something we can do given this requires a few layers of nesting/generics to fail.

Don't fail deserialization.

ferozco:
closes #1196 

## Before this PR
In some cases Jackson would interpret the JSON tokens as a BigDecimal and subsequently fail to initialize a Conjure alias of double since there was no static factory method for Alias of double

## After this PR

==COMMIT_MSG==
Alias of double can be constructed from BigDecimal
==COMMIT_MSG==

## Possible downsides?
This does not address cases where there are nested aliases

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
@@ -43,6 +43,7 @@ import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
 import java.lang.reflect.Method;
+import java.math.BigDecimal;
 import java.util.Collections;
 import java.util.Optional;
 import javax.lang.model.element.Modifier;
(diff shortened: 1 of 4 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -145,6 +146,16 @@ public static JavaFile generateAliasType(TypeMapper typeMapper, AliasDefinition
                     .addCode(intCastCodeBlock)
                     .build());
 
+            spec.addMethod(MethodSpec.methodBuilder("of")
+                    .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
```
carterkozak: We might as well make this private

carterkozak (commented):
Generally lgtm, thanks!

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
ferozco: I followed the pattern of the other factory methods. I'm not sure if making the method private will cause issues for Jackson

carterkozak: Private is supported by Jackson, I don’t think we should make this public if we can help it.

At a certain point we may just want to generate deserializers...

carterkozak (approved):

PR #1211, opened 2021-01-04
work around javapoet deadlock

carterkozak:
See https://github.com/square/javapoet/issues/637

## Before this PR
Deadlock... sometimes.

## After this PR
==COMMIT_MSG==
work around javapoet class loading deadlock between ClassName and TypeName
==COMMIT_MSG==

## Possible downsides?
Code is a little wonky.

### DIFF
--- a/changelog/@unreleased/pr-1211.v2.yml
+++ b/changelog/@unreleased/pr-1211.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: work around javapoet class loading deadlock between ClassName and TypeName
+  links:
+  - https://github.com/palantir/conjure-java/pull/1211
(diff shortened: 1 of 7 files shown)


carterkozak:
Proposed upstream fix here: https://github.com/square/javapoet/pull/821

It would require a public API break though. A while ago I considered writing an errorprone rule to catch these things, I'll revisit that today. It's possible someone else has already written such a rule.

fawind:
Relevant comment with explanation: https://github.com/square/javapoet/issues/637#issuecomment-535304831

fawind (approved):
Thanks for jumping on this! I think it's fine as a workaround to unblock our internal projects until an upstream fix is released.

PR #1223, opened 2021-01-19
Union Visitor builders provide a helper to throw on unknown inputs

carterkozak:
## Before this PR
Previously we found a great deal of variance between how folks
reported unknowns, some using UnsafeArg while others used Safe.
Additionally, this is a common chunk of code that we find duplicated
unnecessarily.

We opted not to implement unknown handling using default methods
on the visitor interface in order to make it an explicit design decision
to throw. The new helper allows developers to make the throwing
behavior obvious without the overhead, complexity, or duplication.

## After this PR
==COMMIT_MSG==
Union visitor builders provide a helper to throw on unknown variants without writing out the throw statement/lambda explicitly.
==COMMIT_MSG==

## Possible downsides?
More generated code.


ferozco:
👍

PR #1230, opened 2021-01-28
Blocking client calls into Clients#callBlocking.

jkozlowski:
## After this PR

==COMMIT_MSG==
Blocking client calls into Clients#callBlocking.
==COMMIT_MSG==


conjure-java-core/src/test/java/com/palantir/conjure/java/TestBase.java
```diff
@@ -52,7 +52,7 @@ protected static void validateGeneratorOutput(List<Path> files, Path outputDir)
     protected static void validateGeneratorOutput(List<Path> files, Path outputDir, String suffix) throws IOException {
         for (Path file : files) {
             Path output = outputDir.resolve(file.getFileName() + suffix);
-            if (Boolean.valueOf(System.getProperty("recreate", "false"))) {
+            if (true) {
```
carterkozak: Don't think you meant to commit this.

Easiest way to trigger this is using `./gradlew test -Drecreate=true`

jkozlowski: I like doing this from IntelliJ, and for some reason it wasn't taking the system property...

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/StaticFactoryMethodType.java
```diff
@@ -0,0 +1,41 @@
...
+
+package com.palantir.conjure.java.services.dialogue;
+
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+
+enum StaticFactoryMethodType {
+    Blocking,
+    Async;
+
+    public void ifBlocking(Runnable ifBlocking) {
```
carterkozak: unused? I might have a file collapsed though

jkozlowski: Indeed, I mostly sent this PR to see if the build is green! and I didn't code review this myself yet.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/StaticFactoryMethodType.java
```diff
@@ -0,0 +1,41 @@
...
+        if (this == Blocking) {
+            ifBlocking.run();
+        }
+    }
+
+    public <R> R switchBy(R blocking, R async) {
```
carterkozak, line 25: May want to take `Supplier<? extends R>` to avoid fully calculating branches unnecessarily.

We could avoid the switch by using an abstract method and implementing the branches on the enum variants (entirely up to you, just typing out my thoughts)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/StaticFactoryMethodType.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.services.dialogue;
+
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+
+enum StaticFactoryMethodType {
+    Blocking,
+    Async;
```
carterkozak: nit `BLOCKING` and `ASYNC`

carterkozak: I'm surprised we don't have static analysis rules for enum constant names.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -65,30 +66,33 @@
...
+            StaticFactoryMethodType methodType) {
         this.options = options;
         this.typeNameResolver = typeNameResolver;
         this.parameterTypes = parameterTypes;
         this.returnTypes = returnTypes;
+        this.methodType = methodType;
     }
 
     @Override
     public MethodSpec generate(ServiceDefinition def) {
```
jkozlowski, line 93: I kinda feel like there's a small refactoring to do with this class and everything, to stop having to pass params into methods, as that makes the code more bulky.

Just make everything into instance variables and simplify. But I don't really feel like doing this now.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/StaticFactoryMethodType.java
jkozlowski: Done

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueServiceGenerator.java
```diff
@@ -69,11 +69,19 @@ public DialogueServiceGenerator(Options options) {
...
+                new ReturnTypeMapper(returnTypes),
+                StaticFactoryMethodType.Async);
 
-        BlockingGenerator blockingGenerator = new BlockingGenerator(
-                options, new ParameterTypeMapper(parameterTypes), new ReturnTypeMapper(returnTypes));
+        StaticFactoryMethodGenerator blockingGenerator = new DefaultStaticFactoryMethodGenerator(
+                options,
+                typeNameResolver,
+                new ParameterTypeMapper(parameterTypes),
+                new ReturnTypeMapper(returnTypes),
```
carterkozak, line 83: We could reuse the same ParameterTypeMapper and ReturnTypeMapper objects, probably makes no difference in practice though

jkozlowski: See my comment above, I think this whole thing should be made so that you instantiate one instance for each class we're generating, to simplify everything.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
jkozlowski: Or rather I don't mind doing this, just not in this PR.

carterkozak:
generally lgtm, thanks!

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/StaticFactoryMethodType.java
jkozlowski: This is for codegeneration, I think I'm going to go for brevity.

carterkozak (approved):

Issue #355, opened 2019-04-20
Consider Staged Builders


carterkozak:
## What happened?

If we neglect to set all required fields in a bean builder, currently it results in failure at runtime.

## What did you want to happen?

We can use staged builders to cause compilation failures in cases that currently only fail at runtime.
Downside of this approach is that existing builders would have to be re-ordered, so it would have to exist behind a feature flag.

markelliot:
I think a more flexible check than staged builders would be something using errorprone, which could also provide for warnings on optionals.

________________________________

carterkozak:
Errorprone has the potential to be more flexible, but the check itself would be complicated. It could allow us to avoid changing any existing correct code, but the downsides are higher compile-time computational cost for larger projects, lack of integration with other tools (IDE), and requiring error-prone integration. Given the option, I would prefer to leverage the compiler as much as possible for problems that the compiler is good at (type checking), and rely on errorprone for things that the compiler cannot do.

Perhaps the cost of changing generated code is too high at this point (it would certainly be painful for the large project I primarily work on) and the benefits of additional safety aren't worth this sort of change, in which case I agree that errorprone is a reasonable compromise.

markelliot:
Right, I think the cost of change here would be very high and for some usages perhaps not possible (consider cases where today a builder is passed as an argument).

An errorprone check may or may not be possible for complex usage, but for straight-forward builder calls should provide a low cost compile-time signal of issues.

________________________________

raiju:
Can be enabled using the `useStagedBuilders` flag

PR #1243, opened 2021-02-15
Feature: Add support for staged builders in object generation

johnhany97:
## Before this PR
Generated beans cannot leverage `stagedBuilders` - Explained in https://immutables.github.io/immutable.html#staged-builder and briefly discussed in https://github.com/palantir/conjure-java/issues/355

## After this PR

==COMMIT_MSG==
Add support for staged builders in object generation
==COMMIT_MSG==

In this implementation, I've made it such that we sort the fields by alphabetical order, and if the field is not a set/list/map/optional, we create a "stage" interface for it leading up to a final stage which contains any set/list/map/optional field(s) along with their auxiliary utils + the `build` method.

This allowed me to keep the `from` method, should the conjure definition of the object only state fields that don't need their own stage.

We can do that since for sets/lists/maps/optionals we have default values that ensure we won't get an `IllegalStateException` if they're not initialized.

## Possible downsides?
- This can't be turned on by default as it'd be considered a breaking change, mostly because of the ordering of the properties.
~- If the generated object has 1+ stages, they lose the `from` helper method.~ No longer the case as long as the `from` method is the first thing used.


carterkozak:
I’ve been thinking about writing this feature for a while, thanks! Big fan of compile-time validation over runtime validation.

I’m out of the office this week, but I’ll plan to review when I’m back next week.

I think we can excavate something out to automatically reorder most builder calls to make migration painless, but it will be a little tricky to do.

conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleOrderedStages.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    public static ItemStageBuilder builder() {
+        return new Builder();
+    }
+
+    public interface ItemStageBuilder {
+        TokenStageBuilder item(@Nonnull String item);
+
+        Completed_StageBuilder from(MultipleOrderedStages other);
```
carterkozak: I don't think this is correct, it doesn't allow any required (non-collection) fields to be modified.

The `Builder` type should be renamed to something like `DefaultBuilder` and made private, while this method returns a `Builder` interface which extends all the stages, allowing all methods to be called.

```java
interface ItemStageBuilder {
    TokenStageBuilder item(@Nonnull String item);
    Builder from(MultipleOrderedStages other);
}

public interface TokenStageBuilder {
    Completed_StageBuilder token(@Nonnull OneField token);
}

interface Completed_StageBuilder {
   MultipleOrderedStages build();
   /*etc*/
}

// new one:
interface Builder extends ItemStageBuilder, TokenStageBuilder, Completed_StageBuilder {
    @Override
    Builder item(@Nonnull String item);
    @Override
    Builder token(@Nonnull OneField token);
    /* etc, all builder methods are overridden to return this Builder type */
}

private static final class DefaultBuilder implements Builder {
   /* implementation */
}
````

Comment (johnhany97) [reply conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleOrderedStages.java]: I like the idea just because i think this looks cleaner.

You'll notice however that in my current implementation, in the final stage, i also provide methods to set any of the required fields, therefore effectively allowing folks that used the `from` method to override any of those fields to whatever they want.

Comment (carterkozak) [reply conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleOrderedStages.java]: Ah, I did not notice that! I think I'd prefer if the final stage didn't expose fields that we've already set in the purely functional workflow since it reduces complexity. What do you think?

Comment (johnhany97) [reply conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleOrderedStages.java]: I agree with that! Let me adjust my implementation to match your example :)

File: conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleOrderedStages.java [hunk 2]
```diff
@@ -0,0 +1,252 @@
...
+        if (fieldValue == null) {
+            if (missingFields == null) {
+                missingFields = new ArrayList<>(4);
+            }
+            missingFields.add(fieldName);
+        }
+        return missingFields;
+    }
+
+    public static ItemStageBuilder builder() {
```
johnhany97, line 119: @carterkozak would recommend taking a look at this example test that demonstrates the implementation we discussed

johnhany97: Actually, let me adjust it such that methods on `Builder` interface all return `Builder` - That way, the consumers of this won't have to worry about following ordering if they've reached the `Builder` position (which can be only be reached from a `from` (he he)

johnhany97: Done in https://github.com/palantir/conjure-java/pull/1243/commits/4c8b0beafff350427e7d47555a6bc6e9bad2b6cb

conjure-java-core/src/integrationInput/java/com/palantir/product/CovariantListExample.java
```diff
@@ -65,8 +65,8 @@ public String toString() {
         return "CovariantListExample{items: " + items + ", externalItems: " + externalItems + '}';
     }
 
-    public static CovariantListExample of(List<Object> items, List<ExampleExternalReference> externalItems) {
-        return builder().items(items).externalItems(externalItems).build();
+    public static CovariantListExample of(List<ExampleExternalReference> externalItems, List<Object> items) {
```
jkozlowski: Is it ok to change this order?

carterkozak: This order shouldn't change, good catch.

johnhany97: Addressed in https://github.com/palantir/conjure-java/pull/1243/commits/15faf55f196ceb1daf6f3051529b7e8bdaab05b6

carterkozak (approved):
Let's see what happens :-)

PR #1255, opened 2021-03-02
Dialogue generators use the new DialogueService annotation

carterkozak:
## Before this PR
Clearer dialogue construction codepath, allows us to consolidate conjure codegen and the annotation processor.

## After this PR

==COMMIT_MSG==
Dialogue generators use the new DialogueService annotation
==COMMIT_MSG==

## Possible downsides?
If codegen updates prior to the dialogue dependency, code won't compile. Updating deps is easy.
If consumers depend on new generated code missing the factory classes at runtime, they should be fine, the Factory class isn't loaded.

### DIFF
--- a/changelog/@unreleased/pr-1255.v2.yml
+++ b/changelog/@unreleased/pr-1255.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Dialogue generators use the new DialogueService annotation
+  links:
+  - https://github.com/palantir/conjure-java/pull/1255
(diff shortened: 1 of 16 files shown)


jkozlowski:
Isn't there some kinda test that the generated code compiles? This repo has a dialogue dependency, why can't we bump this?

carterkozak:
Ya, I think it's fine, `compileJava` may fail until the dialogue dependency is upgraded, but it's pretty darn obvious. If you consume generated dialogue clients from a new generator, you're fine because the `Factory` class isn't initialized until it's referenced.

carterkozak:
Actually, that's also unlikely because this repo provides the cjr (feign shim) to dialogue dependency. We're safe to ship it.

jkozlowski (approved):
Alright I'm happy to try this

PR #1257, opened 2021-03-03
Fix the builder deserializer link, fix static of method builder order

carterkozak:
==COMMIT_MSG==
Fix the builder deserializer link, fix static of method builder order
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleOrderedStages.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleOrderedStages.java
@@ -22,7 +22,7 @@ import java.util.Set;
 import javax.annotation.Generated;
 import javax.annotation.Nonnull;
 
-@JsonDeserialize(builder = MultipleOrderedStages.Builder.class)
+@JsonDeserialize(builder = MultipleOrderedStages.DefaultBuilder.class)
 @Generated("com.palantir.conjure.java.types.BeanGenerator")
 public final class MultipleOrderedStages {
     private final OneField token;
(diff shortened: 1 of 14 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleOrderedStages.java
```diff
@@ -180,7 +180,7 @@ public static ItemStageBuilder builder() {
 
     @Generated("com.palantir.conjure.java.types.BeanBuilderGenerator")
     @JsonIgnoreProperties(ignoreUnknown = true)
-    private static final class DefaultBuilder implements Builder {
+    static final class DefaultBuilder implements Builder {
```
carterkozak, line 183: unfortunately when this is private, ti cannot be referenced bythe `JsonDeserialize` annotation.

conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleOrderedStages.java
```diff
@@ -22,7 +22,7 @@
 import javax.annotation.Generated;
 import javax.annotation.Nonnull;
 
-@JsonDeserialize(builder = MultipleOrderedStages.Builder.class)
+@JsonDeserialize(builder = MultipleOrderedStages.DefaultBuilder.class)
```
carterkozak, line 25: The implementation must be used because it has the annotations. Jackson doesn't care whether it's staged or not and isn't constrained by the nice fluent API.

johnhany97: Ah that's a good find 👍

johnhany97 (approved):
👍

johnhany97:
Looks like `check` CircleCI task is failing for `Cyclomatic Complexity`. We can likely separate a bunch of things out to their own methods to make things simpler.

PR #1273, opened 2021-03-11
Feature: Server request context providing access to some parts of the request

carterkozak:
## Before this PR
No way to read additional request metadata from a resource class. We don't provide the entire HttpServerExchange for a couple reasons: It can be subtly misused, and it can allow the response to be mutated in ways that no longer align with conjure. This is meant to be a safer, albeit limited, alternative.

## After this PR
==COMMIT_MSG==
Server request context providing access to some parts of the request
==COMMIT_MSG==

## Possible downsides?
More knobs, more branching, however this feature is safer than the workarounds we've observed.

### DIFF
--- a/changelog/@unreleased/pr-1273.v2.yml
+++ b/changelog/@unreleased/pr-1273.v2.yml
@@ -1,6 +1,6 @@
 type: feature
 feature:
-  description: conjure-undertow supports an additional `ServerRequestContext` parameter
+  description: conjure-undertow supports an additional `RequestContext` parameter
     providing access to arbitrary request headers and query parameters. This feature is
     enabled on an opt-in basis by adding the `server-request-context` tag to an endpoint.
     Note that this is still considered an experimental feature.
--- a/changelog/@unreleased/pr-1273.v2.yml
+++ b/changelog/@unreleased/pr-1273.v2.yml
@@ -1,0 +1,5 @@
+type: feature
+feature:
+  description: Server request context providing access to some parts of the request
+  links:
+  - https://github.com/palantir/conjure-java/pull/1273
--- a/changelog/@unreleased/pr-1273.v2.yml
+++ b/changelog/@unreleased/pr-1273.v2.yml
@@ -1,5 +1,8 @@
 type: feature
 feature:
-  description: Server request context providing access to some parts of the request
+  description: conjure-undertow supports an additional `ServerRequestContext` parameter
+    providing access to arbitrary request headers and query parameters. This feature is
+    enabled on an opt-in basis by adding the `server-request-context` tag to an endpoint.
+    Note that this is still considered an experimental feature.
   links:
   - https://github.com/palantir/conjure-java/pull/1273
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/NameCollisionServiceEndpoints.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/NameCollisionServiceEndpoints.java
@@ -1,6 +1,7 @@
 package com.palantir.product;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.palantir.conjure.java.undertow.lib.Deserializer;
 import com.palantir.conjure.java.undertow.lib.Endpoint;
 import com.palantir.conjure.java.undertow.lib.Serializer;
@@ -14,10 +15,12 @@ import io.undertow.util.HeaderMap;
 import io.undertow.util.HttpString;
 import io.undertow.util.Methods;
 import io.undertow.util.PathTemplateMatch;
+import io.undertow.util.StatusCodes;
 import java.io.IOException;
 import java.util.Deque;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.annotation.Generated;
 
 @Generated("com.palantir.conjure.java.services.UndertowServiceHandlerGenerator")
@@ -34,7 +37,10 @@ public final class NameCollisionServiceEndpoints implements UndertowService {
 
     @Override
     public List<Endpoint> endpoints(UndertowRuntime runtime) {
-        return ImmutableList.of(new IntEndpoint(runtime, delegate));
+        return ImmutableList.of(
+                new IntEndpoint(runtime, delegate),
+                new NoContextEndpoint(runtime, delegate),
+                new ContextEndpoint(runtime, delegate));
     }
 
     private static final class IntEndpoint implements HttpHandler, Endpoint {
@@ -97,6 +103,105 @@ public final class NameCollisionServiceEndpoints implements UndertowService {
         @Override
         public String name() {
             return "int";
+        }
+
+        @Override
+        public HttpHandler handler() {
+            return this;
+        }
+    }
+
+    private static final class NoContextEndpoint implements HttpHandler, Endpoint {
+        private final UndertowRuntime runtime;
+
+        private final UndertowNameCollisionService delegate;
+
+        private final Deserializer<String> deserializer;
+
+        NoContextEndpoint(UndertowRuntime runtime, UndertowNameCollisionService delegate) {
+            this.runtime = runtime;
+            this.delegate = delegate;
+            this.deserializer = runtime.bodySerDe().deserializer(new TypeMarker<String>() {});
+        }
+
+        @Override
+        public void handleRequest(HttpServerExchange exchange) throws IOException {
+            String requestContext = deserializer.deserialize(exchange);
+            delegate.noContext(requestContext);
+            exchange.setStatusCode(StatusCodes.NO_CONTENT);
+        }
+
+        @Override
+        public HttpString method() {
+            return Methods.POST;
+        }
+
+        @Override
+        public String template() {
+            return "/no/context";
+        }
+
+        @Override
+        public String serviceName() {
+            return "NameCollisionService";
+        }
+
+        @Override
+        public String name() {
+            return "noContext";
+        }
+
+        @Override
+        public HttpHandler handler() {
+            return this;
+        }
+    }
+
+    private static final class ContextEndpoint implements HttpHandler, Endpoint {
+        private static final ImmutableSet<String> TAGS = ImmutableSet.of("server-request-context");
+
+        private final UndertowRuntime runtime;
+
+        private final UndertowNameCollisionService delegate;
+
+        private final Deserializer<String> deserializer;
+
+        ContextEndpoint(UndertowRuntime runtime, UndertowNameCollisionService delegate) {
+            this.runtime = runtime;
+            this.delegate = delegate;
+            this.deserializer = runtime.bodySerDe().deserializer(new TypeMarker<String>() {});
+        }
+
+        @Override
+        public Set<String> tags() {
+            return TAGS;
+        }
+
+        @Override
+        public void handleRequest(HttpServerExchange exchange) throws IOException {
+            String requestContext_ = deserializer.deserialize(exchange);
+            delegate.context(requestContext_, runtime.contexts().createContext(exchange, this));
+            exchange.setStatusCode(StatusCodes.NO_CONTENT);
+        }
+
+        @Override
+        public HttpString method() {
+            return Methods.POST;
+        }
+
+        @Override
+        public String template() {
+            return "/context";
+        }
+
+        @Override
+        public String serviceName() {
+            return "NameCollisionService";
+        }
+
+        @Override
+        public String name() {
+            return "context";
         }
 
         @Override
--- a/conjure-java-core/src/test/resources/dangerous-name-service.yml
+++ b/conjure-java-core/src/test/resources/dangerous-name-service.yml
@@ -52,3 +52,12 @@ services:
             markers:
               - Safe
         returns: string
+      noContext:
+        http: POST /no/context
+        args:
+          requestContext: string
+      context:
+        http: POST /context
+        args:
+          requestContext: string
+        tags: [server-request-context]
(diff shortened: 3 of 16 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServerRequestContext.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.palantir.conjure.java.undertow.lib;
+
+import com.google.common.annotations.Beta;
+import com.google.common.collect.ListMultimap;
+import java.util.List;
+import java.util.Optional;
+
+@Beta
+public interface ServerRequestContext {
```
carterkozak: Alternatively: could drop the `Server` component of the name, the `com.palantir.conjure.java.undertow.lib` package name implies that to some extent.

needs class-level javadoc

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServerRequestContext.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Returns the value of the first header named {@code headerName}. The name is case insensitive.
+     * An {@link Optional#empty()} is returned if no such header exists.
+     */
+    Optional<String> firstHeader(String headerName);
+
+    /**
+     * Returns all query parameters associated with the current request.
+     */
+    ListMultimap<String, String> queryParameters();
```
carterkozak: This differs from the header accessors, I expect there to be fewer query parameters, so this should be reasonable, where we headers will be queried more frequently. Unlike headers, there's a use case for iterating through all query parameter names.

Note: this api uses guava ListMultimap. The async feature already uses guava ListenableFuture, and the type is cleaner than `Map<String,List<String>>` so I think it's okay.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServerContexts.java
```diff
@@ -0,0 +1,27 @@
...
+package com.palantir.conjure.java.undertow.lib;
+
+import com.google.common.annotations.Beta;
+import io.undertow.server.HttpServerExchange;
+
+@Beta
+public interface ServerContexts {
+
+    /** Returns a new {@link ServerRequestContext} which describes the incoming request. */
+    ServerRequestContext createContext(HttpServerExchange exchange, Endpoint endpoint);
```
carterkozak: The `Endpoint` parameter isn't used yet, but it's aligned with the request-context, and we wouldn't want to allow contexts to be created without a known endpoint.

conjure-java-core/src/test/resources/dangerous-name-service.yml
```diff
@@ -52,3 +52,12 @@ services:
...
               - Safe
         returns: string
+      noContext:
+        http: POST /no/context
+        args:
+          requestContext: string
+      context:
+        http: POST /context
+        args:
+          requestContext: string
```
CRogers, line 62: This is just to test that we can generate the code properly with a request context, right?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServerRequestContext.java
CRogers: I think I prefer `RequestContext` rather than `ServerRequestContext`. The latter makes me think the lifetime of this object is attached to the the lifetime of the server rather than the request.

conjure-java-core/src/test/resources/dangerous-name-service.yml
```diff
@@ -52,3 +52,12 @@ services:
...
+        tags: [server-request-context]
```
CRogers, line 63: I like people have to add this to get it.

CRogers: I guess we need it for backcompat.

conjure-java-core/src/test/resources/dangerous-name-service.yml
carterkozak: That's correct, we're testing that an existing parameter with the expected `ServerRequestContext` parameter name compiles properly.

The endpoint above tests that when you have an existing variable named `requestContext` which would clash with the special parameter, it's _not_ renamed unnecessarily when there's no `ServerRequestContext` parameter.

conjure-java-core/src/test/resources/dangerous-name-service.yml
carterkozak: Most endpoints shouldn't need the context, so I don't want to generate unnecessary inputs that must be _ignored. When it is needed, it's not necessarily required for every endpoint, so I prefer the tag over a generator option.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/ServerRequestContext.java
carterkozak: Good idea.

jkozlowski:
can you link to some of the workarounds you've observed?

changelog/@unreleased/pr-1273.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: feature
+feature:
+  description: conjure-undertow supports an additional `ServerRequestContext` parameter
```
jkozlowski: ServerRequestContext -> RequestContext

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Returns the value of the first header named {@code headerName}. The name is case insensitive.
+     * An {@link Optional#empty()} is returned if no such header exists.
+     */
+    Optional<String> firstHeader(String headerName);
+
+    /**
+     * Returns all query parameters associated with the current request.
+     */
+    ListMultimap<String, String> queryParameters();
```
jkozlowski, line 46: Why not just Multimap? The underlying is a Map<String, Dequeue<String>>, doesn't feel like you gain anything from list and you'd be able to maybe construct it more cheaply?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public RequestContext createContext(HttpServerExchange exchange, Endpoint _endpoint) {
+        return new ConjureServerRequestContext(exchange);
+    }
+
+    private static final class ConjureServerRequestContext implements RequestContext {
+
+        private final HttpServerExchange exchange;
+
+        private ImmutableListMultimap<String, String> cachedQueryParams;
```
jkozlowski, line 44: Do we not need things to be threadsafe in these contexts? I know it's just a cache, so probs doesn't matter if you instantiate it twice

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/UndertowRuntime.java
```diff
@@ -41,4 +43,8 @@
 
     /** Provides the {@link ExceptionHandler} used to map {@link Throwable failures} to Conjure responses. */
     ExceptionHandler exceptionHandler();
+
+    /** Provides the {@link Contexts} which safely expose request details to resources. */
+    @Beta
+    Contexts contexts();
```
jkozlowski, line 49: Why this layer of indirection? Why can't you return RequestContext directly, but only create it on first call, to maintain laziness?

jkozlowski: Ok, I understand: genned code impls HttpHandler. UndertowRuntime is a per-server factory for stuff that the codegen code needs.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Contexts.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Factory for {@link RequestContext} objects when the {@code server-request-context} tag is used.
+ * This is an internal interface that should only be used by generated code, it may change without
+ * warning, and guarantees are only made for generated code.
+ */
+@Beta
+public interface Contexts {
+
+    /** Returns a new {@link RequestContext} which describes the incoming request. */
+    RequestContext createContext(HttpServerExchange exchange, Endpoint endpoint);
```
jkozlowski, line 31: what usecase do you envision for knowing the endpoint param?

conjure-java-core/src/integrationInput/java/com/palantir/product/NameCollisionServiceEndpoints.java
```diff
@@ -104,4 +110,103 @@ public HttpHandler handler() {
...
+
+        ContextEndpoint(UndertowRuntime runtime, UndertowNameCollisionService delegate) {
+            this.runtime = runtime;
+            this.delegate = delegate;
+            this.deserializer = runtime.bodySerDe().deserializer(new TypeMarker<String>() {});
+        }
+
+        @Override
+        public Set<String> tags() {
+            return TAGS;
```
jkozlowski, line 177: Out of curiosity, why are tags available at runtime?

jkozlowski (commented):
I think just some comments

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Contexts.java
jkozlowski: Also what other things you'd like to eventually expose in RequestContext?

carterkozak:
> can you link to some of the workarounds you've observed?

Provided an internal quip link.

conjure-java-core/src/integrationInput/java/com/palantir/product/NameCollisionServiceEndpoints.java
carterkozak: They're incredibly useful at runtime to opt into custom HttpHandler wrappers, we used them with a couple products to apply a wrapper that opts into less desirable behavior to match old Jersey bugs without forcing it upon new endpoints.

The SLO and limits features we've discussed would likely leverage tags initially as well -- they're a great staging ground for experimental features without adding new blocks to the conjure lang, IR, and reference implementations.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Contexts.java
carterkozak: > what usecase do you envision for knowing the endpoint param?

Endpoint service/name metadata, possibly additional path data. We may add additional data (like tags) to the endpoint object in the future that we need expose through RequestContext.

> Also what other things you'd like to eventually expose in RequestContext?

I'd like to add the ability to mark values safe-loggable in code, but I'm planning to implement that separately. Alta has some thread-local hacks in place to do this, but I'd much rather make it possible using the RequestContext.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
carterkozak: Right, this is lazily initialized. If it's created multiple times, that's fine.

In general we expect requests to be single threaded, that is to say zero to one threads "own" the request at any given point in time, although they may have dispatched N threads to do work, those threads aren't interacting directly with the request.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
carterkozak: I'm not sure there's a multimap wrapper type we could use, I'd rather use a strongly typed `ListMultimap` than `Multimap` to avoid forcing consumers to interact with `Collection`. In this case I'm not concerned with performance because the immutable maps is only created when it's requested, which should be rare. Definitely want to avoid `SetMultimap` because we don't want to hash values, or build trees.

changelog/@unreleased/pr-1273.v2.yml
carterkozak: Good catch!
```suggestion
  description: conjure-undertow supports an additional `RequestContext` parameter
```

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Contexts.java
jkozlowski: Yes I wrote those hacks.

jkozlowski (approved):

carterkozak:
:+1:

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
pkoenig10: The [Guava documentation](https://github.com/google/guava/blob/v30.1/guava/src/com/google/common/collect/Multimap.java#L104-L107) actually discourages the direct use of `Multimap`:
> Instead of using the Multimap interface directly, prefer the subinterfaces ListMultimap and SetMultimap

PR #1274, opened 2021-03-17
Generated stated builders retain original order for required fields

carterkozak:
==COMMIT_MSG==
Generated stated builders retain original order for required fields
==COMMIT_MSG==

Due to readability concerns around the staged builder code when the fields are defined in a particular order. For example it's common to have a `min` and `max` field defined in that order, however alphabetically they're reversed.

It's uncommon for fields to be reordered, and in most cases we don't require abi compatibility between api producers and consumers.


tetigi (approved):
LGTM!

carterkozak:
@tetigi Looks like we still need to give you review perms!

PR #1275, opened 2021-03-17
Support checked exceptions in Undertow and Dialogue.

tetigi:
**This PR is step 3. Step 1 is here: https://github.com/palantir/conjure/pull/816**

**TODO**
- Dialogue implementation.

## Before this PR

Only runtime exceptions are supported when using the Conjure error types. This PR adds support for:

1) declared errors on endpoints as checked exceptions
2) disallowing endpoint-declared errors from standard error factories
## After this PR

==COMMIT_MSG==
Errors declared on endpoints are now exposed in Undertow and Dialogue interfaces as checked exceptions. 
==COMMIT_MSG==


build.gradle
```diff
@@ -48,6 +48,7 @@ allprojects {
     group 'com.palantir.conjure.java'
 
     repositories {
+        mavenLocal()
```
tetigi, line 51: Will be removed once other PRs have been merged.

conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureTags.java
```diff
@@ -46,7 +46,7 @@ public static boolean isUnsafe(Collection<String> tags) {
 
     public static ImmutableList<AnnotationSpec> tagAnnotations(ArgumentDefinition argument) {
         validateTags(argument);
-        List<String> tags = argument.getTags();
```
tetigi, line 49: I think while using the local version of `conjure-api` I picked up another API change where tags became sets instead of lists. Guess this should be resolved as the other PRs resolve.

carterkozak:
Thoughts on naming the exceptions `<name>ServiceException` and `<name>RemoteException` to differentiate client and server failures at a glance?

conjure-java-core/src/integrationInput/java/com/palantir/product/ConjureErrors.java
```diff
@@ -78,4 +95,24 @@ public static void throwIfInvalidTypeDefinition(boolean shouldThrow, String type
...
+            throw shouldBeChecked();
+        }
+    }
+
+    public static final class ShouldBeCheckedException extends CheckedServiceException {
+        ShouldBeCheckedException(ErrorType errorType, Arg<?>... parameters) {
+            super(errorType, parameters);
+        }
+
+        ShouldBeCheckedException(ErrorType errorType, @Nullable Throwable cause, Arg<?>... parameters) {
```
carterkozak: Can we use private constructors?

conjure-java-core/src/integrationInput/java/com/palantir/product/ConjureErrors.java
```diff
@@ -55,6 +64,14 @@ public static ServiceException invalidTypeDefinition(Throwable cause, String typ
                 INVALID_TYPE_DEFINITION, cause, SafeArg.of("typeName", typeName), UnsafeArg.of("typeDef", typeDef));
     }
 
+    public static ShouldBeCheckedException shouldBeChecked() {
+        return new ShouldBeCheckedException(SHOULD_BE_CHECKED);
+    }
+
+    public static ShouldBeCheckedException shouldBeChecked(Throwable cause) {
```
carterkozak: Not directly related to this PR, but we should update the codegen to annotate these methods with `@CheckReturnValue` so it's obvious calling `shouldBeChecked()` without a `throw` keyword doesn't actually throw.

conjure-java-core/src/test/resources/ete-service.yml
```diff
@@ -235,4 +240,9 @@ services:
       receiveSetOfOptionals:
         http: PUT /set/optionals
         args:
-          value: set<optional<string>>
\ No newline at end of file
+          value: set<optional<string>>
+
+      throwsCheckedException:
```
carterkozak, line 245: Let's add a test to verify checked exceptions work as expected. I think we need to update the conjure-undertow ConjureExceptions function to support the checked abstract exception types.

conjure-java-core/src/integrationInput/java/com/palantir/product/ConjureErrors.java
tetigi: Shall I just tag this onto this PR, or open a separate one?

tetigi:
Got a sketch ready for the `RemoteCheckedException` parts with a few of the adjustments mentioned above. Just need to write the generators for it.

PR #1295, opened 2021-04-01
RequestContext supports dynamic parameter markers

carterkozak:
This can be leveraged to add safe-loggable parameters to the
request log.

## Before this PR
ThreadLocals to add request logging metadata... oof.

## After this PR

==COMMIT_MSG==
RequestContext supports dynamic parameter markers
==COMMIT_MSG==

## Possible downsides?
More API surface. Not much, and beats the alternative.

PR #1296, opened 2021-04-01
RequestContext supports adding safe-logging Args to the request

carterkozak:
==COMMIT_MSG==
RequestContext supports adding safe-logging Args to the request
==COMMIT_MSG==

Alternative to #1295

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
@@ -31,28 +31,27 @@ import java.util.Map;
 import java.util.Optional;
 
 final class ConjureContexts implements Contexts {
-    private final RequestLogParameterHandler requestLogParameterHandler;
+    private final RequestArgHandler requestArgHandler;
 
-    ConjureContexts(RequestLogParameterHandler requestLogParameterHandler) {
-        this.requestLogParameterHandler = requestLogParameterHandler;
+    ConjureContexts(RequestArgHandler requestArgHandler) {
+        this.requestArgHandler = requestArgHandler;
     }
 
     @Override
     public RequestContext createContext(HttpServerExchange exchange, Endpoint _endpoint) {
-        return new ConjureServerRequestContext(exchange, requestLogParameterHandler);
+        return new ConjureServerRequestContext(exchange, requestArgHandler);
     }
 
     private static final class ConjureServerRequestContext implements RequestContext {
 
         private final HttpServerExchange exchange;
-        private final RequestLogParameterHandler requestLogParameterHandler;
+        private final RequestArgHandler requestArgHandler;
 
         private ImmutableListMultimap<String, String> cachedQueryParams;
 
-        ConjureServerRequestContext(
-                HttpServerExchange exchange, RequestLogParameterHandler requestLogParameterHandler) {
+        ConjureServerRequestContext(HttpServerExchange exchange, RequestArgHandler requestArgHandler) {
             this.exchange = exchange;
-            this.requestLogParameterHandler = requestLogParameterHandler;
+            this.requestArgHandler = requestArgHandler;
         }
 
         @Override
@@ -77,8 +76,8 @@ final class ConjureContexts implements Contexts {
         }
 
         @Override
-        public void requestLogParameter(Arg<?> arg) {
-            requestLogParameterHandler.addParameter(exchange, arg);
+        public void requestArg(Arg<?> arg) {
+            requestArgHandler.arg(exchange, arg);
         }
 
         private ImmutableListMultimap<String, String> buildQueryParameters() {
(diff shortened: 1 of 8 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
```diff
@@ -68,6 +76,11 @@ public RequestContext createContext(HttpServerExchange exchange, Endpoint _endpo
             return cachedQueryParamsSnapshot;
         }
 
+        @Override
+        public void requestLogParameter(Arg<?> arg) {
```
carterkozak: Ideas for a better method name?

jkozlowski: Yeah, was going to say: I think we discussed that I personally wouldn't be opposed if those actually showed up in normal logs as MDC as well, to aid debugging. So maybe #requestParameter is more generic (because it doesn't tie you to request LOGs)? We kinda used RequestContext already to name the whole concept here, so context is probably out of question.

Other ideas: #diagnosticParameter, but not too keen.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -39,6 +39,7 @@
     private final MarkerCallback markerCallback;
```
jkozlowski, line 39: Does this now get superceeded?

jkozlowski, line 39: Does this now get deprecated/replaced?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
```diff
@@ -44,4 +45,10 @@
      * Returns all query parameters associated with the current request.
      */
     ListMultimap<String, String> queryParameters();
+
+    /**
+     * Associates an {@link Arg} with the current request for observability.
+     * Implementations may choose to include this data in the request log.
+     */
+    void requestLogParameter(Arg<?> arg);
```
jkozlowski: Curious what we do when these are null? Defo not throw right?

jkozlowski (commented):
Name suggestion

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
carterkozak: `requestParameter` is nice, however it implies that the input is a parameter of the request, when it might come from elsewhere. I think the name you propose is easier to understand at a glance though.

carterkozak: Maybe `requestArg(Arg<?>)` based on the `Arg` type name, avoiding naming collision with request parameters entirely, which I took from `requestLogV2.params`?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
carterkozak: Yep, this is where things get a little funny though: I'd like the contract to be defined as non-null because the call isn't helpful in that case, however given an unexpected null input I'd expect the value to be ignored. I have a pretty strong stance against observability frameworks throwing because they're most utilized when assumptions have already been broken.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
carterkozak: Markers are still used for things defined in the conjure yaml via codegen, the RequestContext on the other hand is used directly by developers.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
jkozlowski: yeah, log + metric on misuse is the way to go.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
jkozlowski: yeah, I guess I'm asking if it makes sense to keep both or we should just have one: the Safe annotation in conjure yaml is basically a shortcut, when you know the whole thing is safe.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
```diff
@@ -44,4 +45,10 @@
      * Returns all query parameters associated with the current request.
      */
     ListMultimap<String, String> queryParameters();
+
+    /**
+     * Associates an {@link Arg} with the current request for observability.
+     * Implementations may choose to include this data in the request log.
```
jkozlowski, line 51: "Implementation may choose to use this data for request logging and other observability needs"?

jkozlowski (approved):

PR #1298, opened 2021-04-03
fix #1297: Include safety annotations on undertow and dialogue interfaces

### LINKED ISSUE #1297: Safe tag on argument definition does not render into Java annotation
## What happened?

I added the `safe` tag to an endpoint's argument definition and the service rendered successfully, but none of the generated Java classes actually include the `@Safe` annotation.

The rendered Java classes should include the `@Safe` annotation for the tagged arguments.

carterkozak:
These annotations are helpful both for developers consuming code,
and the static analysis which protects them from foot-guns. Data
leaks are no fun for anyone involved, and the more we can do
to make observability safety classifications obvious the faster
and more safely we'll be able to iterate.

## Before this PR

#1297

## After this PR
==COMMIT_MSG==
Include safety annotations on undertow and dialogue interfaces
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureTags.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureTags.java
@@ -44,7 +44,7 @@ public final class ConjureTags {
         return tags.contains(UNSAFE);
     }
 
-    public static ImmutableList<AnnotationSpec> tagAnnotations(ArgumentDefinition argument) {
+    public static ImmutableList<AnnotationSpec> safetyAnnotations(ArgumentDefinition argument) {
         validateTags(argument);
         List<String> tags = argument.getTags();
         ImmutableList.Builder<AnnotationSpec> builder = ImmutableList.builderWithExpectedSize(1);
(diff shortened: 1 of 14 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureTags.java
```diff
@@ -44,7 +44,7 @@ public static boolean isUnsafe(Collection<String> tags) {
         return tags.contains(UNSAFE);
     }
 
-    public static ImmutableList<AnnotationSpec> tagAnnotations(ArgumentDefinition argument) {
+    public static ImmutableList<AnnotationSpec> safetyAnnotations(ArgumentDefinition argument) {
```
jkozlowski, line 47: I assume dependency on the Safe class is automatically inherited in generated code?

jkozlowski (commented):
Quick qq, can approve afterwards

conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureTags.java
carterkozak: That's correct -- generated code already uses safe-logging preconditions which provides the safe-logging annotation API as an api dependency.

jkozlowski (approved):

carterkozak:
:robot: :popcorn:

PR #1300, opened 2021-04-06
Avoid sending empty optional fields

carterkozak:
These fields are excluded entirely per the conjure spec:
https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#561-coercing-json-null--absent-to-conjure-types

## Before this PR
Always send `{"filedName":null}`

## After this PR
Becomes `{}`.
==COMMIT_MSG==
Avoid sending empty optional values within generated objects.
==COMMIT_MSG==

## Possible downsides?
Risks:
Typescript/javascript clients will _receive_ `undefined` values
rather than `null`. Consumers which use `==` will not be impacted,
however those using `===` (and friends) _may break_.
I have verified that we have static analysis enabled to enforce the correct use of `==`,
so the risk only applies to cases where users have explicitly disabled the check, or
'splat' objects.

No risk for non-conjure objectmapper uses becuase we explicitly opt in with generated conjure objects.

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
@@ -43,6 +43,7 @@ import com.palantir.product.MapExample;
 import com.palantir.product.OptionalAlias;
 import com.palantir.product.OptionalAliasExample;
 import com.palantir.product.OptionalExample;
+import com.palantir.product.PrimitiveOptionalsExample;
 import com.palantir.product.SetAlias;
 import com.palantir.product.SetExample;
 import com.palantir.product.StringAliasExample;
@@ -559,6 +560,28 @@ public final class WireFormatTests {
     public void testOptionalAliasExample() throws IOException {
         assertThat(mapper.readValue("{\"optionalAlias\":null}", OptionalAliasExample.class))
                 .isEqualTo(OptionalAliasExample.of(OptionalAlias.of(Optional.empty())));
+    }
+
+    @Test
+    public void testEmptyOptionalFieldSerialization() throws IOException {
+        assertThat(mapper.writeValueAsString(PrimitiveOptionalsExample.builder().build()))
+                .isEqualTo("{}");
+    }
+
+    @Test
+    public void testEmptyOptionalFieldSerializationOfAlias() throws IOException {
+        assertThat(mapper.writeValueAsString(OptionalAliasExample.builder()
+                        .optionalAlias(OptionalAlias.of(Optional.empty()))
+                        .build()))
+                .isEqualTo("{}");
+    }
+
+    @Test
+    public void testEmptyOptionalFieldSerializationOfAlias_emptyString() throws IOException {
+        assertThat(mapper.writeValueAsString(OptionalAliasExample.builder()
+                        .optionalAlias(OptionalAlias.of(Optional.of("")))
+                        .build()))
+                .isEqualTo("{\"optionalAlias\":\"\"}");
     }
 
     private static final class TestVisitor implements UnionTypeExample.Visitor<Integer> {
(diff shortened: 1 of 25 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -561,6 +567,52 @@ public void testOptionalAliasExample() throws IOException {
                 .isEqualTo(OptionalAliasExample.of(OptionalAlias.of(Optional.empty())));
     }
 
+    @Test
+    public void testEmptyOptionalFieldSerialization() throws IOException {
+        assertThat(mapper.writeValueAsString(PrimitiveOptionalsExample.builder()
+                        .aliasTwo(StringAliasTwo.of(Optional.empty()))
+                        .aliasOptional(OptionalAlias.of(Optional.empty()))
+                        .aliasOptionalMap(OptionalMapAliasExample.of(Optional.empty()))
```
jkozlowski, line 578: maybe we want tests for set/list as well for completeness?

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -561,6 +567,52 @@ public void testOptionalAliasExample() throws IOException {
...
+                .set(ImmutableSet.of())
+                .aliasOne(StringAliasOne.of(""))
+                .aliasTwo(StringAliasTwo.of(Optional.of(StringAliasOne.of(""))))
+                .aliasOptional(OptionalAlias.of(Optional.of("")))
+                .aliasOptionalMap(OptionalMapAliasExample.of(Optional.of(ImmutableMap.of())))
+                .aliasMap(MapAliasExample.of(ImmutableMap.of()))
+                .build();
+        PrimitiveOptionalsExample recreated =
+                mapper.readValue(mapper.writeValueAsBytes(original), PrimitiveOptionalsExample.class);
+        assertThat(recreated).isEqualTo(original);
```
jkozlowski: For documentation purposes really, can you also add a test that shows the actual string serialized output, like in the above test?

jkozlowski (approved):
Small questions

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
carterkozak: That sort of test can be difficult to maintain because we don't provide any guarantees about field order, which doesn't impact tests using zero or one fields.
How about adding the serialized json string to the failure message if the test doesn't pass, and verifying that the expected number of fields are present?

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
carterkozak: Updated :+1:

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
jkozlowski: Alternatively a test per field, that way you don't have any ordering issues, but that sounds like a lot of faff. Should we be specifying field order, I think jackson supports that right? ``@JsonPropertyOrder``?

carterkozak: We could provide consistent order, but I don't think the conjure spec should. Not sure its worth it just for this sort of test, as written I think it should give us sufficient context.

jkozlowski: ok let's ship it. I still feel like consistent order might be useful for these sort of tests, whenever I see serialization code, I want to see the ACTUAL string/bytes. Roundtrip tests are obviously useful, not denying that, but nothing beats just seeing the actual output.

PR #1306, opened 2021-04-09
Annotations to map back to defined service and endpoint name

esword:
## Before this PR
We are working on a project to detect what endpoints are called by a consumer of conjure APIs so that we can determine
what a true minimum version bounds is for that consumer based on the set of used endpoints rather than the entire API.  In order to detect what endpoints are invoked and then match that to the conjure IR, we need to be able to map between the endpoint information in the IR (specifically, HTTP method + path) and the class and method names in the source code.  These annotations allow us to do so.

## After this PR
==COMMIT_MSG==
Annotations to map back to defined service and endpoint name
==COMMIT_MSG==

## Possible downsides?
Too much awesomeness in one place.


carterkozak:
I don't understand why we need annotations on the top level class as well as the endpoint, I think it's brittle to assume all generators will always match the structure of the definition schema.

Should the annotation name reflect that it's meant to specifically capture client interactions? If I build another services server-side API, it has no bearing on my dependency on that service. In that vein, I don't think we should be annotating the conujre-undertow endpoints.

For discussion:
annotating with HTTP verb/path as opposed to service name/endpoint tuple. Either should work, much of our instrumentation is built around the latter, and it's more obvious that the name cannot be used for implementation logic which is why I would lean in that direction. There may be a case I haven't considered though.
Note that in either case we have to map from the source-code metadata to the current IR before checking compatibility, both verb-path and service-endpoint are allowed to change in non-breaking ways over time.

esword:
I am working on scanning for the annotations in another project now and will see if having the service-level one makes things significantly easier.  I like it for identification purposes, even if it isn't strictly required for this project.

I'm open to other names for the annotation(s).  I was drawing a blank.

Can you clarify what you mean by not annotating the conjure-undertow endpoints?  (This is the problem with working in a codebase one doesn't entirely understand - I don't know all what all the different pieces are for the different libraries.)

As for what to annotate with - the only reason I did verb+path was I thought you had indicated it was closest to "canonical id" for an endpoint.  If it's not, then sure, I'll swap it to the more human understandable one.  Or I can just annotate with all of them...

carterkozak:
> Can you clarify what you mean by not annotating the conjure-undertow endpoints?

Of course! We generate bindings for clients (dialogue) and servers (undertow) as well as more ambiguous bindings that may be used for either clients or servers (jersey).
If I build a distribution package which includes jars that call into the dialogue `FooService.ping` method (from an API with declares a product-dependency on `foo`) and I also invoke a conjure-undertow (server implementation, not a client) `BarService.ping` (which may be created by a product `bar`), I should only need to worry about the `FooService.ping` dependency because it's the code that can interact with a remote service, even if I'm impersonating a `bar` server. In that case I would be acting as a `bar`, providing the api, not consuming it.

I hope that made some amount of sense -- I can VTC through it if you like.

esword:
Ah - didn't know undertow was only server.  Makes sense, and given that
context renaming the annotation to ConjureClientEndpoint also makes sense.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientService.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Signifies that the annotated class is a conjure service.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.TYPE)
+public @interface ConjureClientService {
```
jkozlowski: I'll see if I can answer this myself later? Is it not that this should be just @ConjureClient?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Signifies that the given method corresponds to a conjure endpoint with the given http path and method.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.METHOD)
+public @interface ConjureClientEndpoint {
+    String method();
+
+    String path();
```
jkozlowski: Dunno if the coupling is right, but there's HttpMethod in com.palantir.conjure.spec

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientService.java
```diff
@@ -0,0 +1,33 @@
...
+    String package_() default "";
```
jkozlowski: Is this not always just the package/class name of the annotated class?

jkozlowski (commented):
Just questions from me

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
```diff
@@ -0,0 +1,35 @@
...
+ * Signifies that the given method corresponds to a endpoint invoked by a conjure client.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.METHOD)
+public @interface ConjureClientEndpoint {
+    String method();
+
+    String path();
+
+    String name();
```
carterkozak: service-endpoint and method-path should both be unique identifiers, I'm not sure we need both

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientService.java
carterkozak: local codegen plugin applies a different package prefix.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientService.java
carterkozak: I'm still not sure I understand why we need a class-level annotation in addition to method.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
esword: No, we don't need both necessarily, but I was experimenting with switching to service-endpoint so added that.  It turns out using service-endpoint isn't as easy as hoped - the code that generates the method code only has an EndpointDefinition available to it, not a  ServiceDefinition, so we don't have access to the service name. I don't want to go modifying all the endpoint generation code just to pass the service name through.

An alternative would be to use the Service annotation in combination with the Endpoint annotation - getting the service name from one and the endpoint from the other. I know you didn't like having the other annotation, but it may be more clear to readers that way.

Regardless, I think having the name in the Endpoint annotation doesn't do any harm, and makes it easier to recognize when reading.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientService.java
esword: it may be, but, as carter would say, that's fragile to assume. I'd rather have it explicit. Carter also wants to not have this annotation at all and just use the Endpoint annotation. I have the client side code working with just the Endpoint information, so maybe that is good enough, even though I like having the service annotation for reference when reading the source code.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
esword: I want to keep type simple for the annotation. As it is the only place where this gets filled in uses the definition.getHttpMethod().get() to populate, so I expect that is fine.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
jkozlowski: For what it's worth, changing code-generating code is pretty low lift, so I wouldn't be opposed to the refactoring, if it's useful.

carterkozak: +1, I'd prefer modifying the code-gen code rather than sacrificing legibility of the code it produces.

jkozlowski: What I've found useful in dialogue-annotations-processor is instantiating the *Generator classes per ServiceDefinition, instad of one per JVM and then passing ServiceDefinition to #generate method: it simplified a whole bunch of state threading to not have to pass stuff as method args through a million different codepaths. Maybe that could work here to simplify stuff?

esword: I really don't feel comfortable re-writing the generators for this small change. I'll do another commit passing the ServiceDefinition through the couple places needed

esword: Carter and I chatted offline and I think came to the conclusion that the http method-path combo which is there now is fine because it's the most concise form of annotation without introducing in new (and fragile) conventions for specifying a package+service+endpoint unique name as a string. So I'm not going to push anything new unless he comes back and says I missed something.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Signifies that the given method corresponds to a endpoint invoked by a conjure client.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.METHOD)
+public @interface ConjureClientEndpoint {
```
carterkozak: I wonder if we should rename this to `ClientEndpoint` since "Conjure" is implied by `com.palantir.conjure`?

I realize this is pedantic, but it gets much more difficult to change after the feature is merged.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.palantir.conjure.java.lib.internal;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Signifies that the given method corresponds to a endpoint invoked by a conjure client.
```
carterkozak: nit: Endpoint may or may not be invoked. Perhaps "...given method corresponds to an endpoint on a conjure client."

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
carterkozak: Agree that method is simple enough here, and we don't want conjure generated code to have a dependency onto the conjure spec types.

conjure-java-core/src/integrationInput/java/com/palantir/product/AsyncRequestProcessingTestService.java
```diff
@@ -18,41 +19,50 @@
 public interface AsyncRequestProcessingTestService {
     @GET
     @Path("async/delay")
+    @ConjureClientEndpoint(path = "/async/delay", method = "GET")
```
carterkozak: While we're here, any chance we can flip the order to more closely match a request line e.g. `GET /async/delay`

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
esword: done

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureClientEndpoint.java
esword: done

conjure-java-core/src/integrationInput/java/com/palantir/product/AsyncRequestProcessingTestService.java
esword: done

jkozlowski (approved):

PR #1308, opened 2021-04-10
Upgrade gradle-external-publish-plugin to fix published .bat scripts

AlexLandau:
See https://github.com/palantir/gradle-external-publish-plugin/pull/4

## Before this PR

The .bat scripts in conjure-java generators failed for Windows users because the gradle-external-publish-plugin modified the scripts incorrectly while shortening their classpaths.

This failure would manifest as `Error: Could not find or load main class com.palantir.conjure.java.cli.ConjureJavaCli` when trying to run tasks like `compileConjureObjects`.

## After this PR

==COMMIT_MSG==
Fix the .bat start script for the generator (used on Windows machines) to use the correct classpath. 
==COMMIT_MSG==

## Possible downsides?

The upgrade may cause other changes I'm not aware of.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.palantir.gradle.externalpublish:gradle-external-publish-plugin:0.4.10'
+        classpath 'com.palantir.gradle.externalpublish:gradle-external-publish-plugin:1.0.1'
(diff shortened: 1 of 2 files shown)


AlexLandau:
I tested this locally by running `./gradlew :conjure-java:startScripts` and checking that `conjure-java/build/scripts/conjure-java.bat` had the classpath I expected, though I have not run the resulting script on a Windows machine myself.

carterkozak (approved):
Thanks!

PR #1312, opened 2021-04-15
conjure-undertow validates exactly one Content-Type is received

carterkozak:
## Before this PR
Missing validation, allowing invalid requests to potentially succeed, or fail in an unexpected way.

## After this PR
==COMMIT_MSG==
conjure-undertow validates exactly one Content-Type is received
==COMMIT_MSG==

## Possible downsides?
Requests will fail if this was working as expected.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
@@ -251,8 +251,8 @@ final class ConjureBodySerDe implements BodySerDe {
     }
 
     /**
-     * Gets the request {@code Content-Type} header if exactly one value exists, otherwise throws
-     * a {@link SafeIllegalArgumentException}. This prevents unexpected behavior when multiple
+     * Gets the request {@code Content-Type} header if exactly one value exists, otherwise logs
+     * a warning. This notifies us in the unexpected case when multiple
      * content-type headers are incorrectly sent to the server, it's not clear which should
      * be used.
      */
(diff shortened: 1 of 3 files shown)


jkozlowski (commented):
So just thinking about rollout: this will fail requests to the client (because it's a 4xx), but we don't necessarily monitor or take 4xx into account for adjudication.

I think that makes me slightly nervous we'll unintentionally cause a P0, that will only be noticed once upstream requests start failing. With the conjure change, it was nicely compile-time that we'd reject badness, but in this case I feel like there's a chance this could go badly.

Then again, we know our clients don't do anything weird here, so it sounds like the risk is only with external clients.

Alternative is to log/metric first, see it rollout and then push it out.

carterkozak:
Updated to log a warning on failure, we can roll it out and ratchet things up if there aren't any hits.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
```diff
@@ -252,4 +249,23 @@ private static void safelyDrainRequestBody(HttpServerExchange exchange) {
             adapter.tag(target, "contentType", BINARY_CONTENT_TYPE);
         }
     }
+
+    /**
+     * Gets the request {@code Content-Type} header if exactly one value exists, otherwise throws
+     * a {@link SafeIllegalArgumentException}. This prevents unexpected behavior when multiple
```
jkozlowski: Otherwise logs.

jkozlowski (approved):

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
carterkozak:
```suggestion
     * Gets the request {@code Content-Type} header if exactly one value exists, otherwise logs
     * a warning. This notifies us in the unexpected case when multiple
```

PR #1324, opened 2021-04-21
Conjure-undertow uses standard tag names

carterkozak:
This adds few more pieces of data to the top level span.

==COMMIT_MSG==
Conjure-undertow uses standard tag names
==COMMIT_MSG==


carterkozak:
Test coverage would be nice, but I'd rather try this out against apm and see how it does, it may end up changing dramatically.

jkozlowski (approved):

PR #1330, opened 2021-04-23
Re-apply endpoint information to conjure-undertow spans

carterkozak:
## Before this PR
Spans were difficult to understand  when viewed in the trace graph.

==COMMIT_MSG==
Re-apply endpoint information to conjure-undertow spans
==COMMIT_MSG==


smoorpal (approved):

smoorpal:
This looks good to me and I think it makes sense to aggregate by this. As long as we continue to keep the status_code/method tags I think that makes sense.

fawind (approved):

carterkozak:
Yep, I've kept all the structured tag data including the components that are duplicated within the operation name, that way we can do analysis without any sort of string parsing :-)

PR #1359, opened 2021-05-17
Server request context provides an accessor for peer certificates

carterkozak:
==COMMIT_MSG==
Server request context provides an accessor for peer certificates
==COMMIT_MSG==

Ideally we'd have test coverage, but this is a very thin shim over the undertow/java security APIs. We don't currently do any TLS testing with conjure-undertow in this repo.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
@@ -87,16 +87,16 @@ final class ConjureContexts implements Contexts {
         }
 
         @Override
-        public Optional<ImmutableList<Certificate>> peerCertificates() {
+        public ImmutableList<Certificate> peerCertificates() {
             SSLSession sslSession = exchange.getConnection().getSslSession();
             if (sslSession != null) {
                 try {
-                    return Optional.of(ImmutableList.copyOf(sslSession.getPeerCertificates()));
+                    return ImmutableList.copyOf(sslSession.getPeerCertificates());
                 } catch (SSLPeerUnverifiedException e) {
                     log.debug("Failed to extract peer certificates", e);
                 }
             }
-            return Optional.empty();
+            return ImmutableList.of();
         }
 
         private ImmutableListMultimap<String, String> buildQueryParameters() {
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
@@ -24,13 +24,19 @@ import com.palantir.conjure.java.undertow.lib.RequestContext;
 import com.palantir.logsafe.Arg;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.util.HeaderValues;
+import java.security.cert.Certificate;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.net.ssl.SSLSession;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 final class ConjureContexts implements Contexts {
+    private static final Logger log = LoggerFactory.getLogger(ConjureContexts.class);
     private final RequestArgHandler requestArgHandler;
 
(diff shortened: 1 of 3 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
```diff
@@ -49,4 +51,9 @@
      * Implementations may choose to include this data in the request log.
      */
     void requestArg(Arg<?> arg);
+
+    /**
+     * Returns the client certificates associated with the connection used to make the current request.
+     */
+    Optional<ImmutableList<Certificate>> peerCertificates();
```
jkozlowski: Why ImmutableList in the API?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
```diff
@@ -80,6 +86,19 @@ public void requestArg(Arg<?> arg) {
             requestArgHandler.arg(exchange, arg);
         }
 
+        @Override
+        public Optional<ImmutableList<Certificate>> peerCertificates() {
+            SSLSession sslSession = exchange.getConnection().getSslSession();
+            if (sslSession != null) {
+                try {
+                    return Optional.of(ImmutableList.copyOf(sslSession.getPeerCertificates()));
```
jkozlowski, line 94: Wouldn't Collections.unmodifiable suffice?

jkozlowski: Do these get modified after the request gets passed for execution?

jkozlowski (commented):
Some questions

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
jkozlowski, line 94:
```
Note: The returned value may not be a valid certificate chain and should not be relied on for trust decisions.
```

What does that mean?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
carterkozak: No strong opinion on this, `ImmutableList` makes it clear that the list cannot be modified and we already use guava types for conjure ListenableFuture as well as `ListMultimap` a few lines above in this file.
List would be fine, mostly wanted to avoid the `Certificate[]` type returned by the `SSLSession`.

jkozlowski (approved):

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
carterkozak: I shouldn't have used an optional, empty list provides the same information, and consumers will likely already have to check for an empty list.
```suggestion
    ImmutableList<Certificate> peerCertificates();
```

jkozlowski: I thought about that, but then realized that Optional is not a bad way to signal that not everything will use this feature.

jkozlowski: Alternatively, maybe you want to actually have a small POJO? Reading the javadoc:

```
Returns:
an ordered array of peer certificates, with the peer's own certificate first followed by any certificate authorities.
```

There's clearly some structure to this. Do we know how people will validate the certs? Maybe some utils would be useful?

carterkozak: I think the `RequestContext` itself is a collection of features that most consumers won't need, but unblock specific use cases. In thinking about how this would be used, I'm pretty sure the list empty check will still have to be run even when the optional is present, so at that point list.isEmpty is just as easy to check. I don't know what it would mean if you got a present optional of an empty list, that should be an impossible state, but the api doesn't enforce/restrict it.

carterkozak: > Do we know how people will validate the certs? Maybe some utils would be useful?

Since this is a drop-through to the SSLSession, standard tooling should be sufficient and I don't think we want to enforce opinions.

PR #1369, opened 2021-05-24
Fix: Emit bean fields in the same order as the builder stages

ferozco:
## Before this PR
When using the staged builder feature We would emit fields on the bean in a different order than the stages in the builder

## After this PR

==COMMIT_MSG==
Emit bean fields in the same order as the builder stages
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1369.v2.yml
+++ b/changelog/@unreleased/pr-1369.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Emit bean fields in the same order as the builder stages
+  links:
+  - https://github.com/palantir/conjure-java/pull/1369
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
@@ -99,7 +99,12 @@ public final class BeanBuilderGenerator {
             ObjectDefinition typeDef,
             Map<com.palantir.conjure.spec.TypeName, TypeDefinition> typesMap,
             Optional<ClassName> builderInterfaceClass) {
-        Collection<EnrichedField> enrichedFields = enrichFields(typeDef.getFields());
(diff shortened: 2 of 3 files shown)


carterkozak:
I think this will impact the resulting JSON, while we don't specify any particular order I think it's nice to match the definition field order when possible. Can you describe the motivation for this change?

ferozco:
Inside of phase we generate code that uses conjure-java's builders and using staged builders ends up breaking our generated code since we populate the fields in the order they are declared in the object. I was hoping to avoid copying the ordering code into phase by having the fields emitted in the right order

carterkozak:
Hmm, when I wrote the excavator to migrate to staged builders, I found that field order is not retained in java classes, which is why I had to opt into javac `-parameters` and use the constructor parameter order. Fields are sorted to pack bytes by type, and that has recently changed between releases.

Would it be possible to leverage the returned builder type setter methods to set fields in order?

ferozco:
It might be possible but it would be a fair bit more work and feels a more brittle since we would rely on even more of the implementation details of the code gen. Let me see how bad it will be

carterkozak:
Perhaps I didn't explain the idea very well, I was thinking more of a generalized solution for staged builders where we provide data based on the builder class/interface contract in the same way a human would interact with a staged builder.

For example if `Foo.builder()` returns:
```java
interface FooBuilderBazStage {
    FooBuildeFinalStage baz(Baz value);
}
```

We know the first field we need is `baz` of type `Baz`, then we can progress to the next field/stage, which in this case is `FooBuildeFinalStage`. I'm not very familiar with how phase is implemented, I may be underestimating the changes required to implement this, but I think it can be done in a way that's not brittle and closely matches how humans interact with these types.

carterkozak:
To clarify, the first sentence about parameters was an anecdote describing why we can't use the field declarations directly, not suggesting that we use constructor params, that would be brittle. I'm suggesting that we follow the staged builder types to fill out the staged builder similarly to how humans do it -- sorry for the confusion.

ferozco:
I followed Carter's advice and was able to implement something in phase that inferred the order from the stages so this shouldn't be necessary anymore

PR #1375, opened 2021-06-01
Revert default optional field handling to serialize a null value

carterkozak:
This reverts the default behavior to serialize empty optional
values as null until Typescript uses can be cleaned up to use
the provided type information from the typescript generator.

## This is meant to be reverted within the next month

==COMMIT_MSG==
Revert default optional field handling to serialize a null value
==COMMIT_MSG==

## Possible downsides?
More backcompat p0s

### DIFF
--- a/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
+++ b/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
@@ -189,6 +189,12 @@ public final class ConjureJavaCli implements Runnable {
                 description = "Generates compile-time safe builders to ensure all required attributes are set.")
         private boolean useStagedBuilders;
 
+        @CommandLine.Option(
+                names = "--excludeEmptyOptionals",
+                defaultValue = "false",
+                description = "Objects exclude empty optionals in serialization based on the conjure spec.")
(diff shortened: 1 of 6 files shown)


conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -189,6 +189,12 @@ public void run() {
                 description = "Generates compile-time safe builders to ensure all required attributes are set.")
         private boolean useStagedBuilders;
 
+        @CommandLine.Option(
+                names = "--excludeEmptyOptionals",
+                defaultValue = "false",
+                description = "Objects exclude empty optionals in serialization based on the conjure spec.")
+        private boolean excludeEmptyOptionals;
```
jkozlowski, line 196: This can just work with gradle-conjure? can you pass that like e.g. useImmutableBytes?

carterkozak: That’s correct

jkozlowski (approved):

tpetracca:
Linking original PR for clarity: https://github.com/palantir/conjure-java/pull/1300

iamdanfox:
Did this upset typescript consumers??

PR #1383, opened 2021-06-14
conjure-undertow GET endpoints support HEAD requests as well

carterkozak:
## Before this PR
`HEAD` requests to `GET` endpoints produced 404 "Not Found" responses.

## After this PR

HEAD requests don't produce response bytes, however if a response
is written, the bytes are safely, locally, written to an empty
sink. This matches Jersey behavior.
Depending how this is used, we may eventually want to allow folks
to implement their own HEAD handlers, this change doesn't make
that any more difficult.

==COMMIT_MSG==
conjure-undertow GET endpoints support HEAD requests as well
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
@@ -358,6 +358,32 @@ public final class UndertowServiceEteTest extends TestBase {
     }
 
     @Test
+    public void testGetMethodsAllowHeadRequests() throws IOException {
+        URL url = new URL("http://0.0.0.0:8080/test-example/api/base/string");
+        HttpURLConnection con = (HttpURLConnection) url.openConnection();
+        con.setRequestMethod("HEAD");
+        con.setRequestProperty(HttpHeaders.ACCEPT, "application/json");
+        con.setRequestProperty(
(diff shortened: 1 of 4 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -357,6 +357,32 @@ public void testExternalImportOptionalEmptyBodyZeroLength_withContentType() thro
...
+    @Test
+    public void testOptionsOnGetIncludesHead() throws IOException {
+        URL url = new URL("http://0.0.0.0:8080/test-example/api/base/string");
+        HttpURLConnection con = (HttpURLConnection) url.openConnection();
+        con.setRequestMethod("OPTIONS");
+        con.setRequestProperty(HttpHeaders.ACCEPT, "application/json");
+        con.setRequestProperty(
+                HttpHeaders.AUTHORIZATION, AuthHeader.valueOf("authHeader").toString());
+        assertThat(con.getResponseCode()).isEqualTo(204);
+        assertThat(con.getHeaderField(HttpHeaders.ALLOW)).isEqualTo("GET, HEAD");
```
carterkozak, line 383: In retrospect, the options handler should be more self-aware... `OPTIONS, GET, HEAD`. That's a change for a separate PR though!

tetigi (approved):
LGTM!

PR #1468, opened 2021-08-26
conjure-undertow logs I/O exceptions from closed connections at INFO

carterkozak:
## Before this PR
We logged these at the ERROR level, causing unnecessary panic.

## After this PR
This matches existing client error (4xx) logging, and avoids
noisy logging when clients cancel requests. This is particularly
common for use-cases involving larger binary downloads from
browsers, where clients frequently navigate away from the page,
cancel downloads, or go to sleep.

==COMMIT_MSG==
conjure-undertow logs I/O exceptions from closed connections at INFO
==COMMIT_MSG==

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
@@ -47,9 +47,8 @@ public enum ConjureExceptions implements ExceptionHandler {
 
     private static final SafeLogger log = SafeLoggerFactory.get(ConjureExceptions.class);
     // Exceptions should always be serialized using JSON
-    private static final Serializer<SerializableError> serializer = new ConjureBodySerDe(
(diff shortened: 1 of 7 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -66,6 +65,10 @@ public void handle(HttpServerExchange exchange, Throwable throwable) {
             frameworkException(exchange, (FrameworkException) throwable);
         } else if (throwable instanceof Error) {
             error(exchange, (Error) throwable);
+        } else if (throwable instanceof IOException && !exchange.getConnection().isOpen()) {
```
carterkozak, line 68: The IOException component of this check might be unnecessary, and there's no guarantee the throwable hasn't been re-wrapped, but I'd prefer to be conservative here. I'd rather accidentally log a few events at a more severe level than log them too finely.

jkozlowski (approved):

PR #1477, opened 2021-09-08
Generate ErrorNs.is{Error} static methods for checking conjure errors

kristofarkas:
## Before this PR
It was very common for people to write this method that checks for the type of the remote exception. But writing this by hand is very error prone, especially because it's not clear if one should use ServiceException or RemoteException in the catch clause, and what exactly to compare.

## After this PR
Generate is{Error} methods for conjure errors.

An example usage based on this error definition
```yaml
types:
  definitions:
    default-package: com.palantir.example
    errors:
      PermissionDenied:
        namespace: ExampleErrors
        code: PERMISSION_DENIED

```

could be like this:
```java

public void main() {
    try {
        exampleServiceBlocking.getExample();
    } catch (RemoteException e) {
        if (ExampleErrors.isPermissionDenied(e)) {
            log.error("Permission denied");
        } else {
            throw e;
        }
    }
}

```

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
@@ -52,6 +52,8 @@ import org.apache.commons.lang3.StringUtils;
 
 public final class ErrorGenerator implements Generator {
 
+    private static final String REMOTE_EXCEPTION_VAR = "remoteException";
+
     private final Options options;
 
     public ErrorGenerator(Options options) {
@@ -177,16 +179,19 @@ public final class ErrorGenerator implements Generator {
                             CaseFormat.UPPER_UNDERSCORE, entry.getErrorName().getName());
                     String methodName = "is" + entry.getErrorName().getName();
 
-                    String remoteExceptionVar = "remoteException";
-
                     return MethodSpec.methodBuilder(methodName)
                             .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
-                            .addParameter(RemoteException.class, remoteExceptionVar)
+                            .addParameter(RemoteException.class, REMOTE_EXCEPTION_VAR)
                             .returns(TypeName.BOOLEAN)
+                            .addStatement(Expressions.requireNonNull(
+                                    REMOTE_EXCEPTION_VAR, "remote exception must not be null"))
                             .addStatement(
-                                    "return $L.name().equals($L.getError().errorName())", typeName, remoteExceptionVar)
+                                    "return $N.name().equals($N.getError().errorName())",
+                                    typeName,
+                                    REMOTE_EXCEPTION_VAR)
                             .addJavadoc(
-                                    "Returns true if the {@link RemoteException} is named $L:$L",
+                                    "Returns true if the {@link $T} is named $L:$L",
+                                    RemoteException.class,
                                     entry.getNamespace(),
                                     entry.getErrorName().getName())
                             .build();
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
@@ -179,19 +179,17 @@ public final class ErrorGenerator implements Generator {
 
                     String remoteExceptionVar = "remoteException";
 
-                    MethodSpec.Builder methodBuilder = MethodSpec.methodBuilder(methodName)
+                    return MethodSpec.methodBuilder(methodName)
                             .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
                             .addParameter(RemoteException.class, remoteExceptionVar)
                             .returns(TypeName.BOOLEAN)
                             .addStatement(
-                                    "return $L.name().equals($L.getError().errorName())", typeName, remoteExceptionVar);
-
-                    methodBuilder.addJavadoc(
-                            "Returns true if the {@link RemoteException} is named $L:$L",
-                            entry.getNamespace(),
-                            entry.getErrorName().getName());
-
-                    return methodBuilder.build();
+                                    "return $L.name().equals($L.getError().errorName())", typeName, remoteExceptionVar)
+                            .addJavadoc(
+                                    "Returns true if the {@link RemoteException} is named $L:$L",
+                                    entry.getNamespace(),
+                                    entry.getErrorName().getName())
+                            .build();
                 })
                 .collect(Collectors.toList());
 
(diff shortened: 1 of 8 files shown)


palantirtech:
Thanks for your interest in [palantir/conjure-java](https://github.com/palantir/conjure-java), @kristofarkas! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
```diff
@@ -172,12 +171,37 @@ private static JavaFile generateErrorTypesForNamespace(
                 })
                 .collect(Collectors.toList());
 
+        List<MethodSpec> isRemoteExceptionDefinitions = errorTypeDefinitions.stream()
+                .map(entry -> {
+                    String typeName = CaseFormat.UPPER_CAMEL.to(
+                            CaseFormat.UPPER_UNDERSCORE, entry.getErrorName().getName());
+                    String methodName = "is" + entry.getErrorName().getName();
```
kristofarkas, line 180: Some other naming ideas:
`is{Error}Error`
`exceptionIs{Error}`

pkoenig10: It seems overly verbose to repeat the error name in the method name. Then call sites looks like:
```java
MyConjureError.isMyConjureError(e);
```

Would be less verbose to just do something like (not suggesting this exact naming):
```java
MyConjureError.hasName(e);
```

pkoenig10 (commented):
This also need

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
kristofarkas: In your case MyConjureError is the namespace of the error definition, and there can be multiple errors defined in it.

pkoenig10: Ah sorry, I misread this. This looks fine.

tpetracca:
I'm fairly certain the generated Errors classes aren't actually on the client's classpaths which you would need here.

We've discussed historically making error inspection client-side easier but never ended up executing against it.

CC @carterkozak

pkoenig10:
> I'm fairly certain the generated Errors classes aren't actually on the client's classpaths which you would need here.

This isn't true. The generated Errors classes are in the `objects` package. This is how clients do error introspection today, since they need access to the names. This PR just makes it a bit more convenient.

carterkozak:
They're currently shared, however that's a bit of a bug because the types the server deals with differ from the types the client is given -- for example a client should never throw another services ServiceException+ErrorType, but it's reasonable for a client to parse data from a local RemoteException from said service.

It's unfortunate that we've coupled the two. We'd discussed generating specific typed exceptions (one for clients and another for servers) so that they may be handled using standard java practices, I think that may be a better solution but requires more work to deliver.

kristofarkas:
As @pkoenig10 said, this PR just makes what people are already doing a bit more convenient, and easier to get right. I've made the mistake of incorrectly catching `ServiceException`s multiple times, and have seen other people making the same mistake.

@carterkozak happy to explore the path of generating `RemoteException` subclasses, that people could catch, but even so, one might still need to write code to catch the generic `RemoteException` and check the name, etc. because of backward compat and just to be safe, no?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
```diff
@@ -172,12 +171,35 @@ private static JavaFile generateErrorTypesForNamespace(
...
+                    String remoteExceptionVar = "remoteException";
+
+                    return MethodSpec.methodBuilder(methodName)
+                            .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+                            .addParameter(RemoteException.class, remoteExceptionVar)
+                            .returns(TypeName.BOOLEAN)
+                            .addStatement(
+                                    "return $L.name().equals($L.getError().errorName())", typeName, remoteExceptionVar)
+                            .addJavadoc(
+                                    "Returns true if the {@link RemoteException} is named $L:$L",
```
carterkozak: nit: Formatting may be better using something along these lines:
```suggestion
                                    "Returns true if the {@link $T} is named {@code $L:$L}", RemoteException.class,
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
```diff
@@ -172,12 +171,35 @@ private static JavaFile generateErrorTypesForNamespace(
                 })
                 .collect(Collectors.toList());
 
+        List<MethodSpec> isRemoteExceptionDefinitions = errorTypeDefinitions.stream()
+                .map(entry -> {
+                    String typeName = CaseFormat.UPPER_CAMEL.to(
+                            CaseFormat.UPPER_UNDERSCORE, entry.getErrorName().getName());
+                    String methodName = "is" + entry.getErrorName().getName();
+
+                    String remoteExceptionVar = "remoteException";
```
carterkozak: nit: Please extract this to a `private static final` constant in this class

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
```diff
@@ -172,12 +171,35 @@ private static JavaFile generateErrorTypesForNamespace(
...
+
+                    return MethodSpec.methodBuilder(methodName)
+                            .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+                            .addParameter(RemoteException.class, remoteExceptionVar)
```
carterkozak: Let's add a safe-logging preconditions null check on this parameter before we get to the equality check

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
```diff
@@ -172,12 +171,35 @@ private static JavaFile generateErrorTypesForNamespace(
...
+                            .returns(TypeName.BOOLEAN)
+                            .addStatement(
+                                    "return $L.name().equals($L.getError().errorName())", typeName, remoteExceptionVar)
```
carterkozak: https://github.com/square/javapoet#n-for-names

```suggestion
                                    "return $N.name().equals($N.getError().errorName())", typeName, remoteExceptionVar)
```

carterkozak:
Could you update this with a test before it merges? Something that round trips the exceptions we create (`ServiceException`) through `SerializableError.forException(` -> `RemoteException` -> `is...`
A positive and negative case would be helpful, perhaps including an error with the same name from a different service e.g. `Foo:Bar` vs `Baz:Bar`, that way we don't mess it up in the future :-)

kristofarkas:
@carterkozak resolved all the comments.

I couldn't get CI to run and I think that's because I created this PR from a fork of my repo. Created a separate PR that looks okay: https://github.com/palantir/conjure-java/pull/1507.

kristofarkas:
Closing this, please leave any new comments on #1507

PR #1492, opened 2021-09-20
Conjure builders throw when they're reused after a `build()` invocation

carterkozak:
This behavior was never expected to work, and can result in unexpected
mutability in collections.

==COMMIT_MSG==
Conjure builders throw when they're reused after a `build()` invocation
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
@@ -42,6 +42,7 @@ import com.palantir.conjure.spec.Type;
 import com.palantir.conjure.spec.TypeDefinition;
 import com.palantir.conjure.visitor.TypeDefinitionVisitor;
 import com.palantir.conjure.visitor.TypeVisitor;
+import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
@@ -70,6 +71,9 @@ import java.util.stream.Collectors;
 import javax.lang.model.element.Modifier;
 
 public final class BeanBuilderGenerator {
+
+    private static final String BUILT_FIELD = "_buildInvoked";
+
     private final TypeMapper typeMapper;
     private final ClassName builderClass;
     private final ClassName objectClass;
@@ -106,6 +110,7 @@ public final class BeanBuilderGenerator {
                         isInStagedBuilderMode(builderInterfaceClass) ? "DefaultBuilder" : "Builder")
                 .addAnnotation(ConjureAnnotations.getConjureGeneratedAnnotation(BeanBuilderGenerator.class))
                 .addModifiers(Modifier.STATIC, Modifier.FINAL)
+                .addField(FieldSpec.builder(boolean.class, BUILT_FIELD).build())
                 .addFields(poetFields)
                 .addFields(primitivesInitializedFields(enrichedFields))
                 .addMethod(createConstructor())
@@ -215,6 +220,13 @@ public final class BeanBuilderGenerator {
         return MethodSpec.constructorBuilder().addModifiers(Modifier.PRIVATE).build();
     }
 
+    private CodeBlock verifyNotBuilt() {
+        return CodeBlock.builder()
+                .addStatement(
+                        "$T.checkState(!$N, $S)", Preconditions.class, BUILT_FIELD, "Build has already been called")
+                .build();
+    }
+
     private MethodSpec createFromObject(Collection<EnrichedField> enrichedFields) {
         CodeBlock assignmentBlock = CodeBlocks.of(Collections2.transform(
                 enrichedFields,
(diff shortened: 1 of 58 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -106,6 +110,7 @@ private TypeSpec generate(
                         isInStagedBuilderMode(builderInterfaceClass) ? "DefaultBuilder" : "Builder")
                 .addAnnotation(ConjureAnnotations.getConjureGeneratedAnnotation(BeanBuilderGenerator.class))
                 .addModifiers(Modifier.STATIC, Modifier.FINAL)
+                .addField(FieldSpec.builder(boolean.class, BUILT_FIELD).build())
```
jkozlowski, line 114: How much more expensive would a volatile be here?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -225,6 +237,7 @@ private MethodSpec createFromObject(Collection<EnrichedField> enrichedFields) {
                 .addModifiers(Modifier.PUBLIC)
                 .returns(builderClass)
                 .addParameter(objectClass, "other")
+                .addCode(verifyNotBuilt())
```
jkozlowski, line 235: It's a bit concerning that this had to be added in so many places: granted there's probably not much change that's happening in this codepath, but maybe there's something we could refactor here?

I don't see any shared interfaces in javapoet, but maybe something like:

```
        return guardNotBuilt(
                        MethodSpec.methodBuilder("from")
                                .returns(builderClass)
                                .addParameter(objectClass, "other")::addCode,
                        assignmentBlock)
                .addStatement("return this")
                .build();
    }

    private <T> T guardNotBuilt(Function<CodeBlock, T> addCode, CodeBlock codeBlock) {
        addCode.apply(verifyNotBuilt());
        return addCode.apply(codeBlock);
    }
```

There's probably a nicer way to do this.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -394,6 +409,7 @@ private boolean isByteBuffer(Type type) {
     private MethodSpec createOptionalSetter(EnrichedField enriched) {
         OptionalType type = enriched.conjureDef().getType().accept(TypeVisitor.OPTIONAL);
         return BeanBuilderAuxiliarySettersUtils.createOptionalSetterBuilder(enriched, typeMapper, builderClass)
+                .addCode(verifyNotBuilt())
```
jkozlowski, line 407: Code gen vs JVM inlining this? Feels a bit bloaty

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
jkozlowski: How about this: I think we can use the fact that javapoet builders are not really staged builders:

```
  private MethodSpec createBuild(Collection<EnrichedField> enrichedFields, Collection<FieldSpec> fields) {
        MethodSpec.Builder method = newSetter("build")
                .returns(objectClass)
                .addStatement("this.$N = true", BUILT_FIELD);

        if (enrichedFields.stream().filter(EnrichedField::isPrimitive).count() != 0) {
            method.addStatement("validatePrimitiveFieldsHaveBeenInitialized()");
        }

        return method.addStatement("return new $L", Expressions.constructorCall(objectClass, fields))
                .build();
    }

    private MethodSpec.Builder newSetter(String name) {
        return MethodSpec.methodBuilder(name).addCode(verifyNotBuilt());
    }
```

jkozlowski (commented):
Some comments, but thanks for quick turnaround!

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
carterkozak: Ya I got lazy -- should use a private verifier method.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
carterkozak: The custom builder-factory method sounds nice -- I'd avoided doing anything special since the risk is it makes the code a little harder to follow, but becomes easier to miss validation in future changes. I think `newSetter` hits the sweet spot.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
carterkozak: Not particularly compared to other operations, but nothing in the builder is threadsafe so I don't think it would make much sense.

jkozlowski: Yeah, I think you're right. Although it would probably detect some badness if someone *WERE* to do something really crazy like pass a builder around...

jkozlowski (approved):
Can be merged after small FLUPs

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
carterkozak: This would be more involved than I thought because we generate the method signature for some methods using `BeanBuilderAuxiliarySettersUtils` functions, and those are reused between interfaces and the implementation. We can share it, but I'm not sure that the maintainability trade-offs would make sense in that case.

PR #1497, opened 2021-09-23
Make min and max public

ChristopherRogers1991:
## Before this PR
Without looking it up in the code and copying the values, there's no programmatic way to know what the min/max values are.

## After this PR

==COMMIT_MSG==
Make the min and max safelong values public.
==COMMIT_MSG==

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
@@ -27,7 +27,7 @@ public abstract class SafeLong implements Comparable<SafeLong> {
 
     private static final long MIN_SAFE_VALUE = -(1L << 53) + 1;
     private static final long MAX_SAFE_VALUE = (1L << 53) - 1;
-    
+
     public static final SafeLong MAX_VALUE = SafeLong.of(MAX_SAFE_VALUE);
     public static final SafeLong MIN_VALUE = SafeLong.of(MIN_SAFE_VALUE);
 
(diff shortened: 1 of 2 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
```diff
@@ -25,8 +25,8 @@
 @Value.Immutable
 public abstract class SafeLong implements Comparable<SafeLong> {
 
-    private static final long MIN_SAFE_VALUE = -(1L << 53) + 1;
-    private static final long MAX_SAFE_VALUE = (1L << 53) - 1;
+    public static final long MIN_SAFE_VALUE = -(1L << 53) + 1;
```
ferozco: how about we expose a `Optional<SafeLong> safeValueOf(long value)` method instead?

ferozco: cc @carterkozak if he has thoughts

ChristopherRogers1991: We could do that, but I think we would still need to expose the max values, as is done with other numerical types (e.g. https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#MAX_VALUE).

If I _have_ to give a value, and safeValueOf gave me an empty optional, I need a way to determine what to report. I think it's relatively common to use the max value in that case, to represent the "you've gone past the limit" or "infinity" case.

carterkozak: +1, I'm not sure what we'd expect folks to do in the `Optional.empty()` case without enough information to determine why the value is invalid.

Generally agree that it's helpful to expose the bounds in some way.

In terms of API design, what if we match `Long`, `Integer`, etc and provide two (new) public constants: `MAX_VALUE` and `MIN_VALUE`, but expose them as the `SafeLong` type?

carterkozak (approved):
looks good to me, thanks for the contribution!

fawind (approved):

CRogers:
:+1:

PR #1501, opened 2021-09-28
Throw ServiceException instead of SafeIllegalException when fields are missing

viktornordling:
## Before this PR
Currently, if a client submits a json payload which is incomplete (missing fields), the client only gets a `Default:InvalidArgument` back with no information about which fields were missing. These fields do get logged, but it makes it harder for clients to debug the issues.

## After this PR
The client will get a `ServiceException` which contains the missing fields.
==COMMIT_MSG==
Throw ServiceException instead of SafeIllegalException when fields are missing
==COMMIT_MSG==

## Possible downsides?
This is compatibility breaking, if a client is relying on the `Default:InvalidArgument` error to be thrown, their code will no longer do what they expect. Granted, there's not much the code _can_ do in this case, but some catch statements may fail to catch the right kind of exception.


palantirtech:
Thanks for your interest in [palantir/conjure-java](https://github.com/palantir/conjure-java), @viktornordling! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1507, opened 2021-10-06
Generate ErrorNs.is{Error} static methods for checking conjure errors

kristofarkas:
## Before this PR
It was very common for people to write this method that checks for the type of the remote exception. But writing this by hand is very error prone, especially because it's not clear if one should use ServiceException or RemoteException in the catch clause, and what exactly to compare.

## After this PR
Generate is{Error} methods for conjure errors.

An example usage based on this error definition
```yaml
types:
  definitions:
    default-package: com.palantir.example
    errors:
      PermissionDenied:
        namespace: ExampleErrors
        code: PERMISSION_DENIED

```

could be like this:
```java

public void main() {
    try {
        exampleServiceBlocking.getExample();
    } catch (RemoteException e) {
        if (ExampleErrors.isPermissionDenied(e)) {
            log.error("Permission denied");
        } else {
            throw e;
        }
    }
}

```

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
@@ -52,6 +52,8 @@ import org.apache.commons.lang3.StringUtils;
 
 public final class ErrorGenerator implements Generator {
 
+    private static final String REMOTE_EXCEPTION_VAR = "remoteException";
+
     private final Options options;
 
     public ErrorGenerator(Options options) {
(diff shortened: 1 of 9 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/EteErrors.java
```diff
@@ -0,0 +1,53 @@
...
+        RemoteException remoteException = serializeServiceException(ConjureJavaErrors.javaCompilationFailed());
+
+        assertThat(ConjureJavaErrors.isJavaCompilationFailed(remoteException)).isTrue();
+        assertThat(ConjureJavaOtherErrors.isJavaCompilationFailed(remoteException))
+                .isFalse();
+    }
+
+    private static RemoteException serializeServiceException(ServiceException serviceException) {
+        SerializableError serializable = SerializableError.forException(serviceException);
+        return new RemoteException(serializable, 200);
```
carterkozak, line 51: nit: Let's match the expected error-code in case that's taken into account in the future

```suggestion
        return new RemoteException(serializable, serviceException.getErrorType().httpErrorCode());
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
```diff
@@ -172,12 +173,38 @@ private static JavaFile generateErrorTypesForNamespace(
...
+                            .addParameter(RemoteException.class, REMOTE_EXCEPTION_VAR)
+                            .returns(TypeName.BOOLEAN)
+                            .addStatement(Expressions.requireNonNull(
+                                    REMOTE_EXCEPTION_VAR, "remote exception must not be null"))
+                            .addStatement(
+                                    "return $N.name().equals($N.getError().errorName())",
+                                    typeName,
+                                    REMOTE_EXCEPTION_VAR)
+                            .addJavadoc(
+                                    "Returns true if the {@link $T} is named $L:$L",
```
carterkozak, line 193: Slight preference for `code` tags to distinguish the name (super duper minor though)
```suggestion
                                    "Returns true if the {@link $T} is named {@code $L:$L}",
```

carterkozak:
I'm going to release this, if you'd like to resolve any of the nits I called out, feel free to open another PR at your convenience. Thanks for the contribution :-)

:+1:

PR #1514, opened 2021-10-18
[WIP] Add conjure-undertow annotation processor

carterkozak:
Exceedingly WIP -- not ready for full review, but the API bits should be in a good place if folks are interested in providing early feedback :-)

## Before this PR
Hand-rolled endpoints

## After this PR
==COMMIT_MSG==
Add initial version of conjure-undertow annotation processor
==COMMIT_MSG==

## Possible downsides?
Competing objectives supporting conjure and non-conjure targets. Worthwhile because most of our infrastructure can be easily reused between the two much more easily than duplicating support for tracing/logging/codegen/etc between conjure-undertow and jersey.


conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/AnnotationReflector.java
```diff
@@ -0,0 +1,90 @@
...
+    default <T> Optional<T> getFieldMaybe(String fieldName, Class<T> valueClazz) {
+        Preconditions.checkArgument(
+                methods().containsKey(fieldName),
+                "Unknown field",
+                SafeArg.of("fieldName", fieldName),
+                SafeArg.of("type", annotationTypeElement()),
+                SafeArg.of("fields", methods()));
+        Optional<Object> maybeValue = Optional.ofNullable(values().get(fieldName));
+        return maybeValue.map(value -> {
+            Preconditions.checkArgument(valueClazz.isInstance(value), "Value not of the right type");
```
fawind: Should we log some additional info here, e.g. the `fieldName`?

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ArgumentTypesResolver.java
```diff
@@ -0,0 +1,151 @@
...
+            return Optional.of(ArgumentTypes.primitive(typeName, planSerDeMethodName(typeName), Optional.empty()));
+        } else if (listType.map(innerType -> isPrimitive(TypeName.get(innerType)))
+                .orElse(false)) {
+            TypeName innerTypeName = TypeName.get(listType.get());
+            return Optional.of(
+                    ArgumentTypes.primitive(typeName, planSerDeMethodName(innerTypeName), Optional.of(innerTypeName)));
+        } else if (isRawRequestBody(actualTypeMirror)) {
+            return Optional.of(ArgumentTypes.rawRequestBody(TypeName.get(actualTypeMirror)));
+        } else if (optionalType.isPresent()) {
+            // TODO(12345): We only want to go one level down: don't allow Optional<Optional<Type>>.
```
fawind: Would quickly clean up these TODO(12345) bits before merging.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ArgumentTypesResolver.java
```diff
@@ -0,0 +1,151 @@
...
+            } else if (context.isAssignableWithErasure(declaredType, Map.class)) {
+                return Optional.of(ArgumentTypes.mapType(typeName));
+            }
+            return Optional.of(ArgumentTypes.customType(typeName));
+        });
+    }
+
+    private String planSerDeMethodName(TypeName in) {
+        String typeName = PARAMETER_SERIALIZER_TYPES.get(in.box());
+        return Preconditions.checkNotNull(typeName, "Unknown type");
```
fawind, line 127: Should we also add the unknown type name as an error arg to ease debugging?

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/EndpointDefinitions.java
```diff
@@ -0,0 +1,107 @@
...
+                .toJavaUtil()
+                .map(ImmutableAnnotationReflector::of)
+                .orElseThrow();
+
+        EndpointName endpointName =
+                ImmutableEndpointName.of(element.getSimpleName().toString());
+        ServiceName serviceName = ImmutableServiceName.of(
+                element.getEnclosingElement().getSimpleName().toString());
+
+        HttpMethod method = HttpMethod.valueOf(requestAnnotationReflector
```
fawind, line 58: I wonder if we also need to gracefully handle parsing the http method, similar to the `maybeHttpPath` and `maybeReturnType` below?

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/ConjureUndertowAnnotationProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+                    .map(MoreElements::asExecutable)
+                    .map(endpointDefinitions::tryParseEndpointDefinition)
+                    .collect(Collectors.toList());
+
+            Preconditions.checkArgument(
+                    maybeEndpoints.stream()
+                            .filter(Predicates.not(Optional::isPresent))
+                            .collect(Collectors.toList())
+                            .isEmpty(),
+                    "Failed validation");
```
fawind, line 146: Maybe instead of returning an empty optional on failed validation, we should just throw inside `EndpointDefinitions#tryParseEndpointDefinition` where we have additional error context?

conjure-undertow-processor-example/src/main/java/com/palantir/conjure/java/undertow/example/ExampleService.java
```diff
@@ -0,0 +1,105 @@
...
+import com.palantir.conjure.java.undertow.lib.BinaryResponseBody;
+import com.palantir.conjure.java.undertow.lib.RequestContext;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.tokens.auth.AuthHeader;
+import io.undertow.server.HttpServerExchange;
+import java.io.Closeable;
+import java.util.Collection;
+import java.util.Optional;
+
+public interface ExampleService {
```
fawind, line 34: What is the expectation of building a client for the generated services? Do we plan to also generate a client for this going forward? Or is the expectation that users would build/generate their own client (i.e. using [dialogue-annotations](https://github.com/palantir/dialogue/tree/3393922b5a84ccbd94e4c45498b2ec103536703d#dialogue-annotations-processor-generated-client-bindings))?

fawind (approved):
Looks reasonable overall, just added some smaller nits that caught my eye. But happy to merge this in as-is and then iterate from there as this PR is becoming quite large!

conjure-undertow-processor-example/src/main/java/com/palantir/conjure/java/undertow/example/ExampleService.java
fawind: Discussed offline. This annotation processor is primarily intended to unblock one-off consumers that are still stuck on the undertow migration, primarily due to using custom behavior that is not directly supported through generated Conjure clients.

Also generating a client based on the conjure-undertow annotations is tricky because there are subtle differences between client and server handling. Also this would come very close to re-implementing Conjure.

fawind:
Will go ahead and merge this as-is. The API seems to be in a good state and we will not advocate for using this yet, allowing us to still change the API if necessary. That way we can easier iterate and test on some real-world projects.

Can iterate and do some more clean-ups in FLUP PRs!

PR #1519, opened 2021-10-26
Log QosException at `info` (up from `debug`) when cause is present

carterkozak:
This provide additional context, particularly when the qosException
is caused by a remote request. The check for a cause or suppressed
exception helps us avoid log spam on the service which produces qos
feedback becuase clients are unlikely to exhaust all retries, and
even if all retries are always exhausted, they fail through at a
rate of 1 / NUM_RETRIES.

## Before this PR
Unclear when qos exceptions are propagated

## After this PR

==COMMIT_MSG==
Log QosException at `info` (up from `debug`) when a cause or suppressed throwable is present
==COMMIT_MSG==

## Possible downsides?
This may create noise in the logs, however the dynamic approach is meant to limit noise.

### DIFF
--- a/changelog/@unreleased/pr-1519.v2.yml
+++ b/changelog/@unreleased/pr-1519.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Log QosException at `info` (up from `debug`) when a cause or suppressed
+    throwable is present
+  links:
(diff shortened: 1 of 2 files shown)


jkozlowski (approved):
Let's give this a go, I certainly had situations where it would've been nice to know this

PR #1549, opened 2021-11-15
Support tag-based per-endpoint asynchronous processing timeouts

carterkozak:
This granularity is preferred over global configuration because
it's counterproductive to extend the timeout of _all_ endpoints
when only a small hand full are expected to take a long time to
complete.

## Before this PR
Only global configuration was allowed.

==COMMIT_MSG==
Support tag-based per-endpoint asynchronous processing timeouts
==COMMIT_MSG==


conjure-java-core/src/integrationInput/java/com/palantir/product/AsyncRequestProcessingTestServiceEndpoints.java
```diff
@@ -96,6 +100,66 @@ public HttpHandler handler() {
...
+        public Set<String> tags() {
+            return TAGS;
+        }
+
+        @Override
+        public void handleRequest(HttpServerExchange exchange) throws IOException {
+            Map<String, Deque<String>> queryParams = exchange.getQueryParameters();
+            OptionalInt delayMillis = runtime.plainSerDe().deserializeOptionalInteger(queryParams.get("delayMillis"));
+            ListenableFuture<String> result = delegate.delayFiveSecondTimeout(delayMillis);
+            runtime.async().register(result, this, Duration.ofMillis(/* 5 seconds */ 5000), exchange);
```
carterkozak, line 129: Example generated code. I retained the original human readable piece in a comment so that we can avoid expensive string parsing, while still producing readable source.

conjure-java-core/src/test/java/com/palantir/conjure/java/AsyncRequestProcessingTest.java
```diff
@@ -162,6 +162,14 @@ public void testTimeout() throws IOException {
         }
     }
 
+    @Test
+    public void testFiveSecondTimeoutIsNotReached() throws IOException {
+        try (Response response = requestToDelayEndpointWithFiveSecondTimeout(OptionalInt.of(2000))) {
+            assertThat(response).matches(Response::isSuccessful);
+            assertThat(response.body().string()).isEqualTo("\"Completed after 2000ms\"");
```
eraboin, line 169: Is there a similar test to determine if the timeout is enforced (I didn't see it here)?

eraboin (approved):

conjure-java-core/src/test/java/com/palantir/conjure/java/AsyncRequestProcessingTest.java
carterkozak: The test above verified the timeout works -- this test verified the same thing, but overrides with a longer timeout to ensure it can be increased

PR #1553, opened 2021-11-18
Change error code of timed out async requests to TIMEOUT

westinrm:
## Before this PR
Timed out async requests are logged as a CancellationException and propagated to clients as code INTERNAL

## After this PR

==COMMIT_MSG==
Change error code of timed out async requests to TIMEOUT
==COMMIT_MSG==


carterkozak:
There are some subtle potential races and future result leak cases, I'll take a deeper dive into this when I have a moment.

carterkozak:
This should avoid some future-shuffling and retain the original cancellation stack trace as a cause #1558

westinrm:
yeah that seems nicer

PR #1569, opened 2021-12-07
Add deprecated annotation on all completed stage builder methods

mswintermeyer:
## After this PR
==COMMIT_MSG==
All completed stage builder methods now properly include deprecation annotations, if relevant. Previously, they could be missing for some methods.
==COMMIT_MSG==


carterkozak (approved):
lgtm, thanks for the contribution!

PR #1579, opened 2021-12-15
Expose creating unknown unions and visiting their unknown values

tjb9dc:
## Before this PR
There is no way to propagate unknown union types. The most concrete use case for this functionality is to adapt between API types while preserving unknowns. Currently, you are forced to either throw an error or throw away all the unknown types (assuming you have a corresponding empty type to build on the converted side).

## After this PR

1. Expose a static factory for creating unknown types:
```java
    public static Union unknown(String type, Map<String, Object> value) {
        return new Union(new UnknownWrapper(type, value));
    }
```
2. Add a new method to the visitor interface for the richer unknown visitor, but provide a default implementation which delegates to the existing:
```java
        default T visitUnknownWithValue(String type, Map<String, Object> _value) {
            return visitUnknown(type);
        }
```
3. Plumb it through on the VisitorBuilder

==COMMIT_MSG==
Expose creating unknown unions and visiting their unknown values. This will allow propagating unknown union types, for example.
==COMMIT_MSG==

## Possible downsides?
This changes codegen and adds a new visitor method with a default delegating implementation. While this adds no burden to visitor implementers, it isn't obvious that only the new `visitUnknownWithValue()` method will be called if provided even though `visitUnknown()` must still be implemented. However, the VisitorBuilder is easy to use and enforces the exclusive-or semantics due to the StagedBuilder.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
@@ -16,6 +16,7 @@ import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.IntFunction;
 import javax.annotation.Generated;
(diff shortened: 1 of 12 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/Union.java
```diff
@@ -97,6 +102,10 @@ public String toString() {
 
         T visitUnknown(String unknownType);
 
+        default T visitUnknownWithValue(String type, Map<String, Object> _value) {
```
tjb9dc, line 105: Rename param to `unknownType`?

ferozco (commented):
API change seems reasonable to me. Defer to @carterkozak on the code change

tjb9dc:
Closing in favor of a different approach where we generate a new visitor interface behind a feature flag to avoid muddying the existing visitor interface. Will link the PRs as they come in.

tjb9dc:
First PR: https://github.com/palantir/conjure-java/pull/1605

tjb9dc:
Second PR: https://github.com/palantir/conjure-java/pull/1611

PR #1605, opened 2022-01-19
Static factory for creating unknown union type

tjb9dc:
## Before this PR
There was no way to create an unknown union type via plain Java. Unknown unions were only provided via Jackson deserialization.

## After this PR

==COMMIT_MSG==
Provide a static factory for creating an unknown union type
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
@@ -10,9 +10,12 @@ import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.function.Function;
 import javax.annotation.Generated;
 import javax.annotation.Nonnull;
@@ -31,8 +34,10 @@ public final class EmptyUnionTypeExample {
         return value;
     }
 
-    public static EmptyUnionTypeExample unknown(String type, Map<String, Object> value) {
-        return new EmptyUnionTypeExample(new UnknownWrapper(type, value));
+    public static EmptyUnionTypeExample unknown(String type, Object value) {
+        Preconditions.checkArgument(
+                !allKnownTypes().contains(type), "Unknown type cannot be created as the provided type is known");
+        return new EmptyUnionTypeExample(new UnknownWrapper(type, Collections.singletonMap(type, value)));
     }
 
     public <T> T accept(Visitor<T> visitor) {
@@ -56,6 +61,11 @@ public final class EmptyUnionTypeExample {
     @Override
     public String toString() {
         return "EmptyUnionTypeExample{value: " + value + '}';
+    }
+
+    private static Set<String> allKnownTypes() {
+        Set<String> types = new HashSet<>();
+        return types;
     }
 
     public interface Visitor<T> {
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
@@ -31,6 +31,10 @@ public final class EmptyUnionTypeExample {
         return value;
     }
 
+    public static EmptyUnionTypeExample unknown(String type, Map<String, Object> value) {
+        return new EmptyUnionTypeExample(new UnknownWrapper(type, value));
+    }
+
     public <T> T accept(Visitor<T> visitor) {
         return value.accept(visitor);
     }
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
@@ -12,10 +12,8 @@ import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 import java.util.function.Function;
 import javax.annotation.Generated;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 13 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
```diff
@@ -31,6 +31,10 @@ private Base getValue() {
         return value;
     }
 
+    public static EmptyUnionTypeExample unknown(String type, Map<String, Object> value) {
```
carterkozak: These should end up:
```suggestion
    public static EmptyUnionTypeExample unknown(String type, Object value) {
```

I believe the map is an implementation detail, it should always be a singleton in which the key is the value of `type`.

We should add some validation that `type` doesn't match any of the known types, otherwise it would be possible to create an invalid value for a known type

tjb9dc: Ah nice, so you are imagining:
- Change param to `Object`
- In the body:
  - add a Preconditions check that the type is not known
  - populate the map when calling the UnknownWrapper constructor

carterkozak: Yep, exactly :-)

conjure-java-core/src/integrationInput/java/com/palantir/product/SingleUnion.java
```diff
@@ -61,6 +70,12 @@ public String toString() {
         return "SingleUnion{value: " + value + '}';
     }
 
+    private static Set<String> allKnownTypes() {
+        Set<String> types = new HashSet<>();
+        types.add("foo");
+        return types;
+    }
```
carterkozak: No reason to recreate this for every invocation of `unknown`.

We could take advantage of language features above with something along these lines:

```java
public static SingleUnion unknown(String type, Object value) {
    switch (Preconditions.checkNotNull(type, "Type is required")) {
        case "foo":
            throw new SafeIllegalArgumentException(
                "Unknown type cannot be created as the provided type is known: foo");
        default:
            return new SingleUnion(new UnknownWrapper(type, Collections.singletonMap(type, value)));
    }
}
```

What do you think?

tjb9dc: That sounds good to me! I elected for lazy execution and recomputing over making it an unnecessarily computed static variable, but what you are suggesting is best of both worlds.

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
```diff
@@ -31,6 +32,13 @@ private Base getValue() {
         return value;
     }
 
+    public static EmptyUnionTypeExample unknown(String type, Object value) {
+        switch (Preconditions.checkNotNull(type, "Type is required")) {
+            default:
+                return new EmptyUnionTypeExample(new UnknownWrapper(type, Collections.singletonMap(type, value)));
```
carterkozak, line 38: a little wonky, but probably no need to special case this one as long as it compiles :-)

conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/UnionWithUnknownString.java
```diff
@@ -38,6 +39,16 @@ public static UnionWithUnknownString unknown_(String value) {
         return new UnionWithUnknownString(new Unknown_Wrapper(value));
     }
 
+    public static UnionWithUnknownString unknown(String type, Object value) {
+        switch (Preconditions.checkNotNull(type, "Type is required")) {
+            case "unknown_":
```
carterkozak: We should add a unit test for this -- looks like we're using the java variable name instead of the type value.

conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/UnionWithUnknownString.java
```diff
@@ -38,6 +39,16 @@ public static UnionWithUnknownString unknown_(String value) {
...
+                throw new SafeIllegalArgumentException(
+                        "Unknown type cannot be created as the provided type is known: unknown_");
```
carterkozak: in the message as well

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -173,6 +174,40 @@ private static MethodSpec generateGetValue(ClassName baseClass) {
...
+        MethodSpec.Builder builder = MethodSpec.methodBuilder("unknown")
+                .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+                .addParameter(String.class, typeParam)
+                .addParameter(Object.class, valueParam)
+                .returns(unionClass);
+        // begin switch statement
+        builder.beginControlFlow("switch($L)", Expressions.requireNonNull("type", "Type is required"));
+        // add all cases
+        memberTypeDefs.forEach(memberTypeDef -> {
+            String memberName = sanitizeUnknown(memberTypeDef.getFieldName()).get();
```
carterkozak: This doesn't need to be sanitized since it's the raw wire string

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -173,6 +174,40 @@ private static MethodSpec generateGetValue(ClassName baseClass) {
...
+        builder.beginControlFlow("switch($L)", Expressions.requireNonNull(typeParam, "Type is required"));
+        // add all cases
+        memberTypeDefs.forEach(memberTypeDef -> {
+            String memberName = sanitizeUnknown(memberTypeDef.getFieldName()).get();
+            builder.addCode("case $S:", memberName);
+            builder.addStatement(
+                    "throw new $T(\"$L: $L\")",
+                    SafeIllegalArgumentException.class,
+                    "Unknown type cannot be created as the provided type is known",
+                    memberName);
```
carterkozak: Sorry, I should have caught this before:

Let's let javapoet handle strings for us
```suggestion
                    "throw new $T($S)",
                    SafeIllegalArgumentException.class,
                    "Unknown type cannot be created as the provided type is known: " + memberName);
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -173,6 +174,40 @@ private static MethodSpec generateGetValue(ClassName baseClass) {
...
+        });
+        // add default case, which actually builds the unknown
+        builder.addCode("default:");
+        builder.addStatement(
+                "return new $T(new $T($L, $L))",
```
carterkozak: nit: iirc `$N` represents name references
```suggestion
                "return new $T(new $T($N, $L))",
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -173,6 +174,40 @@ private static MethodSpec generateGetValue(ClassName baseClass) {
...
+                unionClass,
+                wrapperClass(unionClass, FieldName.of("unknown")),
+                typeParam,
+                CodeBlock.of("$T.singletonMap($L, $L)", Collections.class, typeParam, valueParam));
```
carterkozak:
```suggestion
                CodeBlock.of("$T.singletonMap($N, $N)", Collections.class, typeParam, valueParam));
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
tjb9dc: ah makes sense, so we'd want users to call `.unknown("unknown", <thing>)` since `unknown` is the type in the spec

carterkozak (approved):

PR #1611, opened 2022-01-19
Unions with unknown values

tjb9dc:
## Before this PR
No way to retrieve the underlying value of an unknown union type, just the value

## After this PR

==COMMIT_MSG==
Generate (behind a feature flag) union types with a new visitor interface which exposes the unknown value
==COMMIT_MSG==

## Possible downsides?

- A compile break if opted into
- Another feature flag and forks in code to be cleaned up some day


conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
```diff
@@ -59,7 +59,7 @@ public String toString() {
     }
 
     public interface Visitor<T> {
-        T visitUnknown(String unknownType);
+        T visitUnknown(String unknownType, Map<String, Object> unknownValue);
```
carterkozak: Same as the other PR:
```suggestion
        T visitUnknown(String unknownType, Object unknownValue);
```
unknownValue should be the result of `map.get(unknownType)`

conjure-java-core/src/integrationInput/java/com/palantir/product/EmptyUnionTypeExample.java
```diff
@@ -68,18 +78,18 @@ public String toString() {
 
     private static final class VisitorBuilder<T>
             implements UnknownStageVisitorBuilder<T>, Completed_StageVisitorBuilder<T> {
-        private Function<String, T> unknownVisitor;
+        private BiFunction<String, Object, T> unknownVisitor;
 
         @Override
-        public Completed_StageVisitorBuilder<T> unknown(@Nonnull Function<String, T> unknownVisitor) {
+        public Completed_StageVisitorBuilder<T> unknown(@Nonnull BiFunction<String, Object, T> unknownVisitor) {
```
carterkozak, line 79: I think this stage should continue to support `Function<String, T>` in addition to `BiFunction<String, Object, T>` -- that way folks who already use the staged builder aren't broken by setting the feature flag.

The implementation of `unknown(@Nonnull Function<String, T> unknownVisitor)` should delegate to `unknown(@Nonnull BiFunction<String, Object, T> unknownVisitor)`, ignoring the value input.

tjb9dc: that makes sense, let's break as few people as possible

carterkozak (approved):
Nice work :-)

PR #1632, opened 2022-02-07
[undertow-annotations] Add default decoders for header, query, and path params

fawind:
## Before this PR
Users would need to provide a decoder implementation for all their used types.

## After this PR

Provide default decoders for header, query, and path parameters. For now, we support default decoders for the same types as supported by [`PlainSerDe`](https://github.com/palantir/conjure-java/blob/bb07e0e2315fe70640fbc0d0ea7e19a0934b8eff/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/PlainSerDe.java). Deserialization is done by delegating to the respective deserialization method provided by [`PlainSerDe`](https://github.com/palantir/conjure-java/blob/bb07e0e2315fe70640fbc0d0ea7e19a0934b8eff/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/PlainSerDe.java).

For other types, users still have to provide their own implementation of the parameter decoder.

==COMMIT_MSG==
Add default decoders for header, query, and path params
==COMMIT_MSG==


conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/ParamDecoders.java
```diff
@@ -0,0 +1,111 @@
...
+
+import com.google.common.collect.Iterables;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public final class ParamDecoders {
```
fawind, line 34: Instead of doing the abstractions in ParamDecoders, it might be simpler to spend the 30min and add an explicit implementation for all the decoder-types we care about. It's more verbose but probably easier to follow along when looking at the generated code?

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/ParamDecoders.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public static ParamDecoder<String> stringDecoder() {
+        return DelegateDecoder.of(Function.identity());
```
fawind: The actual conversion is quite simple but maybe we want to delegate to [PlainSerde](https://github.com/palantir/conjure-java/blob/6866283be95e05d123af57033e3867b18bda7ed5/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjurePlainSerDe.java) for the types that are supported? Doesn't make the code shorted but that way we would have consistent error handling.

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/ParamDecoders.java
carterkozak: Agreed, that's how we do it in the dialogue implementation: https://github.com/palantir/dialogue/blob/develop/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/DefaultParameterSerializer.java

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/ParamDecoders.java
carterkozak: I like it, we'd exactly match conjure-undertow where possible

carterkozak: I imagine we could get some utility out of these: https://github.com/palantir/conjure-java/blob/6866283be95e05d123af57033e3867b18bda7ed5/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjurePlainSerDe.java#L524-L561

fawind: Updated the default decoders to delegate to the respective method of `PlainSerDe`. Will add the missing methods in a flup commit so we support the same types as `PlainSerDe` out-of-the-box.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
```diff
@@ -98,23 +104,73 @@ public ParamTypesResolver(ResolverContext context) {
...
+        TypeMirror typeMirror = annotationReflector.getAnnotationValue("decoder", TypeMirror.class);
+        if (!context.isSameTypes(typeMirror, DefaultParamDecoder.class)) {
+            return Instantiables.instantiate(typeMirror);
+        }
+
+        TypeMirror variableType = variableElement.asType();
+        String decoderMethodName =
+                DefaultDecoderNames.getDefaultDecoderMethodName(variableType, ContainerType.NONE, ContainerType.NONE);
+
+        // TODO(fwindheuser): We should reference the "runtime" var through a javapoet reference ('$N').
```
carterkozak: Unclear that it would be any better, unless we wanted to rename the variable later, no strong opinion from me :-)

carterkozak (approved):
Looks great, happy to merge as is and clean up later -- up to you

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
fawind: Yeah, we would actually need to funnel the reference down through multiple methods, so skipping on this for now.

PR #1637, opened 2022-02-09
Add staged builder missing overrides

carterkozak:
## Before this PR
Generated code was missing some override annotations

## After this PR
==COMMIT_MSG==
Add staged builder missing overrides
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
@@ -30,6 +30,8 @@ public final class ConjureAnnotations {
 
     private static final ImmutableList<AnnotationSpec> DEPRECATED =
             ImmutableList.of(AnnotationSpec.builder(Deprecated.class).build());
+    private static final ImmutableList<AnnotationSpec> OVERRIDE_LIST =
+            ImmutableList.of(AnnotationSpec.builder(Override.class).build());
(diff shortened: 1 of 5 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
```diff
@@ -50,6 +52,10 @@ public static AnnotationSpec getConjureGeneratedAnnotation(Class<?> clazz) {
         return deprecation.isPresent() ? DEPRECATED : ImmutableList.of();
     }
 
+    public static ImmutableList<AnnotationSpec> override(boolean override) {
```
fawind, line 55: Curious why this is returning a list instead of just the `AnnotationSpec`. We could still use the `MethodSpec.Builder#addAnnotation` builder method which does not require a collection.

fawind (approved):
Looks good!

conjure-java-core/src/main/java/com/palantir/conjure/java/ConjureAnnotations.java
carterkozak: This way we can use the method fluently. If override is false, we'd have to break out of the fluent `.builder().setX(X).build()` flow.

fawind: Didn't think of that but makes sense!

PR #1640, opened 2022-02-09
Add annotation for cookie params

fawind:
## Before this PR

No first-class support for accessing (auth) cookies.

## After this PR

Adds a `@Cookie` annotation for retrieving the cookie value.

```java
@Handle(method = HttpMethod.GET, path = "/cookies")
String cookieParams(
        @Handle.Cookie(value = "my-cookie") String cookieValue,
        @Handle.Cookie(value = "AUTH_TOKEN") BearerToken token);
```

With the exception of annotated parameters of type `BearerToken`, we use the `CookieDeserializer` which internally delegates to `HttpServerExchange#getRequestCookie`.
For an annotated parameter of type `BearerToken`, we instead use the `BearerTokenCookieDeserializer` which delegates to the conjure cookie auth flow (`runtime.auth().cookie()`, [ref](https://github.com/palantir/conjure-java/blob/83671df35de5ba8d0452acfaafc0cc7d64232e42/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAuthorizationExtractor.java#L74-L84)).

==COMMIT_MSG==
Add annotation for cookie params
==COMMIT_MSG==


conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/CookieDeserializer.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public T deserialize(HttpServerExchange exchange) throws IOException {
+        Cookie cookie = exchange.getRequestCookie(cookieName);
+        if (cookie == null) {
+            // Let the decoder handle the null value (e.g. fail or return 'Optional#empty').
+            return decoder.decode(null);
+        }
+        return decoder.decode(cookie.getValue());
```
carterkozak: We try not to allow `null` in our APIs (direct interaction with the undertow apis is a bit different though). What do you think about throwing instead?

decoder taking an Optional would make it a bit harder to interact with conjure-undertow plainserde, but we could add a default `T handleMissing()` which throws.

fawind: Hmm yeah, I forwarded the `null` here to support optional cookie params (i.e. `@Cookie(...) Optional<String>`).

I tried adding a separate Serializer for optional cookies so we can handle the `null` case explicitly. That way we don't really have to touch the decoders. But it's a bit of duplication between the `CookieDeserializer` and `OptionalCookieDeserializer`. Lmk what you think!

carterkozak: Hmm, the additional types+conditionals are a bit odd, I think I'd prefer handling all cookies using the same `CookieDeserializer`. That way we can also support types like `OptionalInt` and alias types around an optional values.

What do you think about adding something like this to `ParamDecoder`?
```java
/** May return a value which should be used if the request doesn't include the requested parameter. */
default Optional<T> noValuePresent() {
    return Optional.empty();
}

carterkozak: To clarify the snippet, returning an empty optional would allow the deserializer calling into the decoder to throw the same exception it does by default.

carterkozak (approved):
Lgtm, thanks!

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/CookieDeserializer.java
fawind: I think that makes sense, didn't think of the `OptionalInt` case before. Pushed up a new change that adds the `noValuePresent` default method to the `ParamDecoder`. For all "optional" param decoders, we would return an `Optional.empty()` here and for all "non-optional" param decoders, we would throw. But happy to iterate!

PR #1643, opened 2022-02-10
Generated hashCode() method does not allocate

schlosna:
## Before this PR
Generated conjure beans' `hashCode()` would use `Objects.hash(Object...)` which would implicitly allocate an `Object[]` for the args

## After this PR

==COMMIT_MSG==
Generated hashCode() method does not allocate
==COMMIT_MSG==


conjure-java-core/src/test/java/com/palantir/conjure/java/types/MethodSpecsTest.java
```diff
@@ -0,0 +1,178 @@
...
+                        + "      int hash = 4441;\n"
+                        + "      hash += (hash << 5) + java.lang.Integer.hashCode(this.primitiveInt);\n"
+                        + "      hash += (hash << 5) + java.util.Objects.hashCode(this.string);\n"
+                        + "      hash += (hash << 5) + java.util.Objects.hashCode(this.boxedInteger);\n"
+                        + "      result = hash;\n"
+                        + "      memoizedHashCode = result;\n"
+                        + "    }\n"
+                        + "    return result;\n"
+                        + "  }\n"
+                        + "}\n");
```
carterkozak: I'd prefer not to print source inline like this because we can't regenerate it with `./gradlew test -Drecreate=true` if the implementation is modified (or javapoet changes its default output format)

conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
```diff
@@ -84,42 +86,71 @@ public static MethodSpec createHashCode(Collection<FieldSpec> fields) {
...
+    static List<CodeBlock> computeHashCode(Collection<FieldSpec> fields) {
+        if (fields.size() == 1) {
+            return ImmutableList.of(computeHashCode(Iterables.getOnlyElement(fields)));
         }
-        return CodeBlock.of("$1T.$2N($3L)", Objects.class, "hash", getHashInput(fields));
+
+        String varName = "hash";
+        ImmutableList.Builder<CodeBlock> builder = ImmutableList.builder();
+        for (FieldSpec field : fields) {
+            builder.add(CodeBlock.of("$N += ($N << 5) + $L", varName, varName, computeHashCode(field)));
```
carterkozak: No strong opinion here, this can be represented using:
```suggestion
            builder.add(CodeBlock.of("$1N += ($1N << 5) + $2L", varName, computeHashCode(field)));
```

conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/PrimitiveOptionalsExample.java
```diff
@@ -263,25 +263,26 @@ private boolean equalTo(PrimitiveOptionalsExample other) {
...
+            hash += (hash << 5) + Objects.hashCode(this.list);
+            hash += (hash << 5) + Objects.hashCode(this.set);
+            hash += (hash << 5) + Objects.hashCode(this.aliasOne);
+            hash += (hash << 5) + Objects.hashCode(this.aliasTwo);
+            hash += (hash << 5) + Objects.hashCode(this.aliasList);
+            hash += (hash << 5) + Objects.hashCode(this.aliasMap);
+            hash += (hash << 5) + Objects.hashCode(this.aliasOptional);
+            hash += (hash << 5) + Objects.hashCode(this.aliasOptionalMap);
+            hash += (hash << 5) + Objects.hashCode(this.aliasOptionalList);
+            hash += (hash << 5) + Objects.hashCode(this.aliasOptionalSet);
```
carterkozak: I'm curious about this hashing algorithm, normally we use something like
```java
hash = 31 * hash + fieldHashCode;
```
Which the jvm will optimize into
```java
hash = (hash << 5) - hash + fieldHashCode;
```

carterkozak: What do you think about retaining the existing hashcode values so we can measure the impact of this change in isolation?

carterkozak: I wonder if we should split out a separate private method to compute the hashcode, keeping this method small

conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
```diff
@@ -84,42 +86,71 @@ public static MethodSpec createHashCode(Collection<FieldSpec> fields) {
...
     }
 
-    private static CodeBlock getHashInput(Collection<FieldSpec> fields) {
-        return CodeBlocks.of(fields.stream().map(MethodSpecs::createHashInput).collect(joining(CodeBlock.of(", "))));
+    @VisibleForTesting
+    static CodeBlock computeHashCode(FieldSpec fieldSpec) {
+        if (fieldSpec.type.isPrimitive()) {
+            return CodeBlock.of("$1T.$2N($3L)", fieldSpec.type.box(), "hashCode", createHashInput(fieldSpec));
+        }
+        return CodeBlock.of("$1T.$2N($3L)", Objects.class, "hashCode", createHashInput(fieldSpec));
```
carterkozak: Given we don't support nulls, we could do something like this:
```suggestion
        return CodeBlock.of("$L.hashCode()", createHashInput(fieldSpec));
```

conjure-java-core/src/test/java/com/palantir/conjure/java/types/MethodSpecsTest.java
schlosna: would you prefer removing this unit test entirely and just rely on the integration tests?

carterkozak: That works for me

conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/PrimitiveOptionalsExample.java
schlosna: I can adjust the hash algorithm. I had initially modeled it after immutables, but keeping consistency with `Objects.hash(Objects...)` & `Arrays.hashCode(Object[])` makes sense.

schlosna: I don't have a strong preference on separate `generateHashCode()` method vs inline as I originally had former then inlined it in https://github.com/palantir/conjure-java/commit/8714f3031f7623e06349152e7ee190e457c3da61

carterkozak: > I don't have a strong preference on separate generateHashCode() method vs inline as I originally had former then inlined it

inline is simpler code, happy to go with that

conjure-java-core/src/test/java/com/palantir/conjure/java/types/MethodSpecsTest.java
schlosna: removed

conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
schlosna: updated

conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/PrimitiveOptionalsExample.java
schlosna: updated

conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
schlosna: updated

carterkozak (approved):

PR #1647, opened 2022-02-14
Pass endpoint to serde factory

fawind:
## Before this PR
In some cases, we want to have access to the endpoint metadata when serializing / deserializing a request.

When discussing this offline, an idea was to modify the [`Serializer#serialize`](https://github.com/palantir/conjure-java/blob/bafb0b6715b410f2d12c20935e92940f271604a1/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/Serializer.java#L26) interface to also accept the `Endpoints` object while adding a default method for backwards compatibility:

```java
public interface Serializer<T> {
    default void serialize(T value, HttpServerExchange exchange) throws IOException {
        serialize(value, exchange, Optional.empty())
    }

    void serialize(T value, HttpServerExchange exchange, Optional<Endpoint> endpoint) throws IOException;
}
```

However couldn't we just modify the `SerializerFactory` to receive an endpoint reference? The factory could then pass the endpoint object to the actual `Serializer` and we would avoid changing the `Serializer` interface. It's okay to break the `SerializerFactory` API because it's unused.

## After this PR

==COMMIT_MSG==
Pass endpoint to serde factory
==COMMIT_MSG==

### DIFF
--- a/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/DefaultSerDe.java
+++ b/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/DefaultSerDe.java
@@ -19,6 +19,7 @@ package com.palantir.conjure.java.undertow.annotations;
 import com.palantir.conjure.java.undertow.lib.BinaryResponseBody;
 import com.palantir.conjure.java.undertow.lib.BodySerDe;
 import com.palantir.conjure.java.undertow.lib.Deserializer;
+import com.palantir.conjure.java.undertow.lib.Endpoint;
 import com.palantir.conjure.java.undertow.lib.Serializer;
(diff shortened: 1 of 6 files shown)


carterkozak:
We definitely don't want to modify the Serializer interface, the conjure-undertow design is based around pre-computing as much as possible when services are initialized so that the work doesn't occur on a per-request basis. For the request-size-limit feature, we'd want to wrap construction of the `Deserializer` when it's created by `BodySerDe` based on tags associated with the `Endpoint`, allowing us to avoid conditionals later at runtime.

fawind:
Redid the PR here (passing the endpoint information to the `BodySerDe` factory methods): https://github.com/palantir/conjure-java/pull/1650

PR #1650, opened 2022-02-14
Provide Endpoints to BodySerde factories

fawind:
## After this PR

`BodySerde` serializer and deserializer factory methods now accept an optional `Endpoint` argument. This allows us to modify construction of the serializer/deserializer based on endpoint metadata.

 In a FLUP PR, we'll also update the annotation-processor codegen and `(De)SerializerFactory` to also forward the endpoint.

==COMMIT_MSG==
Provide Endpoints to BodySerde factories
==COMMIT_MSG==

## Possible downsides?
This is a compile break for consumers that have custom implementations of `BodySerde`.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ImmutableList;
 import com.palantir.conjure.java.undertow.lib.BinaryResponseBody;
 import com.palantir.conjure.java.undertow.lib.BodySerDe;
 import com.palantir.conjure.java.undertow.lib.Deserializer;
+import com.palantir.conjure.java.undertow.lib.Endpoint;
 import com.palantir.conjure.java.undertow.lib.Serializer;
 import com.palantir.conjure.java.undertow.lib.TypeMarker;
 import com.palantir.logsafe.Preconditions;
@@ -39,6 +40,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PushbackInputStream;
 import java.util.List;
+import java.util.Optional;
 import org.xnio.IoUtils;
 
 /** Package private internal API. */
@@ -65,12 +67,22 @@ final class ConjureBodySerDe implements BodySerDe {
 
     @Override
     public <T> Serializer<T> serializer(TypeMarker<T> token) {
-        return new EncodingSerializerRegistry<>(encodings, token);
+        return new EncodingSerializerRegistry<>(encodings, token, Optional.empty());
+    }
+
+    @Override
+    public <T> Serializer<T> serializer(TypeMarker<T> token, Endpoint endpoint) {
+        return new EncodingSerializerRegistry<>(encodings, token, Optional.of(endpoint));
     }
 
     @Override
     public <T> Deserializer<T> deserializer(TypeMarker<T> token) {
-        return new EncodingDeserializerRegistry<>(encodings, token);
+        return new EncodingDeserializerRegistry<>(encodings, token, Optional.empty());
+    }
+
+    @Override
+    public <T> Deserializer<T> deserializer(TypeMarker<T> token, Endpoint endpoint) {
+        return new EncodingDeserializerRegistry<>(encodings, token, Optional.of(endpoint));
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
```diff
@@ -64,12 +66,12 @@
     }
 
     @Override
-    public <T> Serializer<T> serializer(TypeMarker<T> token) {
+    public <T> Serializer<T> serializer(TypeMarker<T> token, Optional<Endpoint> _endpoint) {
         return new EncodingSerializerRegistry<>(encodings, token);
```
fawind: For the request-size-limit feature, we would have this registry wrap the encoding with a "request limiting wrapper"?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/BodySerDe.java
```diff
@@ -19,15 +19,34 @@
 import io.undertow.server.HttpServerExchange;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Optional;
 
 /** Request and response Deserialization and Serialization functionality used by generated code. */
 public interface BodySerDe {
```
fawind: Do we also want to provide an `Endpoint` reference when directly calling the serde methods for binary responses:

https://github.com/palantir/conjure-java/blob/d585a57a01e68b4a99b3374137d630f3b3d0b5e2/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/BodySerDe.java#L58

https://github.com/palantir/conjure-java/blob/d585a57a01e68b4a99b3374137d630f3b3d0b5e2/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/BodySerDe.java#L73

fawind: It would require separate handling from the factory methods but would give us the option to also have different serde behavior based on endpoint metadata?

However, this logic would then be on a per-request basis.

carterkozak: I don't think we need to update those

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/BodySerDe.java
```diff
@@ -19,15 +19,34 @@
...
+    default <T> Serializer<T> serializer(TypeMarker<T> type, Endpoint endpoint) {
+        return serializer(type, Optional.of(endpoint));
+    }
+
+    /** Kept for backwards compatibility. */
+    default <T> Serializer<T> serializer(TypeMarker<T> type) {
+        return serializer(type, Optional.empty());
+    }
+
+    <T> Serializer<T> serializer(TypeMarker<T> type, Optional<Endpoint> endpoint);
```
carterkozak: Let's minimize the API surface area by only adding a new `default <T> Serializer<T> serializer(TypeMarker<T> type, Endpoint endpoint) {` which delegates to the existing method, but without the endpoint arg.

As is this change would be an abi break for any consumer who has implemented their own BodySerDe (which they aren't supposed to do, but I'm aware of several)

fawind: Makes sense, updated.

carterkozak (commented):
API looks good!
We'll need to add similar overloads to Encoding and pass the values (when present) from BodySerDe -> Encoding such that wrapping functionality can make decisions based on the endpoint.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
carterkozak: Yep, via an encoding!

carterkozak (approved):
lgtm, thanks!

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
```diff
@@ -65,12 +67,22 @@
...
 
     @Override
     public <T> Deserializer<T> deserializer(TypeMarker<T> token) {
-        return new EncodingDeserializerRegistry<>(encodings, token);
+        return new EncodingDeserializerRegistry<>(encodings, token, Optional.empty());
+    }
+
+    @Override
+    public <T> Deserializer<T> deserializer(TypeMarker<T> token, Endpoint endpoint) {
+        return new EncodingDeserializerRegistry<>(encodings, token, Optional.of(endpoint));
```
fawind, line 85: Was trying out if it reads nicer of we overload this constructor to avoid the optional wrapping here. However reverted this because it imo makes the flow harder to read and the `EncodingDeserializerRegistry` api is private anyway.

fawind:
Thanks for the pointers here! Will follow up with a PR to also let our annotation-processor generated services forward their endpoints.

PR #1653, opened 2022-02-15
Provide Endpoints to conjure-undertow-annotations serde factories

fawind:
## After this PR

FLUP on https://github.com/palantir/conjure-java/pull/1650.

For generated endpoints generated by the annotation-processor, we also want to provide an `Endpoints` reference to the SerDe factories. 

==COMMIT_MSG==
Provide Endpoints to conjure-undertow-annotations serde factories
==COMMIT_MSG==

## Possible downsides?
Technically an API break but we don't have any consumers yet.

### DIFF
--- a/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/DefaultSerDe.java
+++ b/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/DefaultSerDe.java
@@ -19,6 +19,7 @@ package com.palantir.conjure.java.undertow.annotations;
 import com.palantir.conjure.java.undertow.lib.BinaryResponseBody;
 import com.palantir.conjure.java.undertow.lib.BodySerDe;
 import com.palantir.conjure.java.undertow.lib.Deserializer;
+import com.palantir.conjure.java.undertow.lib.Endpoint;
 import com.palantir.conjure.java.undertow.lib.Serializer;
 import com.palantir.conjure.java.undertow.lib.TypeMarker;
 import com.palantir.conjure.java.undertow.lib.UndertowRuntime;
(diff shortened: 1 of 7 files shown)


conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/DefaultSerDe.java
```diff
@@ -40,22 +41,22 @@
...
-        return runtime.bodySerDe().deserializer(marker);
+        return runtime.bodySerDe().deserializer(marker, endpoint);
     }
 
     @Override
     @SuppressWarnings("unchecked")
-    public <T> Serializer<T> serializer(TypeMarker<T> marker, UndertowRuntime runtime) {
+    public <T> Serializer<T> serializer(TypeMarker<T> marker, UndertowRuntime runtime, Endpoint endpoint) {
         Type type = marker.getType();
         Serializer<T> maybeSpecialSerializer = (Serializer<T>) maybeSpecialSerializer(type, runtime);
```
fawind, line 55: I found it a bit confusing, that for some serde types, we want to provide the endpoints metadata while for others (e.g. `BinaryDeserializer`) we don't (similar to https://github.com/palantir/conjure-java/pull/1650#discussion_r806137596).

Can you elaborate a bit on this, Carter? Using the example of the `request-size-limit feature`, is this generally not wanted for streaming input/output? I could imagine a use-case where we want to wrap the input stream and restrict it's total size.

carterkozak: Of course! There's a distinction between binary and structured deserializers. Binary data is not serialized, and doesn't need to be deserialized, the data is passed directly to the consumer. Secondly, binary data is streamed to the service, where the consumer is responsible for handling the data -- there's no reason a multi-gigabyte request couldn't be handled gracefully in a streaming fashion without risking oom, however if the implementation chooses to buffer, it can also length-check (in contrast to the structured data deserializers which fully abstract away the stream access).

Hopefully that makes some kind of sense

carterkozak (approved):

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/DefaultSerDe.java
fawind: That makes a lot of sense, thanks for clearing that up!

PR #1654, opened 2022-02-15
The undertow annotation processor supports globbed path ends

carterkozak:
==COMMIT_MSG==
The undertow annotation processor supports globbed path ends
==COMMIT_MSG==

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/GlobRetainingDecodingHandler.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/GlobRetainingDecodingHandler.java
@@ -1,0 +1,56 @@
+/*
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.undertow.runtime;
+
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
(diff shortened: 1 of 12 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/GlobRetainingDecodingHandler.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        Map<String, String> pathParams =
+                exchange.getAttachment(PathTemplateMatch.ATTACHMENT_KEY).getParameters();
+        String originalValue = pathParams.get("*");
+        next.handleRequest(exchange);
+        if (originalValue != null) {
+            pathParams.put("*", originalValue);
```
fawind, line 48: Why do we need to put the `originalValue` back into the path params map? Do we expect the map to get mutated during `next.handleRequest(exchange)`?

fawind (approved):
Very nice!

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/GlobRetainingDecodingHandler.java
carterkozak: Yep! This handler wraps the URLDecoding handler. The flow is subtle, and could use a comment or two. 
1. We route _before_ the URL is decoded
2. Path parameters are read by endpoints, _after_ routing.

So, for single path-segment variables, we read the correctly decoded value because it's stored off before routing, the urldecodinghandler transforms it into the decoded value, and it's passed along.
This is trickier for glob-matching because we're matching multiple segments. We can no longer make the assumption that the value represents one url segment. We may have a url string like `/foo/bar%2Fbaz` which would decode into `/foo/bar/baz`, so given `/foo/bar/baz` we cannot tell what the original value was. By restoring the encoded value, but only for globbed paths, we can split it into segments when requested, and decode each individual segment. The example above would decode into `['foo', 'bar/baz']`. By providing the values as a list, we solve the ambiguity problem that strings cause, and match the behavior we defined in dialogue.

carterkozak: I'll add comments in a follow-up, there's some other cleanup I'd like to take care of after this merges

fawind: That actually makes sense. Thanks for elaborating!

PR #1664, opened 2022-02-18
[WIP] Add content type filtering annotation

fawind:
## Before this PR

For jersey endpoints, it is quite common to restrict the accepted content types using:

```java
@Consumes(MediaType.TEXT_PLAIN)
void myEndpoint(String body);
```

To support the same behavior using the annotation processor, you would have to provide your own serializer that checks the header and throws the correct exception / error code.

## After this PR

**Work in progress, looking for some feedback:**

Ideally we would want to reuse the [`FrameworkExceptions#unsupportedMediaType` exception](https://github.com/palantir/conjure-java/blob/ed6b1ada6058ee91eb1966d966047dd2100c21bc/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/FrameworkException.java#L53-L55). However this class is package private to the `runtime` package.

To still use this, I currently work around this by defining a filter in the `c-j-undertow-runtime` project that the processor adds the the generated handler. However this extends the API surface of `c-j-undertow-runtime` with something we only use in the processor which I don't like. We also add a direct dependency on `c-j-undertow-runtime`.

We could also match the package names to access it but this also feels hacky and adds the additional dependency.

==COMMIT_MSG==
Add content type filtering annotation
==COMMIT_MSG==


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ContentTypeFilterHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private ContentTypeFilterHandler(Set<String> acceptedContentTypes, HttpHandler delegate) {
+        this.acceptedContentTypes = acceptedContentTypes;
+        this.delegate = delegate;
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        String contentType = ContentTypes.getContentType(exchange);
+        if (acceptedContentTypes.contains(contentType)) {
```
carterkozak: This is a bit tricky because a json consumer may consume `application/json` or `application/json; charset=UTF-8` or `application/json;charset=UTF-8`, even `application/json; foo=bar; baz=bang`

conjure-undertow-processor/src/test/java/com/palantir/conjure/java/undertow/processor/sample/ContentTypeFilter.java
```diff
@@ -0,0 +1,32 @@
...
+
+import com.palantir.conjure.java.undertow.annotations.Handle;
+import com.palantir.conjure.java.undertow.annotations.HttpMethod;
+import javax.ws.rs.core.MediaType;
+
+public interface ContentTypeFilter {
+
+    @Handle(method = HttpMethod.POST, path = "/contentTyp")
+    @Handle.Accepts(MediaType.TEXT_PLAIN)
+    void contentType(@Handle.Body String body);
```
carterkozak: This is a bit awkward because the `Handle.Body(deserializer=<VALUE>)` needs to agree with the `Accepts` values, but we don't have anything in place which forces that to be true. So in this case we'd fail unless we receive `text/plain` content, but the body default deserializer would fail to accept most input because it requires json, smile, or cbor.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ContentTypeFilterHandler.java
fawind: Hmm yeah also wildcards. We could re-use the existing logic for matching those but this seems quite tricky:

https://github.com/palantir/conjure-java/blob/88589c7ba2ff0eeec6c4638d8b347bbaad8a13ef/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java#L133-L134

https://github.com/palantir/conjure-java/blob/8afc1c6c92a33c03eff4d308ee31f855fe017e8b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Encodings.java#L49

conjure-undertow-processor/src/test/java/com/palantir/conjure/java/undertow/processor/sample/ContentTypeFilter.java
fawind: Agree. The common use-cases I have seen so far were mostly InputStream or form data for which we afaik don't check the encoding in the default serializer.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ContentTypeFilterHandler.java
carterkozak: Ya, I could see a case for a few different utility classes, something like this could be fairly generic (with no support for consuming an optional request body, but I think that's fine, it's rare and can be handled by implementing a DeserializerFactory+Deserializer)


```java
public interface SimpleDeserializer<U> extends DeserializerFactory<U> {

    boolean accepts(String contentType);

    U deserialize(InputStream content) throws IOException;

    @Override
    default <T extends U> Deserializer<T> deserializer(TypeMarker<T> type, UndertowRuntime runtime, Endpoint endpoint) {
        return (Deserializer<T>) new Deserializer<U>() {
            @Override
            public U deserialize(HttpServerExchange exchange) throws IOException {
                String maybeContentType = exchange.getRequestHeaders().getFirst(Headers.CONTENT_TYPE);
                if (maybeContentType == null || !accepts(maybeContentType)) {
                    throw new SafeIoException("Unsupported content-type", SafeArg.of("contentType", maybeContentType));
                }
                return SimpleDeserializer.this.deserialize(exchange.getInputStream());
            }
        };
    }
}
```

we could provide a subclass which returns a content-type, where we do a best-effort match (`accepts` does the `startsWith` check)
The Serializer is a bit more complicated since there's some amount of negotiation based on the `Accept` value, but that can largely be ignored in most cases because it doesn't matter if we fail based on a missing Accept header, or write data that causes the client to fail -- in either case the client must also validate the Content-Type of the response.

fawind: I like the idea of just providing a utility class that consumers can use. Updated the PR based on this suggestion.

Example for how an implementation might look like:
```java
enum WildcardDeserialzerFactory implements ContentTypeAcceptingDeserializerFactory<InputStream> {
    INSTANCE;

    @Override
    public List<String> supportedContentType() {
        return ImmutableList.of(MediaType.WILDCARD);
    }

    @Override
    public InputStream deserialize(HttpServerExchange exchange, String _contentType) {
        return exchange.getInputStream();
    }
}
```

fawind: We have some internal use cases can consume multiple content types:
```java
@Consumes({ MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML })
```

Updated the utils class to provide a list of accepted content types instead.

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/ContentTypeAcceptingDeserializerFactory.java
```diff
@@ -0,0 +1,59 @@
...
+    /** The content types that is accepted by this deserializer. */
+    List<String> supportedContentTypes();
+
+    /** Deserialize the request body with the given supported content type. */
+    U deserialize(HttpServerExchange exchange, String contentType) throws IOException;
+
+    /** Per default, a content type is accepted if its value starts with the supported content type. */
+    default boolean accepts(String contentType, String supportedContentType) {
+        return contentType.startsWith(supportedContentType);
+    }
```
carterkozak, line 41: I'm not sure I'd expose this (not yet anyhow). It's easier to expose later than to remove.

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/ContentTypeAcceptingDeserializerFactory.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.IOException;
+import java.util.List;
+
+public interface ContentTypeAcceptingDeserializerFactory<U> extends DeserializerFactory<U> {
+
+    /** The content types that is accepted by this deserializer. */
+    List<String> supportedContentTypes();
+
+    /** Deserialize the request body with the given supported content type. */
+    U deserialize(HttpServerExchange exchange, String contentType) throws IOException;
```
carterkozak, line 36: Since we're handling the `Accept` header, we can make this more specific (decoupling implementations from conujre-undertow) with something like:
```java
U deserialize(InputStream requestBody, String contentType) throws IOException;
```

PR #1667, opened 2022-02-22
[processor] Add support for safe logging annotations

fawind:
## Before this PR
There was no way to explicitly mark a parameter as safe or unsafe.

## After this PR

When a parameter is annotated as `@Safe` or `@Unsafe` we generate the respective marking into the handler:

```java
@Override
public void handleRequest(HttpServerExchange exchange) throws Exception {
    RequestContext context = this.runtime.contexts().createContext(exchange, this);
    String safeParam = this.safeParamDeserializer.deserialize(exchange);
    context.requestArg(SafeArg.of("safeParam", safeParam));
    String unsafeParam = this.unsafeParamDeserializer.deserialize(exchange);
    context.requestArg(UnsafeArg.of("unsafeParam", unsafeParam));
    write(this.delegate.queryParams(safeParam, unsafeParam), exchange);
}
```

We currently don't support marking the body (we ignore the annotation) as well as `AuthHeader` and `BearerToken` params (we throw an exception) as safe (or unsafe).

==COMMIT_MSG==
Add support for safe logging annotations
==COMMIT_MSG==


carterkozak:
Another option is to match conjure-java safe/unsafe handling using something like this:
https://github.com/palantir/conjure-java/blob/e1991a0a09d106ad9d5bbd13e3f010e9674d8c9a/conjure-java-core/src/test/resources/test/api/TestServiceEndpoints.java.undertow#L399

Slightly less conditionality (don't have to create a reusable context), however I think the context is easier code to follow.

carterkozak:
> Should we always mark AuthHeader and BearerToken as @ Unsafe (probably fine to leave them unmarked)?

There are really three safety classifications: Safe, Unsafe, and Unloggable. Credential material falls in the third category, if it's logged (even as unsafe) that's a serious security flaw!

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
```diff
@@ -222,11 +224,17 @@ public Void path(String paramName, String deserializerFieldName, CodeBlock deser
                                         deserializerFactory)
                                 .build())
                         .build());
+                handlerBuilder.addStatement("$T $N = $L", paramType, paramName, invokeDeserializer(def));
```
carterkozak, line 227: We need to be a bit careful here, if a param is named `exchange`, it will clash with the `exchange` parameter. Worth adding a test for this sort of thing

fawind: We could rename the variable name to something like `requestContext_` which clashes with checkstyle and should usually not get used as a param name.

fawind: Added a test here: https://github.com/palantir/conjure-java/pull/1667/commits/b5672d42bf6a38ab843aa39bd21a246a9b8744cd

fawind:
> > Should we always mark AuthHeader and BearerToken as @ Unsafe (probably fine to leave them unmarked)?

> There are really three safety classifications: Safe, Unsafe, and Unloggable. Credential material falls in the third category, if it's logged (even as unsafe) that's a serious security flaw!

Makes sense! Updated the preconditions so that we also error out on something like `[@Unsafe|@Safe] AuthHeader`.

conjure-undertow-processor/src/test/java/com/palantir/conjure/java/undertow/processor/sample/ContextParamNameClash.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.palantir.conjure.java.undertow.annotations.Handle;
+import com.palantir.conjure.java.undertow.annotations.HttpMethod;
+import com.palantir.conjure.java.undertow.lib.RequestContext;
+
+public interface ContextParamNameClash {
+
+    @SuppressWarnings("checkstyle:ParameterName")
+    @Handle(method = HttpMethod.GET, path = "/context")
+    String context(@Handle.QueryParam("requestContext") String requestContext_, RequestContext context);
```
carterkozak: Worth testing something like this as well:

```java
    @Handle(method = HttpMethod.GET, path = "/{exchange}")
    String exchange(@Safe @Handle.PathParam String exchange);
```

fawind: Yep, this will also fail. Should we also rename our internal `HttpServerExchange` to `exchange_` to avoid clashes in most cases?

fawind: Updated the internal name to `exchange_` and added a failing test.

carterkozak: I could go either way. On one hand, using the `exchange` name is nice for readability, and we can rename parameters around it for consistency if they happen to crash. On the other hand, that's more work.

fawind: Let's keep exchange as the internal param name for now. It's trivial and safe to change once someone runs into this problem but maybe we'll just never encounter it. Updated.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParameterTypeVisitors.java
```diff
@@ -0,0 +1,180 @@
...
+import com.squareup.javapoet.CodeBlock;
+
+public final class ParameterTypeVisitors {
+
+    public enum UsesRequestContextVisitor implements Cases<Boolean> {
+        INSTANCE;
+
+        @Override
+        public Boolean body(CodeBlock _deserializerFactory, String _deserializerFieldName) {
+            return false;
```
carterkozak, line 30: fwiw this isn't supported in jersey, but I think we do allow it in conjure-undertow

carterkozak (approved):
lgtm :-)

PR #1669, opened 2022-02-22
Test coverage for missing annotations

carterkozak:
Also removes `Optional` return type from implementation
methods which always return a present value.

==COMMIT_MSG==
Test coverage for missing annotations
==COMMIT_MSG==


carterkozak:
I should have probably created two separate PRs for this

fawind (approved):

PR #1670, opened 2022-02-22
Support exclusion of empty collections using `experimentalExcludeEmptyCollections`

carterkozak:
This matches empty optional exclusion, including discovery
and remediation of a bug:

Aliases may be excluded, however they were not handled correctly
in the other direction, where null values were handled correctly.
Now we initialize builder instances with a reference to an empty
instance of alias types. This includes a new `public static T empty()`
method on aliases which can handle defaults.

The current implementation only checks one level of alias indirection,
so alias-of-alias-of-optional/collection still won't do quite what we
want, but it's a start. Eventually we should consolidate this code.

### Staged Builders

Note that this change could impact staged builders which consider `alias<optional/collection>` a required field, however such a change would be an abi break. Staged builders are not impacted by this change.

Non-staged builders will no longer throw when `alias<optional/collection>` are not explicitly set, matching the behavior of an optional/collection field.

## After this PR
==COMMIT_MSG==
Support exclusion of empty collections using `experimentalExcludeEmptyCollections`
==COMMIT_MSG==

## Possible downsides?
Super duper dangerous if this is set on servers because our typescript clients do not fully implement conjure.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/SetAlias.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/SetAlias.java
@@ -10,6 +10,8 @@ import javax.annotation.Nonnull;
 
 @Generated("com.palantir.conjure.java.types.AliasGenerator")
 public final class SetAlias {
+    private static final SetAlias EMPTY = new SetAlias();
+
     private final Set<String> value;
 
     private SetAlias(@Nonnull Set<String> value) {
@@ -43,5 +45,9 @@ public final class SetAlias {
     @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
     public static SetAlias of(@Nonnull Set<String> value) {
         return new SetAlias(value);
+    }
+
+    public static SetAlias empty() {
+        return EMPTY;
     }
 }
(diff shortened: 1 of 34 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/SetAlias.java
```diff
@@ -44,4 +46,8 @@ public int hashCode() {
     public static SetAlias of(@Nonnull Set<String> value) {
         return new SetAlias(value);
     }
+
+    public static SetAlias empty() {
```
carterkozak, line 50: I also considered `of()` rather than `empty()`, matching empty object factory method, but I think this is more descriptive.

fawind: Yeah, no strong preference myself. `empty()` sounds good!

fawind:
Current state looks reasonable to me! 👍

fawind:
> Possible downsides?
> Super duper dangerous if this is set on servers because our typescript clients do not fully implement conjure.

Linking the issue for future reference because I think its part of the reason here? https://github.com/palantir/conjure-typescript/issues/78

PR #1698, opened 2022-03-08
Allow cross-origin requests

pkoenig10:
## Before this PR
Conjure endpoints do not support cross-origin requests.

## After this PR
Conjure endpoints do not support cross-origin requests. Supporting cross origin requests OOTB feels like a reasonable design choice for an API server.

When a CORS preflight request is made, the response now includes the following response headers:
- [`Access-Control-Allow-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers.html) is set to `Authorization`
- [`Access-Control-Allow-Methods`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods.html) is set to the list of methods supported for that path
- [`Access-Control-Allow-Origin`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin.html) is set to the value of the `Origin` request header, if present
- [`Access-Control-Max-Age`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age.html) is set to 600 seconds
- [`Vary`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary.html) is set to `Origin` because the `Access-Control-Allow-Origin` response header varies based on the origin

For details see the [MDN documentation for CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).

## Security considerations

- What about [`Access-Control-Expose-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers.html)?

    API endpoints generally do not return headers that need to be read by clients. If this changes in the future, we can revisit this decision.

- What about [`Access-Control-Allow-Credentials`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials.html)?

    We definitely do not want to allow credentials to be sent in cross-origin requests. If we do, endpoints that use cookie authentication will be vulnerable to CSRF.

    An explicitly set `Authorization` request header is still allowed even without `Access-Control-Allow-Credentials`. `Access-Control-Allow-Credentials` controls [request credentials](https://fetch.spec.whatwg.org/#credentials), which only includes things that send credentials _implicitly_,  such as cookies, TLS client certificates, and [HTTP authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication) (which may use the `Authorization` request header). An _explicitly_ set `Authorization` request header doesn't have the same CSRF risks and is not blocked by CORS.

- What about request headers other than the `Authorization` header?

    Request headers other than the `Authorization` header are not supported.

    We could support arbitrary request headers by setting the `Access-Control-Allow-Headers` response header to the value of the `Access-Control-Request-Headers` value. I don't see any obvious security concerns with doing so ([this SO answer](https://stackoverflow.com/a/45045181) seems to confirm that there's no specific security-related reason for restricting request headers), but I biased towards security in this initial attempt.

    Most API endpoints do not use custom request headers, especially those that should be exposed to third-party developers making cross-origin requests. So the cost of not supporting custom header parameters is likely small. As always, we can revisit this decision in the future.

### DIFF
--- a/changelog/@unreleased/pr-1698.v2.yml
+++ b/changelog/@unreleased/pr-1698.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Conjure Undertow endpoints now support cross-origin requests.
+  links:
+  - https://github.com/palantir/conjure-java/pull/1698
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/OptionsHandler.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/OptionsHandler.java
@@ -18,6 +18,7 @@ package com.palantir.conjure.java.undertow.runtime;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.net.HttpHeaders;
 import com.palantir.logsafe.Preconditions;
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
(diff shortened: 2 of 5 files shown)


pkoenig10:
Reopening this PR following internal discussions regarding the urgency of supporting cross-origin requests. This PR is the agreed upon short-term solution.

tpetracca:
We've decided to move forward with a different tactic at the network-layer/frontdoor of our environments that is strictly safer from a security perspective

PR #1705, opened 2022-03-11
Support Intellij Gradle integration

pkoenig10:
## Before this PR
The configuration of the `integrationInput` source set does not work with the Intellij Gradle integration. Intellij fails to find `integrationInput` classes when imported in `test` classes.

Screen Shot 2022-03-11 at 9 03 15 AM

When using the IDEA plugin, each subproject is a single module. So the integrationInput and test source sets are in test source directories of the same module and everything works.

Screen Shot 2022-03-11 at 9 46 48 AM

When using the Gradle integration, each sources set is a separate module.

Screen Shot 2022-03-11 at 9 48 04 AM
Screen Shot 2022-03-11 at 9 48 09 AM

So although the `test` module depends on the `integrationInput` module, it's depending on the "sources" of the `integrationInput` module, not the "test sources".

Screen Shot 2022-03-11 at 9 48 15 AM

In my testing, it appears that the Intellij Gradle integration uses the IDEA plugin configuration to inform how the project is structured. Even when using the Gradle integration, changing `testSourceDirs` to `sourceDirs` changed how the integration input source directory was configured in Intellij.

## After this PR
- The `integrationInput` source is now marked as "sources" instead of "test sources".
- The `integrationInput` dependency is now declared using standard Gradle dependency declarations instead of manual classpath modifications.

## Possible downsides?
- When using the IDEA plugin it will now be possible to import `integrationInput` classes. But this will not compile in CI because there is no Gradle dependency.

### DIFF
--- a/conjure-java-core/build.gradle
+++ b/conjure-java-core/build.gradle
@@ -18,10 +18,6 @@ apply plugin: 'com.palantir.external-publish-jar'
 
 sourceSets {
     integrationInput
-    test {
-        compileClasspath += integrationInput.output
-        runtimeClasspath += integrationInput.output
-    }
 }
 
 spotless {
@@ -32,7 +28,7 @@ spotless {
 
 idea {
     module {
-        testSourceDirs += sourceSets.integrationInput.java.srcDirs
+        sourceDirs += sourceSets.integrationInput.java.srcDirs
     }
 }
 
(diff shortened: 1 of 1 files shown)


carterkozak (approved):
thanks!

PR #1706, opened 2022-03-16
The conjure-undertow processor retains source element deprecation

carterkozak:
This allows deprecation linting to work as expected without failing
due to non-deprecated generated code calling into deprecated
types.

==COMMIT_MSG==
The conjure-undertow processor retains source element deprecation
==COMMIT_MSG==


fawind (approved):
Looks good!

PR #1708, opened 2022-03-16
conjure-undertow processor allows overloaded `@Handle` methods

carterkozak:
==COMMIT_MSG==
conjure-undertow processor allows overloaded `@Handle` methods
==COMMIT_MSG==

## Possible downsides?
This will simplify migration from Jersey, however our metrics often
assume service-name + endpoint-name pairs are unique, so per-endpoint
metric data may not be as nice as it could be. It's probably fair
to assume overloaded methods should be analyzed together though.


fawind (approved):

carterkozak:
Rebasing and regenerating since the generated code is likely different after the body safe-logging change merged

PR #1712, opened 2022-03-18
Handle query/header/path `Optional<Long>` and `OptionalLong`

carterkozak:
This uses the same approach as conjure primitive types as opposed
to special methods, this is because there are only three special
optional types, the other two of which were already handled
using `ParamDecoders`.

==COMMIT_MSG==
Handle query/header/path `Optional<Long>` and `OptionalLong`
==COMMIT_MSG==

### DIFF
--- a/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/ParamDecoders.java
+++ b/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/ParamDecoders.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.java.undertow.annotations;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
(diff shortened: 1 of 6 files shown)


conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/ParamDecoders.java
```diff
@@ -273,7 +305,7 @@
         return DelegatingCollectionParamDecoder.of(value -> serde.deserializeOptionalComplex(value, factory));
     }
 
-    public <T> CollectionParamDecoder<List<T>> complexListCollectionParamDecoder(
+    public static <T> CollectionParamDecoder<List<T>> complexListCollectionParamDecoder(
```
carterkozak, line 308: Caught this bug when I tried to delegate to this method :-)

fawind (approved):

PR #1723, opened 2022-03-24
annotation-processor uses `of(String)` factory methods by default

carterkozak:
This is least preferred, after `valueOf(String)` and a single-string
constructor.

==COMMIT_MSG==
annotation-processor uses `of(String)` factory methods by default
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1723.v2.yml
+++ b/changelog/@unreleased/pr-1723.v2.yml
@@ -1,5 +1,5 @@
 type: improvement
 improvement:
-  description: annotation-processor uses `of(String)` factory methods by default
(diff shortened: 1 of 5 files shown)


changelog/@unreleased/pr-1723.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: annotation-processor uses `of(String)` factory methods by default
```
carterkozak:
```suggestion
  description: annotation-processor uses `of(String)` factory methods by default.
```

fawind (approved):

PR #1749, opened 2022-04-05
Limit conjure codegen concurrency to the expected parallelism

carterkozak:
This was the expected behavior, however `MoreStreams` has a bug
which allows greater concurrency than expected when the underlying
stream is constructed using flatMap.

==COMMIT_MSG==
Limit conjure codegen concurrency to the expected parallelism
==COMMIT_MSG==


carterkozak:
Closing, this will be resolved for real via https://github.com/palantir/streams/pull/36

PR #1753, opened 2022-04-11
Implement basic safety features

carterkozak:
## Before this PR
No true concept of log-safety in conjure beyond a brief mention in ErrorDefinition.

==COMMIT_MSG==
Implement support for conjure log-safety
==COMMIT_MSG==

## Possible downsides?
Complexity.

It is possible that some consumers use enums in unsafe ways, with known unsafe data in enum unknown variants. This change considers all enums `safe` including unsafe variants because they're meant for cross-version compatibility, not completely unknown data.

Similarly, we consider all unknown union variants `unsafe` in all cases, regardless of the guarantees within any given API. I considered leaving Unions unmarked initially until we've rolled this feature out a bit, but the flip will likely result in some churn whenever we decide to do it.


build.gradle
```diff
@@ -49,6 +50,7 @@ allprojects {
 
     repositories {
         mavenCentral() { metadataSources { mavenPom(); ignoreGradleMetadataRedirection() } }
+        mavenLocal() { metadataSources { mavenPom(); ignoreGradleMetadataRedirection() } }
```
carterkozak: These will be removed once we release the conjure change and update this branch

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
```diff
@@ -0,0 +1,285 @@
...
+
+        @Override
+        public Optional<LogSafety> visitUnknown(String unknownType) {
+            throw new IllegalStateException("Unknown type: " + unknownType);
+        }
+
+        private Optional<LogSafety> with(TypeName typeName, Supplier<Optional<LogSafety>> task) {
+            if (!inProgress.add(typeName)) {
+                // Given recursive evaluation, we return the least restrictive type: SAFE.
+                return Optional.of(LogSafety.SAFE);
```
fawind, line 127: Not sure I follow here. Why is the inProgress tracking needed here? I don't see the case yet, where this where we start processing a type name that is already in progress?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
```diff
@@ -0,0 +1,285 @@
...
+            }
+            Optional<LogSafety> result = task.get();
```
fawind, line 129: Would it make sense to cache the computed safety value for a type as this should all be statically known? But maybe not worth the effort if its fast enough already?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
carterkozak: That's correct, I may define something along these lines (although in practice they tend to have a great deal more indirection)

```yaml
MyNode:
  fields:
    foo:
      type: integer
      safety: unsafe
    next: optional<MyNode>
```

We need to avoid infinite recursion in this case

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
carterkozak: Definitely, I started writing it, but decided it was more complex than necessary for a first iteration. I figure we can always add it later if this turns out to have a measurable cost at build-time.
Note the cache would need to avoid caching any nested evaluations which include a cycle-detection `Optional.of(LogSafety.SAFE)` value, since that only guarantees correctness at the top level.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
fawind: Ah, I see. Because `task.get()` will evaluate the whole subtree which might recurse. Thanks!

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
fawind: Good flag regarding that edge case! Yeah, fine to leave it like this and optimize later if required!

fawind (approved):
Looks sane and think the added Conjure API works as proposed!

PR #1756, opened 2022-04-14
Validate safety agreement between existing arguments and types

carterkozak:
Ideally validation would occur within the conjure parser, not
at codegen time, and it certainly will for new argument-specific
safety fields, however the marker and tag handling is older than
new safety features, and requires us to fully evaluate safety
of types.

## After this PR
==COMMIT_MSG==
Validate safety agreement between existing argument markers/tags and types
==COMMIT_MSG==

I've also included a change to consolidate handling of legacy `Marker` values with safety `tag` values, where weren't supported the same way across the board.

## Possible downsides?
Could fail to upgrade, but that's signal that something is very wrong.

### DIFF
--- a/conjure-java-core/src/test/resources/test/api/TestServiceBlocking.java.dialogue.prefix
+++ b/conjure-java-core/src/test/resources/test/api/TestServiceBlocking.java.dialogue.prefix
@@ -60,33 +60,33 @@ public interface TestServiceBlocking {
      * @apiNote {@code POST /catalog/datasets}
      */
     @ClientEndpoint(method = "POST", path = "/catalog/datasets")
-    Dataset createDataset(AuthHeader authHeader, String testHeaderArg, CreateDatasetRequest request);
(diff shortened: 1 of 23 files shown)


conjure-java-core/src/test/resources/test/api/TestServiceBlocking.java.dialogue.prefix
```diff
@@ -60,33 +60,33 @@ public interface TestServiceBlocking {
      * @apiNote {@code POST /catalog/datasets}
      */
     @ClientEndpoint(method = "POST", path = "/catalog/datasets")
-    Dataset createDataset(AuthHeader authHeader, String testHeaderArg, CreateDatasetRequest request);
+    Dataset createDataset(AuthHeader authHeader, @Safe String testHeaderArg, CreateDatasetRequest request);
```
carterkozak, line 63: We previously didn't include safety from markers, only tags! I don't think that was intentional

fawind (approved):
Looks good!

PR #1787, opened 2022-05-09
conjure-undertow argument safety is evaluated taking tags into account

carterkozak:
## Before this PR
This fixes a regression in which the argument declared safety and marker safety were used, but tag safety was ignored.

## After this PR
==COMMIT_MSG==
conjure-undertow argument safety is evaluated taking tags into account
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -416,7 +416,8 @@ final class UndertowServiceHandlerGenerator {
     private static final String QUERY_PARAMS_VAR_NAME = "queryParams";
     private static final String HEADER_PARAMS_VAR_NAME = "headerParams";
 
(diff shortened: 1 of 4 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -428,7 +429,7 @@ private static boolean isSafeLoggableArgument(ArgumentDefinition argument, Safet
                 // Allow opt-out for noisy parameters
                 && ConjureTags.isServerSafeLoggingAllowed(argument)
                 && safetyEvaluator
-                        .evaluate(argument.getType(), ConjureMarkers.markerSafety(argument))
+                        .evaluate(argument.getType(), ConjureTags.safety(argument))
```
carterkozak, line 432: `ConjureTags.safety` delegates to `ConjureMarkers.markerSafety`

tpetracca (approved):

PR #1790, opened 2022-05-11
Ignore decoder factory functions that throw checked exceptions

pkoenig10:
## Before this PR
If I attempt to declare an endpoint with a parameter that has a constructor/static factory method that throws a checked exception, the generated code will use that constructor/static factory method but the generated code will fail to compile.

```java
final class MyService {

    @Handle(method = HttpMethod.GET, path = "/uri")
    public void uri(@QueryParam("uri") URI uri) {}
}
```

```
MyServiceEndpoints.java:50: error: incompatible thrown types URISyntaxException in functional expression
                    "uri", ParamDecoders.complexCollectionParamDecoder(runtime.plainSerDe(), URI::new));
```

## After this PR
When finding decoders for unknown types, constructors and static factory methods are only considered if they do not throw checked exceptions.

### DIFF
--- a/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/DefaultDecoderNames.java
+++ b/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/DefaultDecoderNames.java
@@ -41,6 +41,7 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 
 final class DefaultDecoderNames {
@@ -118,15 +119,15 @@ final class DefaultDecoderNames {
                 return Optional.of(CodeBlock.of("$T::parseShort", Short.class));
             case DECLARED:
                 DeclaredType declaredType = (DeclaredType) typeMirror;
-                return getValueOfDecoderFactoryFunction(declaredType, "valueOf")
-                        .or(() -> getStringConstructorDecoderFactoryFunction(declaredType))
-                        .or(() -> getValueOfDecoderFactoryFunction(declaredType, "of"));
+                return getFactoryDecoderFactoryFunction(declaredType, "valueOf")
+                        .or(() -> getConstructorDecoderFactoryFunction(declaredType))
+                        .or(() -> getFactoryDecoderFactoryFunction(declaredType, "of"));
             default:
                 return Optional.empty();
         }
     }
 
-    private static Optional<CodeBlock> getValueOfDecoderFactoryFunction(DeclaredType declaredType, String methodName) {
+    private static Optional<CodeBlock> getFactoryDecoderFactoryFunction(DeclaredType declaredType, String methodName) {
         TypeElement typeElement = (TypeElement) declaredType.asElement();
         // T valueOf(String)
         return typeElement.getEnclosedElements().stream()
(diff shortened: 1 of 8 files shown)


conjure-undertow-processor/src/test/java/com/palantir/conjure/java/undertow/processor/sample/DefaultDecoderService.java
```diff
@@ -32,49 +31,55 @@
...
-            path = "/pathParam/{stringParam}/{booleanParam}/{decoderParam}/{floatBoxed}/{floatUnboxed}/{bigInt}")
+            path = "/pathParam/{stringParam}/{booleanParam}/{decoderParam}/{floatBoxed}/{floatUnboxed}/{bigInt}"
+                    + "/{constructor}/{ofFactory}/{valueOfFactory}")
     String pathParam(
             @Handle.PathParam String stringParam,
             @Handle.PathParam Boolean booleanParam,
             @Handle.PathParam(decoder = StringDecoder.class) String decoderParam,
             @Handle.PathParam Float floatBoxed,
             @Handle.PathParam float floatUnboxed,
-            @Handle.PathParam BigInteger bigInt);
```
pkoenig10, line 77: Changed this to use custom inner types so it's clear what is being tested here (`valueOf` factory methods).

Also added parameters that use constructors and `of` factory methods to ensure those are tested as well.

conjure-undertow-processor/src/test/java/com/palantir/conjure/java/undertow/processor/sample/DefaultDecoderService.java
```diff
@@ -32,49 +31,55 @@
...
     String queryParam(
-            @Handle.QueryParam(value = "stringParam") String stringParam,
-            @Handle.QueryParam(value = "booleanParam") Boolean booleanParam,
-            @Handle.QueryParam(value = "stringSetParam") Set<String> stringSetParam,
-            @Handle.QueryParam(value = "stringListParam") List<String> stringListParam,
-            @Handle.QueryParam(value = "optionalStringParam") Optional<String> optionalStringParam,
-            @Handle.QueryParam(value = "decoderParam", decoder = StringCollectionDecoder.class) String decoderParam);
-
-    @Handle(method = HttpMethod.GET, path = "/moreQueryParams")
-    String moreQueryParams(
```
pkoenig10, line 47: Suppressed the `TooManyArguments` so we don't have to awkwardly split the query params endpoint.

conjure-undertow-processor/src/test/java/com/palantir/conjure/java/undertow/processor/sample/ParameterOfFactoryThrows.java
```diff
@@ -17,27 +17,20 @@
 package com.palantir.conjure.java.undertow.processor.sample;
 
 import com.palantir.conjure.java.undertow.annotations.Handle;
+import com.palantir.conjure.java.undertow.annotations.Handle.QueryParam;
 import com.palantir.conjure.java.undertow.annotations.HttpMethod;
 
-public interface OfFactory {
```
pkoenig10, line 22: This tests has been moved into `DefaultDecoderService`

fawind (approved):
Makes sense to filter this out instead of failing on compile. Thanks for cleaning up the tests!

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/DefaultDecoderNames.java
```diff
@@ -216,6 +219,21 @@ private static boolean isStringMirror(TypeMirror typeMirror) {
         return ClassName.get(String.class).equals(TypeName.get(typeMirror));
     }
 
+    private static boolean isRuntimeException(TypeMirror typeMirror) {
+        if (typeMirror.getKind() != TypeKind.DECLARED) {
+            return false;
+        }
+
+        if (ClassName.get(RuntimeException.class).equals(TypeName.get(typeMirror))) {
```
carterkozak, line 227: Ideally we'd use the annotation-processing `Types` object for this rather than implementing it ourselves, but if it works I'm not terribly worried :-]

pkoenig10: Ah sorry, didn't know about `Types`. It's a bit tricky to plumb in here since `DefaultDecoderNames` is just a utility class.

https://github.com/palantir/conjure-java/pull/1799

PR #1791, opened 2022-05-11
Support more decoder factory method names

pkoenig10:
## Before this PR
conjure-undertow-annotations supports parameters with static factory methods named `of` or `valueOf`.

## After this PR
conjure-undertow-annotations supports parameters with static factory methods named `fromString` or `create`.

This PR was motivated by my desire to use `URI` as a parameter.

## Possible downsides?
- Is this too much magic built-in behavior? It's not hard to define custom deserializer, but supporting these common factory names out of the box is a ergonomic improvement.

### DIFF
--- a/changelog/@unreleased/pr-1791.v2.yml
+++ b/changelog/@unreleased/pr-1791.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: conjure-undertow-annotations supports parameters with static factory
+    methods named `fromString` or `create`.
+  links:
+  - https://github.com/palantir/conjure-java/pull/1791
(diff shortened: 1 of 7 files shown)


carterkozak:
Slight preference to define custom serializers for factory methods that aren't exceptionally common

pkoenig10:
Ran into another case where this would be very helpful.

We have a number of custom ID types that are UUIDs internally. To avoid overloading and make failable conversions more obvious to callers, the static factory method that takes a `String` is named `fromString` instead of `of`. It would be really nice if these types worked OOTB with the conjure-undertow-annotations.
```java
@Value.Immutable(builder = false)
@ImmutablesStyle
public abstract class BrowserId {

    @Value.Parameter
    public abstract UUID get();

    @Override
    public final String toString() {
        return get().toString();
    }

    public static BrowserId of(UUID browserId) {
        return ImmutableBrowserId.of(browserId);
    }

    public static BrowserId fromString(String browserId) {
        return of(UUID.fromString(browserId));
    }

    public static BrowserId create() {
        return of(UUID.randomUUID());
    }
}
```

`fromString` and `create` are pretty common names for static factory methods and it doesn't seem unreasonable to support them here.

In particular, built-in support makes it trivial to use these types in wrappers like `Optional`, `Set`, etc. Otherwise I have to define a parameter decoder for both `BrowserId` and `Optional<BrowserId>`.

carterkozak:
Fair enough -- we should take the additional step to match behavior in the dialogue annotation processor which iirc is already a bit stricter than the conjure-undertow processor.

carterkozak (approved):

PR #1799, opened 2022-05-13
Use Types to check throws declarations

pkoenig10:
See https://github.com/palantir/conjure-java/pull/1790#discussion_r871793546

### DIFF
--- a/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
+++ b/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
@@ -59,9 +59,11 @@ public final class ParamTypesResolver {
             .collect(ImmutableSet.toImmutableSet());
 
     private final ResolverContext context;
+    private final DefaultDecoderNames defaultDecoderNames;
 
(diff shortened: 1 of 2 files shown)


conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
```diff
@@ -59,9 +59,11 @@ public final class ParamTypesResolver {
             .collect(ImmutableSet.toImmutableSet());
 
     private final ResolverContext context;
+    private final DefaultDecoderNames defaultDecoderNames;
 
     public ParamTypesResolver(ResolverContext context) {
         this.context = context;
+        this.defaultDecoderNames = new DefaultDecoderNames(context);
```
pkoenig10, line 66: Should we just move `DefaultDecoderNames` into `ParamTypesResolver`? We don't instantiate `DefaultDecoderNames` anywhere else so it seems unnecessary for it to be a separate class.

carterkozak: Entirely up to you

carterkozak (approved):
Looks great, thank you!

PR #1805, opened 2022-05-17
Sanitize generated getClass method

raiju:
## After this PR

==COMMIT_MSG==
Sanitize generated getClass method to support `class` field names.
==COMMIT_MSG==


carterkozak (approved):
Great stuff, thanks!

PR #1811, opened 2022-05-25
Add more examples for PathMultiParam

pkoenig10:
When adding an endpoint using `@PathMultiParam`, I was initially surprised that a request without any path segments resulted in a list with a single empty string element rather than an empty list.

This behavior is sensible (since we need to differentiate between `/foo` and `/foo/`), but it can be unexpected. I would have have found it helpful if this was documented.

### DIFF
--- a/conjure-undertow-annotations/readme.md
+++ b/conjure-undertow-annotations/readme.md
@@ -125,9 +125,15 @@ public interface MyService {
 }
 ```
 
-For the above endpoint, calling `GET /path/foo/bar/baz` will decode the path parameter to `["foo", "bar", "baz"]`. Note
-that unless configured differently, path segments will still be url decoded. E.g. calling `GET /path/foo%2fbar` will be
-decoded to `["foo/bar"]`.
+For the above endpoint, the following table shows how various requests are deserilized into wildcard path parameters.
+
+| Request | Params |
+| - | - |
+| `GET /path/` | `[""]` |
+| `GET /path/foo` | `["foo"]` |
+| `GET /path/foo/` | `["foo", ""]` |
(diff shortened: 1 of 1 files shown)


conjure-undertow-annotations/readme.md
```diff
@@ -125,9 +125,15 @@ public interface MyService {
...
 ```
 
-For the above endpoint, calling `GET /path/foo/bar/baz` will decode the path parameter to `["foo", "bar", "baz"]`. Note
-that unless configured differently, path segments will still be url decoded. E.g. calling `GET /path/foo%2fbar` will be
-decoded to `["foo/bar"]`.
+For the above endpoint, the following table shows how various requests are deserilized into wildcard path parameters.
+
+| Request | Params |
+| - | - |
+| `GET /path/` | `[""]` |
```
carterkozak, line 132: Probably best to add the second case mentioned in the pr description
```suggestion
| `GET /path` | `[]` |
| `GET /path/` | `[""]` |
```

carterkozak (approved):

conjure-undertow-annotations/readme.md
pkoenig10: `GET /path` doesn't match this endpoint

pkoenig10: Implementations will never receive an empty list for a `PathMultiParam` parameter. While this may be surprising to devs, it makes sense and is consistent with the path matching rules.

carterkozak: :+1: that makes sense given the pattern requires a slash prior to the wildcard

PR #1831, opened 2022-06-20
Add docs for async handlers

fawind:
## After this PR

==COMMIT_MSG==
Add docs for async handlers
==COMMIT_MSG==


carterkozak (approved):
Thanks, Fabian!

PR #1835, opened 2022-06-30
The conjure-undertow annotation processor supports endpoint tags

carterkozak:
## Before this PR
No support for tags in the annotation processor as we hadn't captured a use-case where they were necessary. As we add support for opt-in metrics using endpoint tags, support is helpful.

## After this PR
==COMMIT_MSG==
The conjure-undertow annotation processor supports endpoint tags
==COMMIT_MSG==


pkoenig10 (approved):
Thanks!

PR #1838, opened 2022-07-12
[draft] Generate 'sealed interfaces' for Conjure unions (feature flagged)

iamdanfox:
## Before this PR

For hackweek, I'm hoping to enable folks to use Java 17's *pattern-matching-switch-expressions* (https://openjdk.org/jeps/406) to visit the different possible variants of a Conjure union. My primary goal is to make code more readable, but I think it should come with a performance boost too.

## After this PR
==COMMIT_MSG==
A new feature flag (`--sealedUnions`) allows users to opt-in to generating Java 17 sealed interfaces for their conjure unions.
==COMMIT_MSG==

Other related work is tracked on https://quip/3YEnAt26mmtE.  

Outstanding TODOs:
- [ ] don't require conjure-java to build and publish using a java 17 JVM.
- [x] Generate a `throwOnKnown` method
  - [ ] do we want to change from a `SafeIllegalArgumentException` now?
- [x] Generate a way to get the `Known` interface (maybe just optional initially?)
- [x] Name escaping of the 'Known' interface name
- [ ] Testing
  - [ ] get `--enable-preview` working in Intellij (https://github.com/palantir/gradle-baseline/pull/2319) so I can write some tests
  - [ ] support switch expressions in palantir-java-format (maybe goethe 0.8.0 is all I need... pls [mavencentral](https://search.maven.org/artifact/com.palantir.goethe/goethe) hurry up)
  - [ ] exhaustive serialization tests, plus tostring tests, plus equality tests


conjure-java-core/src/test/resources/sealedunions/sealedunions/com/palantir/product/SingleUnion.java
```diff
@@ -0,0 +1,163 @@
...
+
+@JsonTypeInfo(
+        use = JsonTypeInfo.Id.NAME,
+        include = JsonTypeInfo.As.EXISTING_PROPERTY,
+        property = "type",
+        visible = true,
+        defaultImpl = UnknownWrapper.class)
+@JsonSubTypes(@JsonSubTypes.Type(FooWrapper.class))
+@JsonIgnoreProperties(ignoreUnknown = true)
+public sealed interface SingleUnion {
```
carterkozak: There are a few things I might suggest changing:
1. We could use this opportunity to become more opinionated about the visitor. It's a lot of code that becomes unnecessary with sealed type pattern matching. What if we drop the visitor entirely?
2. We're exposing the `*Wrapper` types as public api, we can rename them dropping the `Wrapper` suffix to make them more concise. `SingleUnion.Foo` is pretty clear imo.
3. Possibly a controversial take: What if we drop the static factories, and use standard records to implement the union type? We don't use constructors in most places, however these are wrappers around a single value (or a single key-value-pair in the case of unknown).

iamdanfox: I've got some open threads about some of these questions on https://quip/3YEnAt26mmtE - to summarise:

1. `what if we drop the visitor entirely?` - as a long-term state, I definitely agree we should ditch the visitor. My thinking is that it'll probably be most efficient to keep generating it for now, to make the adoption of this flag as easy as possible. 
  - I want to write the error-prone rule to rewrite visitors -> switch expressions. Then maybe when it's fully switched you over, it would flip another feature flag to stop generating these?
  - I had the tantalising thought that maybe if you have sourceCompat=17, we could actually just start generating these new sealed interfaces for _everyone_ (because sealed interfaces are GA in 17, it's only the pattern-matching-switch-expressions that require the `--enable-preview` thing)... sadly I think revapi/ABI changes might derail this plan though.
2. `*Wrapper` types - 100% agree - this is just a commit I haven't made yet.
3. `Use standard records to implement the union type` - yep I'm planning on doing this too, especially because I think destructuring is coming which should make the switch statements _even nicer_ https://openjdk.org/jeps/405.
  - `drop the static factories` I think I actually wanna keep them, because I find the IntelliJ autocomplete very natural when you type `MyUnion.` and then you get a nice dropdown of all the variants.

iamdanfox: I also intend to add that `.Known` sub interface because i think the `.throwOnUnknown` ergonomics will be quite neat

carterkozak: Re dropping the visitor: I suppose it becomes a sequencing problem -- my preference is generally to reach for the aspirational goal, which we can set in new template projects, and follow-up with the sanest possible migration path.

Re jep405: I thought this was why you were opting into experimental features on jdk17, I don't think the sealed types and such need the experimental pattern matching support otherwise, no?

> I find the IntelliJ autocomplete very natural when you type `MyUnion.`

Wouldn't that be equivalent to `new MyUnion.`? If we're using records, I'd prefer not to provide two public constructors for the same type, as that forces developers to decide between the two.

iamdanfox: The main reason I have been opting-in to `--enable-preview` is strictly to get https://openjdk.org/jeps/406 (pattern matching for switch).  I only just discovered https://openjdk.org/jeps/405 (Record Patterns), but I think it'll be another usability benefit!

conjure-java-core/src/test/java/com/palantir/conjure/java/types/Union2.java
```diff
@@ -0,0 +1,215 @@
...
+        use = JsonTypeInfo.Id.NAME,
+        include = JsonTypeInfo.As.EXISTING_PROPERTY,
+        property = "type",
+        visible = true,
+        defaultImpl = UnknownVariant.class)
+@JsonSubTypes({@JsonSubTypes.Type(Foo.class), @JsonSubTypes.Type(Bar.class), @JsonSubTypes.Type(Baz.class)})
+@JsonIgnoreProperties(ignoreUnknown = true)
+public sealed interface Union2 {
+
+    sealed interface Known permits Foo, Bar, Baz {}
```
carterkozak, line 54:
```suggestion
    sealed interface Known extends Union2 permits Foo, Bar, Baz {}
```

Then below we don't need to implement `Union2, Known`, only `Known`

conjure-java-core/src/test/java/com/palantir/conjure/java/types/Union2.java
```diff
@@ -0,0 +1,215 @@
...
+        @SuppressWarnings("UnusedMethod")
+        private String getType() {
+            return "baz";
+        }
+
+        @JsonProperty("baz")
+        public long getValue() {
+            return value;
+        }
+    }
```
carterkozak, line 157: Entirely untested, but I wonder if can do something like this:
```suggestion
    @JsonTypeName("baz")
    @JsonAppend(attrs = @JsonAppend.Attr(propName = "type", value = "baz"), prepend = true)
    record Baz(long value) implements Union2, Known {
        @JsonCreator(mode = JsonCreator.Mode.PROPERTIES)
        public Baz(@JsonSetter("baz") @Nonnull long value) {
            Preconditions.checkNotNull(value, "baz cannot be null");
            this.value = value;
        }

        @JsonProperty("baz")
        public long getValue() {
            return value;
        }
    }
```

carterkozak: Diff didn't render how I expected... The above suggests adding an annotation to the record: `@JsonAppend(attrs = @JsonAppend.Attr(propName = "type", value = "baz"), prepend = true)` and removing the `getType()` method.

conjure-java-core/src/test/java/com/palantir/conjure/java/types/Union2.java
carterkozak: Idea here is that we can use `MyUnion.Known` internally for business logic, ensuring that nothing has to validate against unknown variants, but no casting/validation is required to return or pass `MyUnion.Known` as a `MyUnion`.

We may be able to do something clever with safety annotations on `MyUnion.Known` that can't be guaranteed on the base type with an unknown variant (some issues today since we don't allow safety to be redefined as safer than a supertype, but I think it's reasonable in this instance, trick is ensuring we know when is safe, and when isn't)

PR #1840, opened 2022-07-12
Use consistent name for auth parameters

pkoenig10:
## Before this PR
- Different implementations used different names for auth parameters (Undertow uses `cookieToken` while Jersey uses `token`)
- We duplicate the parameter names in a couple places

## After this PR
- All implementations use `token` as the name for cookie auth parameters
- We use the `Auth` constants everywhere

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
@@ -19,6 +19,7 @@ package com.palantir.conjure.java.util;
 import com.google.common.collect.ImmutableSet;
 import com.palantir.conjure.CaseConverter;
 import com.palantir.conjure.java.ConjureTags;
+import com.palantir.conjure.java.services.Auth;
 import com.palantir.conjure.spec.AuthType;
 import com.palantir.conjure.spec.CookieAuthType;
 import com.palantir.conjure.spec.EndpointDefinition;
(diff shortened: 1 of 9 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
```diff
@@ -64,12 +65,12 @@ public static String sanitizeParameterName(String input, EndpointDefinition endp
                 .map(authType -> authType.accept(new AuthType.Visitor<String>() {
                     @Override
                     public String visitHeader(HeaderAuthType _header) {
-                        return "authHeader";
+                        return Auth.AUTH_HEADER_PARAM_NAME;
                     }
 
                     @Override
                     public String visitCookie(CookieAuthType _cookie) {
-                        return "cookieToken";
```
pkoenig10, line 72: Currently the non-Undertow service generators can produce code that does not compile because the name used in the sanitizer does not match the name used in the generator.

If I have a service definition like this:
```yaml
services:
  CookieService:
    name: Cookie Service
    package: test.api
    default-auth: cookie:PALANTIR_TOKEN
    base-path: /cookies
    endpoints:
      eatCookies:
        http: GET /{token}
        args:
          token: string
```

This will generate the following Jersey service interface (notice that there are two parameters named `token`):
```java
@Consumes(MediaType.APPLICATION_JSON)
@Produces(MediaType.APPLICATION_JSON)
@Path("/")
@Generated("com.palantir.conjure.java.services.JerseyServiceGenerator")
public interface CookieService {

    @GET
    @Path("cookies/{token}")
    @ClientEndpoint(method = "GET", path = "/cookies/{token}")
    void eatCookies(@CookieParam("PALANTIR_TOKEN") @NotNull BearerToken token, @PathParam("token") String token);
}
```

carterkozak (approved):

PR #1842, opened 2022-07-14
Generated equals uses memoizedHashCode if available

schlosna:
## Before this PR
Conjure generated objects with `memoizedHashCode` field did not utilize this in `equals(Object)` comparison for short circuiting mismatches.

## After this PR
==COMMIT_MSG==
Generated equals uses memoizedHashCode if available

When we generate a memoizedHashCode field for a conjure object, the
equals & equalTo implementation now compare memoizedHashCode for objects
being compared, and will short circuit and return false if both objects
have already computed their hashCode and they do not match.

This is beneficial for complex conjure types that may be used as keys in
hash based collections, including graphs that rely on both hashCode and
equals.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/AliasAsMapKeyExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/AliasAsMapKeyExample.java
@@ -95,6 +95,11 @@ public final class AliasAsMapKeyExample {
     }
 
     private boolean equalTo(AliasAsMapKeyExample other) {
+        if (this.memoizedHashCode != 0
+                && other.memoizedHashCode != 0
+                && this.memoizedHashCode != other.memoizedHashCode) {
+            return false;
+        }
         return this.strings.equals(other.strings)
                 && this.rids.equals(other.rids)
                 && this.bearertokens.equals(other.bearertokens)
--- a/changelog/@unreleased/pr-1842.v2.yml
+++ b/changelog/@unreleased/pr-1842.v2.yml
@@ -1,0 +1,15 @@
+type: improvement
+improvement:
+  description: |-
+    Generated equals uses memoizedHashCode if available
+
+    When we generate a memoizedHashCode field for a conjure object, the
+    equals & equalTo implementation now compare memoizedHashCode for objects
+    being compared, and will short circuit and return false if both objects
+    have already computed their hashCode and they do not match.
+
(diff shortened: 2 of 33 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/AliasAsMapKeyExample.java
```diff
@@ -95,6 +95,11 @@ public boolean equals(Object other) {
     }
 
     private boolean equalTo(AliasAsMapKeyExample other) {
+        if (this.memoizedHashCode != 0
+                && other.memoizedHashCode != 0
+                && this.memoizedHashCode != other.memoizedHashCode) {
+            return false;
+        }
```
schlosna, line 102: this is a decent representative example of the value of avoiding recomputing `equals` for each of the `Map` fields

iamdanfox (approved):
Sounds great to me!

I was trying to figure out if there are any possible downsides to this, and the only scenarios that came to mind were:
- This introduces a few more statements and a little more branching on the equals methods. Perhaps this reduces inlining somewhere? Although I imaging branch prediction should be very consistent here.
- Any conjure objects which literally only store a one or two integers (or longs). Then memoizing the hashcode might actually increase the memory consumption (depending on alignment I guess), and now we're adding branching necessary to do an `.equals` call, without being more efficient than just comparing the integers? (I'm thinking perhaps some service has an Alias of a number, or two or three dimensional coordinates defined in conjure?).

Overall though, we've already made the memoizedHashCode change, so I think this will be strictly net positive.

carterkozak (approved):
Thanks!

schlosna:
Thanks for the reviews. 

@iamdanfox the only time we currently generate the `memoizedHashCode` field is for conjure objects that have container fields -- if they have one or more list, set, or map field. I haven't created any specific microbenchmarks for this specific case to see how this impacts inlining (this adds 27 bytes to `equalTo` method bytecode so might push some over the 35 byte inline threshold). We could extract the logic out to a separate `hashCodeMismatch` method if we wanted, but my sense is that it is not worth it for now.

https://github.com/palantir/conjure-java/blob/9e114bed57860ed1cbd13143a4cd545e6aa89101/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java#L645-L668

PR #1847, opened 2022-07-20
Equals compares collections last

schlosna:
## Before this PR
When evaluating `equals(Object)`, collection types are compared before other relative cheaper fields.

## After this PR

==COMMIT_MSG==
Equals compares collections last
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
@@ -19,6 +19,7 @@ package com.palantir.conjure.java.types;
 import static com.palantir.logsafe.Preconditions.checkState;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.palantir.conjure.java.util.JavaNameSanitizer;
 import com.palantir.conjure.spec.FieldName;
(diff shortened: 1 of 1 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
```diff
@@ -38,6 +40,8 @@
 
 public final class MethodSpecs {
 
+    private static final ImmutableSet<String> COLLECTION_TYPES =
+            ImmutableSet.of("java.util.List", "java.util.Set", "java.util.Map");
```
carterkozak, line 44: I wonder if we could pass through the original conjure type information so that we can avoid regressions if we eventually specialize the field types?

schlosna: We could potentially pipe the `List<EnrichedField>` through and use a `DefaultTypeVisitor` to determine if field should be treated as a collection. It might be worth piping something like `List<Pair<EnrichedField, FieldSpec>>` through to ensure proper association between the two.

schlosna:
Right now this is still just a raw WIP I wanted to brain dump as I was digging through a few JFR and saw some conjure types `equals` methods evaluating `java.util.Collections$UnmodifiableSet.equals(Object)`. 

There are some places where we might want to just manually reorder fields in yaml, but was thinking we could defer collection comparison to short-circuit in case of mismatches.

image

PR #1848, opened 2022-07-21
Allow customization of Jackson Undertow Encodings

westinrm:
## Before this PR
Some old endpoints can't easily be migrated to using Undertow because they feature response/request structures that require additional ObjectMapper modules to be registered.

## After this PR

==COMMIT_MSG==
Allow customization of the ObjectMapper used by Jackson based Undertow encodings
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1848.v2.yml
+++ b/changelog/@unreleased/pr-1848.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Allow customization of the ObjectMapper used by Jackson based Undertow
+    encodings
+  links:
+  - https://github.com/palantir/conjure-java/pull/1848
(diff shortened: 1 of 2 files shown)


carterkozak (changes requested):
The conjure-undertow annotation processor allows per-endpoint customization. For pure conjure endpoints, I’d prefer not to open the door for this sort of thing as there’s a great deal of risk that customizations break the conjure spec for unintended endpoints

westinrm:
The problem is it's a pain to do those per endpoint customizations:
- There isn't any shared state available to the DeserializerFactory/SerializerFactory which means you have to create a new ObjectMapper each time.
- Most of the exception handling in AbstractJacksonEncoding uses package private types that aren't even visible, so you can't even get consistent exception behavior even if you copy the code.

If we don't want to expose it like this, can we at least add something to conjure-undertow-annotations so that you just have to provide an ObjectMapperCustomizer on the endpoint if all you need to do is tweak an object mapper flag?
Failing that could we at least expose some utility methods `static <T> Deserializer<T> deserializer(TypeMaker<T>, ObjectMapper)` and `static <T> Serializer<T> serializer(TypeMarker<T>, ObjectMapper)`?

PR #1852, opened 2022-07-22
Collection setters use constructor when clearing

schlosna:
## Before this PR
Passing a `Collection` to the generated builder method for `set` fields did not benefit from the presizing estimate logic in [`LinkedHashSet(Collection)`](https://github.com/openjdk/jdk/blob/c79baaa811971c43fbdbc251482d0e40903588cc/src/java.base/share/classes/java/util/LinkedHashSet.java#L177) 

## After this PR

==COMMIT_MSG==
Collection setters use constructor when clearing
==COMMIT_MSG==

## Possible downsides?
This adds a couple methods to `ConjureCollections` that return concrete `ArrayList` and `LinkedHashSet` implementations that ideally we wouldn't expose, but we need these to avoid Guava dependency.

We might also want to (separately?) consider generating builder methods that take a `Stream` and/or `Iterator` of elements to be added to Conjure objects as we have many places using Guava `Collections2`/`Lists`/`Maps`.`transform` or `Stream`s unnecessarily collecting to intermediate collection (or `Stream.forEach(builder::addFoo)`) which makes concise builder transformations slightly more difficult.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/CollectionsTestObject.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/CollectionsTestObject.java
@@ -11,6 +11,7 @@ import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
@@ -216,8 +217,12 @@ public final class CollectionsTestObject {
         @JsonSetter(value = "items", nulls = Nulls.SKIP)
         public Builder items(@Nonnull Iterable<String> items) {
             checkNotBuilt();
-            this.items.clear();
-            ConjureCollections.addAll(this.items, Preconditions.checkNotNull(items, "items cannot be null"));
+            if (items instanceof Collection) {
+                this.items = new ArrayList<>((Collection) Preconditions.checkNotNull(items, "items cannot be null"));
+            } else {
+                this.items.clear();
+                ConjureCollections.addAll(this.items, Preconditions.checkNotNull(items, "items cannot be null"));
+            }
             return this;
         }
 
@@ -236,8 +241,7 @@ public final class CollectionsTestObject {
         @JsonSetter(value = "itemsMap", nulls = Nulls.SKIP)
         public Builder itemsMap(@Nonnull Map<String, Integer> itemsMap) {
             checkNotBuilt();
-            this.itemsMap.clear();
-            this.itemsMap.putAll(Preconditions.checkNotNull(itemsMap, "itemsMap cannot be null"));
+            this.itemsMap = new LinkedHashMap<>(Preconditions.checkNotNull(itemsMap, "itemsMap cannot be null"));
             return this;
         }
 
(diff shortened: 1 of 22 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -349,17 +367,36 @@ private CodeBlock typeAwareAssignment(EnrichedField enriched, Type type, boolean
                     ConjureCollections.class,
```
schlosna, line 367: I think we’re trying to keep the dependency footprint small so don’t want to necessarily pull in Guava, but we could simplify this with something like `Lists.newArrayList(Iterable)` / `Sets.newHashSet(Iterable)`

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -349,17 +367,36 @@ private CodeBlock typeAwareAssignment(EnrichedField enriched, Type type, boolean
...
                     spec.name,
                     Expressions.requireNonNull(spec.name, enriched.fieldName().get() + " cannot be null"));
-            return shouldClearFirst
-                    ? CodeBlocks.of(CodeBlocks.statement("this.$1N.clear()", spec.name), addStatement)
-                    : addStatement;
+            if (shouldClearFirst) {
+                return CodeBlock.builder()
+                        .beginControlFlow("if ($1N instanceof $2T)", spec.name, Collection.class)
```
schlosna: I think we’re trying to keep the dependency footprint small so don’t want to necessarily pull in Guava, but we could simplify this with something like `Lists.newArrayList(Iterable)` / `Sets.newHashSet(Iterable)`

conjure-java-core/src/integrationInput/java/com/palantir/product/CollectionsTestObject.java
```diff
@@ -216,8 +217,12 @@ public Builder from(CollectionsTestObject other) {
...
         public Builder items(@Nonnull Iterable<String> items) {
             checkNotBuilt();
-            this.items.clear();
-            ConjureCollections.addAll(this.items, Preconditions.checkNotNull(items, "items cannot be null"));
+            if (items instanceof Collection) {
+                this.items = new ArrayList<>((Collection) Preconditions.checkNotNull(items, "items cannot be null"));
+            } else {
+                this.items.clear();
+                ConjureCollections.addAll(this.items, Preconditions.checkNotNull(items, "items cannot be null"));
+            }
```
schlosna: this is an example of what could be much more concise with Guava's `Lists` `static <T> ArrayList<T> newArrayList(Iterable<T>)` or similar if we wanted to add to `ConjureCollections` 

```suggestion
            this.items = Lists.newArrayList(Preconditions.checkNotNull(items, "items cannot be null"));
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
schlosna: Added 2 methods to `ConjureCollections` to handle these and simplify generated code

conjure-java-core/src/integrationInput/java/com/palantir/product/CollectionsTestObject.java
schlosna: moved to `ConjureCollections`

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -40,4 +44,30 @@ public static <T> void addAll(Collection<T> addTo, Iterable<? extends T> element
...
+        Preconditions.checkNotNull(iterable, "iterable cannot be null");
+        if (iterable instanceof Collection) {
+            return new LinkedHashSet<>((Collection<T>) iterable);
+        }
+        LinkedHashSet<T> set = new LinkedHashSet<>();
+        for (T item : iterable) {
+            set.add(item);
+        }
+        return set;
+    }
```
carterkozak, line 72: These look reasonable to me

carterkozak (approved):

conjure-java-core/src/integrationInput/java/com/palantir/product/CollectionsTestObject.java
carterkozak: :+1: this allows hotspot to do what it's good at

PR #1854, opened 2022-07-25
Compare content-type case insensitive

fawind:
## Before this PR
Small drivy-by fix, but I think we should match content types case insensitive. We do the same when determining a decoding: [example](https://github.com/palantir/conjure-java/blob/8afc1c6c92a33c03eff4d308ee31f855fe017e8b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/Encodings.java#L61) added in https://github.com/palantir/conjure-java/pull/686.

## After this PR

==COMMIT_MSG==
Compare content-type case insensitive
==COMMIT_MSG==

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
@@ -100,7 +100,7 @@ final class ConjureBodySerDe implements BodySerDe {
     @Override
     public InputStream deserializeInputStream(HttpServerExchange exchange) {
         String contentType = getContentType(exchange);
-        if (!contentType.startsWith(BINARY_CONTENT_TYPE)) {
+        if (!contentType.toLowerCase().startsWith(BINARY_CONTENT_TYPE)) {
             throw FrameworkException.unsupportedMediaType(
                     "Unsupported Content-Type", SafeArg.of("Content-Type", contentType));
         }
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
@@ -40,6 +40,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PushbackInputStream;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import org.xnio.IoUtils;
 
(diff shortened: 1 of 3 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
```diff
@@ -100,7 +100,7 @@ public void serialize(BinaryResponseBody value, HttpServerExchange exchange) thr
     @Override
     public InputStream deserializeInputStream(HttpServerExchange exchange) {
         String contentType = getContentType(exchange);
-        if (!contentType.startsWith(BINARY_CONTENT_TYPE)) {
+        if (!contentType.toLowerCase().startsWith(BINARY_CONTENT_TYPE)) {
```
carterkozak: Let's avoid locale-dependent behavior:
```suggestion
        if (!contentType.toLowerCase(Locale.ROOT).startsWith(BINARY_CONTENT_TYPE)) {
```

fawind: Good flag, updated!

carterkozak (approved):
Great catch, thanks!

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureBodySerDe.java
```diff
@@ -100,7 +101,7 @@ public void serialize(BinaryResponseBody value, HttpServerExchange exchange) thr
     @Override
     public InputStream deserializeInputStream(HttpServerExchange exchange) {
         String contentType = getContentType(exchange);
-        if (!contentType.startsWith(BINARY_CONTENT_TYPE)) {
+        if (!contentType.toLowerCase(Locale.ROOT).startsWith(BINARY_CONTENT_TYPE)) {
```
schlosna: should we push the toLowerCase into `getContentType`?

carterkozak: Sounds reasonable to me

schlosna: Actually we could do this without allocating

```suggestion
        if (contentType.length() < BINARY_CONTENT_TYPE.length()
                || !contentType.regionMatches(/* ignoreCase= */ true, 0, BINARY_CONTENT_TYPE, 0, BINARY_CONTENT_TYPE.length())) {
```

fawind: Thanks, both good suggestions! Went with the `contentType#regionMatches` approach for now given that we only have a single place where we do this comparison (all other paths are handled in the respective encodings). Can pull this into a helper method once we have more usage.

PR #1866, opened 2022-08-09
Change Generated to processing.Generated

mglazer:
## Before this PR

This change is mostly needed for the Jakarta migration, this will enable us to not use
the legacy javax annotation import types.

## After this PR

==COMMIT_MSG==

Modern JDKs (anything >= JDK9) use the newer version of this annotation
to represent Generated types. This change does force a minimum JRE
dependency of Conjure to JRE9 or later at compile time (this annotation
only has source retention).

Annotation processors are able to get away with detecting the JDK
and generating code specfic to that JDK, I don't believe we can
do that with Java Poet, so this is potentially a more invasive change
than I'd traditionally like to make.

One option is to add an option to the Options type to opt out of this
behavior, or we can just force Java 9+.
==COMMIT_MSG==


carterkozak (approved):

carterkozak:
You’ll need to regen generated sources with

./gradlew test -Drecreate=true

mglazer:
thanks! I just wanted provisional approval first before anything else.

PR #1869, opened 2022-08-09
Add support for jakarta package prefixes in code generation

mglazer:
## Before this PR

Jersey code generation didn't support generating code with the newer jakarta package prefixes.

## After this PR

==COMMIT_MSG==
Add support for jakarta package prefixes in code generation
==COMMIT_MSG==


carterkozak (approved):
Thanks! We likely shouldn't use this flag until CJR supports client interfaces using the jakarta annotations.

mglazer:
Yup! Unfortunately I can't actually complete CJR until I get this version :)

PR #1872, opened 2022-08-19
Add FormParam annotation

pkoenig10:
Adds a `@FormParam` annotation for easily declaring endpoints that use form data.

The implementation in this PR does not support file form parameters. Supporting files doesn't fit as neatly into the existing structure and is a more niche feature. We can consider adding support in the future, but it might be better for consumers who need file form parameters to just implement it themselves.

### DIFF
--- a/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/FormParamDeserializer.java
+++ b/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/FormParamDeserializer.java
@@ -62,7 +62,7 @@ public final class FormParamDeserializer<T> implements Deserializer<T> {
             return Collections.emptyList();
         }
 
-        Collection<String> values = new ArrayList<>();
+        Collection<String> values = new ArrayList<>(maybeValues.size());
 
         for (FormValue value : maybeValues) {
             Preconditions.checkArgument(!value.isFileItem(), "Form value was a file item");
--- a/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/FormParamDeserializer.java
+++ b/conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/FormParamDeserializer.java
@@ -1,0 +1,76 @@
+/*
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.undertow.annotations;
+
+import com.google.common.collect.Collections2;
+import com.palantir.conjure.java.undertow.lib.Deserializer;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.server.handlers.form.FormData.FormValue;
+import io.undertow.server.handlers.form.FormDataParser;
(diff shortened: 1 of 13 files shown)


conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/FormParamDeserializer.java
```diff
@@ -0,0 +1,76 @@
...
+    private Collection<String> getValues(Deque<FormValue> maybeValues) {
+        if (maybeValues == null) {
+            return Collections.emptyList();
+        }
+
+        for (FormValue value : maybeValues) {
+            Preconditions.checkArgument(!value.isFileItem(), "Form value was a file item");
+        }
+
+        return Collections.unmodifiableCollection(Collections2.transform(maybeValues, FormValue::getValue));
```
pkoenig10: Alternatively we could just copy into a new list. Happy to do whichever you'd prefer.

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/FormParamDeserializer.java
```diff
@@ -0,0 +1,76 @@
...
+        return Preconditions.checkNotNull(decoder.decode(getValues(maybeValues)), "Decoder produced a null value");
+    }
+
+    private Collection<String> getValues(Deque<FormValue> maybeValues) {
+        if (maybeValues == null) {
+            return Collections.emptyList();
+        }
+
+        for (FormValue value : maybeValues) {
+            Preconditions.checkArgument(!value.isFileItem(), "Form value was a file item");
```
carterkozak, line 68: I suppose we could add support for file items later on with a new attribute on `@FormParam`. Just worth thinking about whether there's something we can do right now to ensure that API is as clean as possible when it's added.

pkoenig10: Agree that it's worth considering this now.

Couple questions:
- What does this API look like for consumers? I don't think a "decoder" makes sense here. We'd probably want to require that the parameter type is always `Path`.
- Given this, does it make sense to reuse the same annotation?

carterkozak: I'm not sure we'd always want the parameter to be a path, rather some collection of byte-source and metadata describing the file, but I think you're spot on about using a different annotation for that sort of thing 👍

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/FormParamDeserializer.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
```
carterkozak:
```suggestion
 * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
```

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/Handle.java
```diff
@@ -122,6 +122,18 @@
         Class<? extends CollectionParamDecoder<?>> decoder() default DefaultParamDecoder.class;
     }
 
+    @Retention(RetentionPolicy.SOURCE)
```
carterkozak: rebase + class retention (changed recently to support local codegen of published generic interfaces)

conjure-undertow-annotations/src/main/java/com/palantir/conjure/java/undertow/annotations/FormParamDeserializer.java
carterkozak: I'd bias toward copying here, lazy collections can result in unexpected behavior later on. Very slight preference though, either way works.

conjure-undertow-processor-example/src/test/java/com/palantir/conjure/java/undertow/example/ExampleResource.java
```diff
@@ -90,6 +90,11 @@ public String queryParam(String queryParameter) {
         return Preconditions.checkNotNull(queryParameter, "Query parameter is required");
     }
 
+    @Override
+    public String formParam(String formParameter) {
+        return Preconditions.checkNotNull(formParameter, "Query parameter is required");
```
carterkozak:
```suggestion
        return Preconditions.checkNotNull(formParameter, "Form parameter is required");
```

carterkozak (approved):
lgtm, thanks!

PR #1873, opened 2022-08-23
conjure-undertow processor supports specializing generic interfaces

carterkozak:
## Before this PR
No way to specialize generic interfaces.

## After this PR

==COMMIT_MSG==
conjure-undertow processor supports specializing generic interfaces
==COMMIT_MSG==

This provides an annotation to opt out of codegen for an interface meant to be extended using class-level `@Handle.Generate(false)`. Similarly, an interface which extends such an interface may be annotated `@Handle.Generate` to opt into codegen without any `@Handle(...` methods.

## Possible downsides?
conjure-undertow annotations are retained by classes as opposed to source-only. This may result in compiler warnings when they aren't present.

### DIFF
--- a/changelog/@unreleased/pr-1873.v2.yml
+++ b/changelog/@unreleased/pr-1873.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: conjure-undertow processor supports specializing generic interfaces
+  links:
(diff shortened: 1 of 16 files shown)


ricksurya (approved):
This is awesome thanks!

fawind (approved):
Looks good!

PR #1876, opened 2022-09-02
Validate path params

pkoenig10:
## Before this PR
`conjure-undertow-processor` does not validate that the path template parameters match the `@PathParam` method parameters.

## After this PR
`conjure-undertow-processor` validates that the path template parameters match the `@PathParam` method parameters.

Caught a couple violations in the existing test cases.


carterkozak (approved):
Thanks!

PR #1883, opened 2022-09-21
Make .equals accept null values

JacekLach:
This solves a bunch of nullaway warnings

## Before this PR

passing a nullable value to .equals causes nullaway warnings
## After this PR

==COMMIT_MSG==
Mark the argument to .equals on a conjure bean as `@Nullable`
==COMMIT_MSG==


carterkozak (approved):
thanks!

PR #1884, opened 2022-09-25
Conjure errors may provide Jetbrains annotations

carterkozak:
This adds a new `--jetbrainsContractAnnotations` parameter which opts into the new feature. The gradle plugin will be updated to both add the jetbrains annotations dependency and add this parameter.

## Before this PR
No way to let static analysis tools understand that these methods are preconditions checks.

## After this PR
==COMMIT_MSG==
Conjure errors may provide Jetbrains annotations when the `--jetbrainsContractAnnotations` flag is provided
==COMMIT_MSG==

## Possible downsides?
Breadth of feature-set.


carterkozak:
Haven't been able to run tests on in-flight wifi, heh

tpetracca (approved):

PR #1897, opened 2022-10-14
Add docs for visitor builders to README

### LINKED ISSUE #1896: Add docs for visitor builders to README
## What happened?

I wanted to point someone to the reference for how to use visitor builders, but they aren't described in the README.  The best example is in the PR description of the feature: https://github.com/palantir/conjure-java/pull/465

We should add an example for these visitor builders to the docs for conjure-java (is that just the README?) so people can reference it and use them more widely.

ash211:
Fixes https://github.com/palantir/conjure-java/issues/1896

## Before this PR
No mention in README for visitor builders

## After this PR
==COMMIT_MSG==
Add docs for visitor builders to README
==COMMIT_MSG==

## Possible downsides?
This README code isn't compiled as part of CI so could have compile errors or other errors that aren't caught in CI, potentially misleading readers.

### DIFF
--- a/changelog/@unreleased/pr-1897.v2.yml
+++ b/changelog/@unreleased/pr-1897.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add docs for visitor builders to README
+  links:
+  - https://github.com/palantir/conjure-java/pull/1897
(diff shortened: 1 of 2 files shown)


carterkozak (approved):
thanks!

PR #1916, opened 2022-11-16
fix eclipse warnings

### LINKED ISSUE #1635: suppress or fix warnings in generated code
## What happened?
There are warnings in eclipse on generated code.
I tried to suppress them in https://github.com/palantir/gradle-conjure/pull/951 but if we can suppress them at the source, that would be great.

All methods in a `class DefaultBuilder implements Builder` seem to be missing `@Override` 
These can be unused:
`_plainSerDe` on Async and Blocking classes.
`UndertowRuntime runtime` on an Endpoint.

There may be more, but I am currently working on a very small project with only these warnings.

No warnings in eclipse on generated code.

rzpt:
Only generated _plainSerDe when it is used.

Partial fix for #1635

## Before this PR
Generated code has a lot of warnings in eclipse.

## After this PR
==COMMIT_MSG==
Fix unused warnings in eclipse
==COMMIT_MSG==

## Possible downsides?
Untested in intellij since I don't use it. Only tested on a single internal repo.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
@@ -52,7 +52,6 @@ import com.palantir.dialogue.Serializer;
 import com.palantir.dialogue.TypeMarker;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
-import com.squareup.javapoet.AnnotationSpec;
 import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.FieldSpec;
@@ -75,6 +74,7 @@ public final class DefaultStaticFactoryMethodGenerator implements StaticFactoryM
     private final ParameterTypeMapper parameterTypes;
     private final ReturnTypeMapper returnTypes;
     private final StaticFactoryMethodType methodType;
+    private final ThreadLocal<Boolean> shouldGeneratePlainSerDe = ThreadLocal.withInitial(() -> false);
 
     public DefaultStaticFactoryMethodGenerator(
             Options options,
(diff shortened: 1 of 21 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -75,6 +74,7 @@ public final class DefaultStaticFactoryMethodGenerator implements StaticFactoryM
     private final ParameterTypeMapper parameterTypes;
     private final ReturnTypeMapper returnTypes;
     private final StaticFactoryMethodType methodType;
+    private final ThreadLocal<Boolean> shouldGeneratePlainSerDe = ThreadLocal.withInitial(() -> false);
```
ash211: ThreadLocals have a bunch of gotchas that can be really troublesome to debug.  Is there a reason we can't pass info up and down the stack to set this as needed all within the `generate()` method and its calls?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -910,11 +910,6 @@ private static MethodSpec createWrapperAcceptMethod(
         } else {
             methodBuilder.addStatement("return $N.$N($N)", visitor, visitMethodName, valueName);
         }
-        if (isDeprecated) {
-            methodBuilder.addAnnotation(AnnotationSpec.builder(SuppressWarnings.class)
-                    .addMember("value", "$S", "deprecation")
-                    .build());
-        }
```
ash211: this change to remove the SuppressWarnings annotations seems independent from the plainSerDe change.  Can they be separate PRs, or are they interrelated somehow?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
rzpt: Done.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
rzpt: Moved deprecation stuff to https://github.com/palantir/conjure-java/pull/1917

PR #1917, opened 2022-11-18
fix eclipse deprecation warnings

rzpt:
Remove `@SuppressWarnings("deprecation")` on deprecated methods. The annotation is for suppressing calls of deprecated methods, not implementation of deprecated methods.

partial fix for #1635

Splitting out from https://github.com/palantir/conjure-java/pull/1916

## Before this PR
Generated code produces warnings in eclipse

## After this PR
==COMMIT_MSG==
Generated code does not produce deprecation warnings in eclipse by default
==COMMIT_MSG==

## Possible downsides?
Not tested in intellij. Only tested on one internal repo.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/Union.java
+++ b/conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/Union.java
@@ -305,7 +305,6 @@ public final class Union {
         }
 
         @Override
-        @SuppressWarnings("deprecation")
         public <T> T accept(Visitor<T> visitor) {
             return visitor.visitBar(value);
         }
@@ -351,7 +350,6 @@ public final class Union {
         }
 
         @Override
-        @SuppressWarnings("deprecation")
         public <T> T accept(Visitor<T> visitor) {
             return visitor.visitBaz(value);
         }
--- a/changelog/@unreleased/pr-1917.v2.yml
+++ b/changelog/@unreleased/pr-1917.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Generated code does not produce deprecation warnings in eclipse by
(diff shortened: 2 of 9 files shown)


conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/Union.java
```diff
@@ -305,7 +305,6 @@ private int getValue() {
         }
 
         @Override
-        @SuppressWarnings("deprecation")
```
carterkozak, line 308: Is this correct? I would have though a non-deprecated method (accept) calling a deprecated method (Visitor::visitBar) would be problematic?

rzpt: Hypothesis: if the deprecated method is defined in the same file, then the usage does not produce a warning.

carterkozak: iirc we introduced the suppressions because we use javac `-Xlint:deprecation` in some places where it may apply to generated code. I'd like to ensure we're not regressing that functionality

rzpt: Okay, then would you be willing to approve something similar to https://github.com/palantir/gradle-conjure/pull/951? I'm thinking of suppressing all warnings from the generated project.

carterkozak: I think that’s reasonable given the warnings may not necessarily align with deprecation checks elsewhere.

rzpt:
Doing this instead https://github.com/palantir/gradle-conjure/pull/1136

PR #1919, opened 2022-11-25
Performance: faster enum equals method

j-baker:
Conjure enums are usually not unknown. Here, equals devolves to string equality when the two enum cases are _not_ equal, which is slower than you'd think. This PR short circuits in most cases with an object equality check, at the cost of doing an extra reference equality checks when comparing unknown types (aka extremely unlikely).

In my case this is approximately 3% of all JMH samples. Slightly contrived example but this should be a small win everywhere.


carterkozak (approved):
Looks good. Thanks, James!

PR #1921, opened 2022-11-28
Always log QosException at info

pkoenig10:
Continuation of https://github.com/palantir/conjure-java/pull/1519

## Before this PR
It is common for services to throw `QosException` without a cause, particularly in cases where requests are being rate-limited.

When this happens, it is difficult to determine why requests are failing without a stacktrace, especially when `QosException` is thrown in multiple places.

## After this PR
Requests that fail with `QosException` always log at the `INFO` level so operators can determine why a request failed.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
@@ -94,7 +94,9 @@ public enum ConjureExceptions implements ExceptionHandler {
     private static void qosException(HttpServerExchange exchange, QosException qosException) {
         qosException.accept(QOS_EXCEPTION_HEADERS).accept(exchange);
 
-        if (qosExceptionHasAdditionalMetadata(qosException) || qosLoggingRateLimiter.tryAcquire()) {
+        if (log.isDebugEnabled()) {
+            log.debug("Quality-of-Service error handling request", qosException);
+        } else if (qosExceptionHasAdditionalMetadata(qosException) || qosLoggingRateLimiter.tryAcquire()) {
             log.info("Quality-of-Service error handling request", qosException);
         }
 
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
@@ -89,27 +89,8 @@ public enum ConjureExceptions implements ExceptionHandler {
 
     private static void qosException(HttpServerExchange exchange, QosException qosException) {
         qosException.accept(QOS_EXCEPTION_HEADERS).accept(exchange);
-        if (qosExceptionHasAdditionalMetadata(qosException)) {
-            log.info("quality-of-service intervention", qosException);
-        } else {
-            log.debug("quality-of-service intervention", qosException);
-        }
+        log.info("quality-of-service intervention", qosException);
         writeResponse(exchange, Optional.empty(), qosException.accept(QOS_EXCEPTION_STATUS_CODE));
-    }
-
-    /** Returns  true if {@link QosException} provides additional metadata and should be logged at {@code info}. */
-    private static boolean qosExceptionHasAdditionalMetadata(QosException qosException) {
-        try {
-            if (qosException.getCause() != null) {
-                return true;
-            }
-            Throwable[] suppressed = qosException.getSuppressed();
-            return suppressed != null && suppressed.length > 1;
-        } catch (Throwable t) {
-            // Fallback in case of unexpected bytecode manipulation (getCause or getSuppressed throws due to
-            // poor instrumentation/mocks)
-            return true;
-        }
     }
 
     // RemoteExceptions are thrown by Conjure clients to indicate a remote/service-side problem.
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
@@ -16,6 +16,7 @@
 
 package com.palantir.conjure.java.undertow.runtime;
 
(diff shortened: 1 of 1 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -89,29 +89,10 @@ private static void serviceException(HttpServerExchange exchange, ServiceExcepti
 
     private static void qosException(HttpServerExchange exchange, QosException qosException) {
         qosException.accept(QOS_EXCEPTION_HEADERS).accept(exchange);
-        if (qosExceptionHasAdditionalMetadata(qosException)) {
-            log.info("quality-of-service intervention", qosException);
```
carterkozak, line 93: What if we add a rate limiter here, and log at info ~1 permit per second? That should strike a balance between causing our storage service to spam logs and the current state where these aren’t terribly debuggable.
If that works well enough, I could see removing the additional-metadata check in favor of a pure-ratelimiter approach.

What do you think?

kevinychen: That'll make the log much less useful - I'd expect normally we'd have a specific trace(Id) that throws a 429, and we'd miss this stacktrace unless it happens to be one that is able to acquire the rate limiter.

Why is there a concern that this will spam? This will log at most once per request, so will be no worse than request logs.

carterkozak: Stack traces are much denser than anything in the request log

pkoenig10: I don't expect the volume of `QosException` to be that high through.

I agree with what @kevinychen said above - it's very valuable to take a trace ID for a 4xx or 5xx request and be able to easily determine why the request failed.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -89,27 +93,12 @@ private static void serviceException(HttpServerExchange exchange, ServiceExcepti
...
-                return true;
-            }
-            Throwable[] suppressed = qosException.getSuppressed();
-            return suppressed != null && suppressed.length > 1;
-        } catch (Throwable t) {
-            // Fallback in case of unexpected bytecode manipulation (getCause or getSuppressed throws due to
-            // poor instrumentation/mocks)
-            return true;
+        if (qosLoggingRateLimiter.tryAcquire()) {
+            log.info("Quality-of-Service error handling request", qosException);
```
pkoenig10, line 100: Updated this log message to be a bit more human-readable and more closely match the other log messages used here.

pkoenig10:
@carterkozak Is there a reason the check for suppressed exceptions requires 2 suppressed exceptions rather than just 1?

https://github.com/palantir/conjure-java/blob/develop/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java#L107

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -89,15 +93,15 @@ private static void serviceException(HttpServerExchange exchange, ServiceExcepti
...
     private static void qosException(HttpServerExchange exchange, QosException qosException) {
         qosException.accept(QOS_EXCEPTION_HEADERS).accept(exchange);
-        if (qosExceptionHasAdditionalMetadata(qosException)) {
-            log.info("quality-of-service intervention", qosException);
-        } else {
-            log.debug("quality-of-service intervention", qosException);
+
+        if (qosExceptionHasAdditionalMetadata(qosException) || qosLoggingRateLimiter.tryAcquire()) {
+            log.info("Quality-of-Service error handling request", qosException);
         }
```
carterkozak: Let's retain the `debug` functionality, and avoid burning the rate limiter when debug is enabled:
```suggestion
        if (log.isDebugEnabled()) {
            log.debug("Quality-of-Service error handling request", qosException);
        } else if (qosExceptionHasAdditionalMetadata(qosException) || qosLoggingRateLimiter.tryAcquire()) {
            log.info("Quality-of-Service error handling request", qosException);
        }
```

carterkozak: It might be helpful to mention in the `info` line that these exceptions are logged at info once per second.

carterkozak:
> @carterkozak Is there a reason the check for suppressed exceptions requires 2 suppressed exceptions rather than just 1?

I think that's probably a bug :facepalm: My bias would be to leave it as is at this point, since we don't have a great idea how much noise it would produce if we undid it. Based on feedback from the rate-limiter logic, I'd love to remove `qosExceptionHasAdditionalMetadata` entirely since it's not obvious when we add-suppressed or create with cause that it causes additional qos data to be logged.
I also like the idea we discussed of supporting QosException factory methods taking a `reason` as a longer term solution to the observability issue.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
pkoenig10: I don't think the log line is an appropriate place for this, it just makes it more cumbersome to consume these logs.

pkoenig10:
> I'd love to remove qosExceptionHasAdditionalMetadata entirely

I don't think this is desirable. Often when QoS exceptions are thrown with a cause, that cause is very useful. We rely on this pretty extensively in the products I am responsible for.

pkoenig10:
> My bias would be to leave it as is at this point, since we don't have a great idea how much noise it would produce if we undid it.

Sure but we didn't have a good idea how much noise this would produce when adding this logging in the first place. Feels like we should fix this to match the original intention.

carterkozak (approved):
Thanks!

PR #1932, opened 2023-01-11
Only set the Content-Type response header if is isn't already set

SerialVelocity:
## Before this PR

Servers which use a custom content-type response header had this header overwritten silently.

## After this PR

==COMMIT_MSG==
Only set the Content-Type response header if is isn't already set
==COMMIT_MSG==

## Possible downsides?

It should have very little impact on standard code-paths.


pkoenig10 (changes requested):
`BinaryResponseBody` is explicitly only intended to represent `application/octet-stream` responses.

If you need to create a response with a different `Content-Type` you should define a custom type. It should be fairly straight-forward to create a custom type that is similar to `BinaryResponseBody`.

PR #1936, opened 2023-01-17
Update wildfly to 1.6.0.Final

fawind:
## Before this PR

`org.wildfly.common:wildfly-common ≤ 1.5.4.Final` gets flagged for "an XXE (Xml eXternal Entity) injection flaw in javax.xml.parsers.DocumentBuilderFactory and javax.xml.stream.XMLInputFactory". Can't find a CVE for the finding and saw previous discussions on this possibly being a false positive.

## After this PR

Let's update to avoid the finding. All of our internal servers already get a `1.6.0.Final` constraint from our server framework but I saw two dependency bundles that consume conjure-java and get flagged for this.

==COMMIT_MSG==
Update wildfly to 1.6.0.Final
==COMMIT_MSG==

### DIFF
--- a/versions.props
+++ b/versions.props
@@ -42,6 +42,7 @@ com.palantir.human-readable-types:* = 1.6.0
 org.derive4j:* = 1.1.1
 com.google.auto:auto-common = 1.2.1
 com.google.testing.compile:compile-testing = 0.19
+org.wildfly.common:wildfly-common = 1.6.0.Final
 
 # dependency-upgrader:OFF
 # Avoid forcing consumers to upgrade jackson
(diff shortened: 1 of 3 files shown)


versions.props
```diff
@@ -42,6 +42,7 @@ com.palantir.human-readable-types:* = 1.6.0
 org.derive4j:* = 1.1.1
 com.google.auto:auto-common = 1.2.1
 com.google.testing.compile:compile-testing = 0.19
+org.wildfly.common:wildfly-common = 1.6.0.Final
```
carterkozak, line 45: I think you'll need to add `runtimeOnly 'org.wildfly.common:wildfly-common'` to the module which brings this in (likely conjure-java-undertow-lib) otherwise the version bump won't be applied to dependent projects.

fawind: Yep, you're right that this is needed, added!

In the case I was looking at, wildfly gets introduced through:

```
com.palantir.conjure.java:conjure-java-undertow-runtime
    +--- conjure-undertow-lib
        +--- undertow
```

carterkozak (approved):

PR #1944, opened 2023-02-01
Generator Should Read ExternalImportType Safety

lsingh123:
If an Alias, Field, or Argument is an external import type, use the underlying import's safety instead of the declared safety (which will be empty). Converts all usages of AliasDefinition.getSafety(), FieldDefinition.getSafety(), and ArgumentDefinition.getSafety() to the corresponding wrapper methods in SafetyUtils.java.


lsingh123:
https://github.com/palantir/conjure-java/pull/1945 is a better solution

conjure-java-core/src/integrationInput/java/com/palantir/product/SafeExternalLongExample.java
```diff
@@ -0,0 +1,189 @@
...
+
+    @JsonProperty("safeExternalLongValue")
+    @Safe
+    public long getSafeExternalLongValue() {
+        return this.safeExternalLongValue;
+    }
+
+    @JsonProperty("optionalSafeExternalLong")
+    @JsonInclude(JsonInclude.Include.NON_ABSENT)
+    public Optional<Long> getOptionalSafeExternalLong() {
```
lsingh123, line 59: not sure if this should have a safety annotation on it

conjure-java-core/src/integrationInput/java/com/palantir/product/SafeExternalLongExample.java
```diff
@@ -0,0 +1,189 @@
...
+        @JsonSetter("safeExternalLongValue")
+        public Builder safeExternalLongValue(@Safe long safeExternalLongValue) {
+            checkNotBuilt();
+            this.safeExternalLongValue = safeExternalLongValue;
+            this._safeExternalLongValueInitialized = true;
+            return this;
+        }
+
+        @JsonSetter(value = "optionalSafeExternalLong", nulls = Nulls.SKIP)
+        public Builder optionalSafeExternalLong(@Nonnull Optional<? extends Long> optionalSafeExternalLong) {
```
lsingh123: again, not sure if we want the parameter to have a safety annotation, or if we don't support wrapping external references in optional

conjure-java-core/src/main/java/com/palantir/conjure/java/util/SafetyUtils.java
```diff
@@ -0,0 +1,50 @@
...
+package com.palantir.conjure.java.util;
+
+import com.palantir.conjure.java.visitor.MoreVisitors;
+import com.palantir.conjure.spec.AliasDefinition;
+import com.palantir.conjure.spec.ArgumentDefinition;
+import com.palantir.conjure.spec.FieldDefinition;
+import com.palantir.conjure.spec.LogSafety;
+import java.util.Optional;
+
+public final class SafetyUtils {
```
lsingh123, line 33: @carterkozak I think you mentioned that this class might need to map a Reference type to an ExternalReference with the appropriate name. I thought the parser [handles](https://github.com/palantir/conjure/blob/master/conjure-core/src/main/java/com/palantir/conjure/defs/ConjureTypeParserVisitor.java#L90-L106) that mapping for us, but I could be missing something

carterkozak: You may be right -- I don't recall the specifics :-)

conjure-java-core/src/integrationInput/java/com/palantir/product/SafeExternalLongExample.java
lsingh123: i think i am actually going to handle optionals wrapping externals, pushing that up soon

carterkozak: Ideally I think we'd end up with `Optional<@Safe ? extends Long>`

lsingh123: Would we want `@Safe Optional<@Safe ? extends Long>` like [here](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/integrationInput/java/com/palantir/product/OptionalExample.java#L100)? Or would we omit the external annotation for external imports?

carterkozak: I think the linked example should ideally only have a single `@Safe`:
```java
public Builder item(@Nonnull Optional<@Safe String> item) {
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -488,9 +492,11 @@ private boolean isPrimitiveOptional(OptionalType optionalType) {
                         optionalType.getItemType().accept(TypeVisitor.PRIMITIVE).get());
     }
 
-    private MethodSpec createItemSetter(EnrichedField enriched, Type itemType, boolean override) {
+    private MethodSpec createItemSetter(
+            EnrichedField enriched, Type itemType, boolean override, Optional<LogSafety> safety) {
         FieldSpec field = enriched.poetSpec();
-        return BeanBuilderAuxiliarySettersUtils.createItemSetterBuilder(enriched, itemType, typeMapper, builderClass)
+        return BeanBuilderAuxiliarySettersUtils.createItemSetterBuilder(
```
lsingh123, line 498: previously, item setters for collection types didn't have safety annotations, even for primitive collections. For example, a field of type `List<boolean>` would generate the following setter:

```
            checkNotBuilt();
            return this;
        }
```

```
            checkNotBuilt();
            return this;
        }
```

conjure-java-core/src/test/java/com/palantir/conjure/java/types/ExternalImportSafetyTests.java
```diff
@@ -0,0 +1,214 @@
...
+                builderSubclass, "addAllSafeExternalLongList", "safeExternalLongList", Safe.class);
+    }
+
+    @Test
+    public void testSetAnnotations() {
+        assertMethodHasAnnotation(SafeExternalLongExample.class, "getSafeExternalLongSet", Safe.class);
+
+        Class<?> builderSubclass = getMatchingSubclass(SafeExternalLongExample.class, "$Builder");
+        assertFieldTypeParamHasAnnotation(builderSubclass, "safeExternalLongSet", "Long", Safe.class);
+        assertMethodParamHasAnnotation(builderSubclass, "safeExternalLongSet", "safeExternalLongSet", Safe.class);
```
lsingh123, line 124: this includes the item setter and the collection setter because the internal method uses allMatch

conjure-java-core/src/integrationInput/java/com/palantir/product/SafeExternalLongExample.java
lsingh123: For posterity:

Poet has a bug that will make handling containers of external imports hard (like optional<long>). We use [WildcardTypeName](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderAuxiliarySettersUtils.java#L99-L129) to represent the wildcard type extending an external import. But when emitting, WildcardTypeName essentially [ignores its annotations](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/WildcardTypeName.java#L64-L71), which I verified in this (extremely rough) [test](https://github.com/palantir/conjure-java/pull/1944/files#diff-cc2d5fa6d815eb2ce4ba4691ac1ceced1dd649598ba8400ab31cc3b1bced1b9bR202-R217). So in order to generate code that looks like `@Safe ? extends Long>` we would have to do something extremely hacky, like eschew javapoet and just stick in a string which doesn't feel......great.

So we decided to settle for the annotation on the wrapper itself, as in `@Safe Optional<...>`

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderAuxiliarySettersUtils.java
```diff
@@ -68,13 +70,19 @@ public static MethodSpec.Builder createOptionalSetterBuilder(
...
     public static MethodSpec.Builder createItemSetterBuilder(
-            EnrichedField enriched, Type itemType, TypeMapper typeMapper, ClassName returnClass) {
+            EnrichedField enriched,
+            Type itemType,
+            TypeMapper typeMapper,
+            ClassName returnClass,
+            Optional<LogSafety> safety) {
         FieldSpec field = enriched.poetSpec();
-        return publicSetter(enriched, returnClass).addParameter(typeMapper.getClassName(itemType), field.name);
+        return publicSetter(enriched, returnClass)
```
lsingh123, line 83: this is to support this https://github.com/palantir/conjure-java/pull/1944/files#r1103276431

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
```diff
@@ -113,7 +115,7 @@ public Optional<LogSafety> visitUnion(UnionDefinition value) {
             return with(value.getTypeName(), () -> {
                 Optional<LogSafety> safety = UNKNOWN_UNION_VARINT_SAFETY;
                 for (FieldDefinition variant : value.getUnion()) {
-                    safety = combine(safety, getSafety(variant.getType(), variant.getSafety()));
+                    safety = combine(safety, getSafety(variant.getType(), SafetyUtils.getMaybeExternalSafety(variant)));
```
carterkozak, line 118: If I understand this correctly, I don't think any of the `SafetyUtils` utility-method invocations used in this class are necessary. This class will walk the tree until it finds something that declares safety, so the updated `visitExternal(ExternalReference value)` should take care of it.

conjure-java-core/src/main/java/com/palantir/conjure/java/util/SafetyUtils.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        return field.getSafety();
+    }
+
+    public static Optional<LogSafety> getMaybeExternalSafety(ArgumentDefinition argument) {
+        if (argument.getType().accept(WrapsExternalType.INSTANCE)) {
+            return argument.getType().accept(GetMaybeExternalSafety.INSTANCE);
+        }
+        return argument.getSafety();
+    }
```
carterkozak, line 56: I might reframe and refactor these a bit to decouple behavior from external types. There's a class of values (applies to fields, arguments, and aliases) which require their safety to be explicitly listed. So, given a `Type`, we need to determine whether the generated class declares safety itself, or if we need to add an annotation where the type is used.

Primitives (with the exception of bearer-token) and external-type-imports require safety at each place they're used. Determining the safety value can be handled using the existing `SafetyEvaluator` class.

conjure-java-core/src/test/resources/external-long-test-service.yml
```diff
@@ -0,0 +1,20 @@
...
+    package: com.palantir.product
+    default-auth: header
+    base-path: /external-long
+
+    endpoints:
+      testExternalLongArg:
+        http: POST /test
+        args:
+          externalLong:
+            type: DangerousLong
```
carterkozak, line 20: Let's add a few more args:
* declared-safe external imported long
* external imported long without any safety information declared
* Alias of each of the existing examples

conjure-java-core/src/test/java/com/palantir/conjure/java/types/ExternalImportSafetyTests.java
```diff
@@ -0,0 +1,229 @@
...
+
+    @Test
+    public void testAliasAnnotations() {
+        assertThat(SafeExternalLongAlias.class).hasAnnotation(Safe.class);
+
+        assertMethodHasAnnotation(SafeExternalLongAlias.class, "toString", Safe.class);
+        assertMethodHasAnnotation(SafeExternalLongAlias.class, "get", Safe.class);
+        assertMethodParamHasAnnotation(SafeExternalLongAlias.class, "valueOf", "value", Safe.class);
+        assertMethodParamHasAnnotation(SafeExternalLongAlias.class, "of", "value", Safe.class);
+    }
```
carterkozak, line 50: I could go either way on reflection-based testing. We have tests which generate `integrationInput`, so future changes will have clear diffs which can be reviewed more clearly than reading reflection code, but that does rely on the reviewer to understand and account for all changes.

lsingh123: Do you think there's any harm in having both? I see your point about the integrationInput being more readable, but I feel like external import safety is kind of brittle, because if someone forgets to use the SafetyUtils wrappers then we lose safety markings. So I thought a test at least provides _some_ in-your-face protection against future breaks

conjure-java-core/src/main/java/com/palantir/conjure/java/util/SafetyUtils.java
lsingh123: Ah, so I could instead think of this class as getting the usage time safety, which, for primitives, is declared in the wrapper type (field, argument, alias) itself and for external imports, is declared in the external-import-type itself.

carterkozak: Yep!

conjure-java-core/src/test/java/com/palantir/conjure/java/types/ExternalImportSafetyTests.java
carterkozak: Totally fine to have both :-)

PR #1945, opened 2023-02-02
read safety of external types

lsingh123:
This PR follows up https://github.com/palantir/conjure/pull/1315 to support log safety annotations for external imports, declared at import time.


conjure-java-core/src/test/java/com/palantir/conjure/java/types/SafetyEvaluatorTest.java
```diff
@@ -150,7 +191,7 @@ void testMapSafetyUnsafeValue() {
                 .types(SAFE_ALIAS)
                 .types(UNSAFE_ALIAS)
                 .build();
-        ConjureDefinitionValidator.validateAll(conjureDef);
+        ConjureDefinitionValidator.validateAll(conjureDef, SafetyDeclarationRequirements.ALLOWED);
```
lsingh123, line 193: this method changed between the previous conjure version and the one I bumped to

conjure-java-core/src/test/java/com/palantir/conjure/java/types/SafetyEvaluatorTest.java
```diff
@@ -320,9 +360,52 @@ void testEmptyEnum() {
...
         SafetyEvaluator evaluator = new SafetyEvaluator(conjureDef);
         assertThat(evaluator.evaluate(Iterables.getOnlyElement(conjureDef.getTypes())))
                 .hasValue(LogSafety.SAFE);
     }
+
+    private static Stream<Arguments> getTypes(Type externalReference) {
+        TypeDefinition objectType = TypeDefinition.object(ObjectDefinition.builder()
```
lsingh123, line 370: maybe it's sufficient to just test objects and the rest is overkill

carterkozak:
Could you update `example-types.yml` to take advantage of this, and regenerate the `integrationInput` sources using `./gradlew test -Drecreate=true`?

Separately (potentially as a separate PR from this one as it may become a little more involved):

I think as written, this will update generated types which contain external imports to reflect the safety of those types, however we probably also want to annotate some of the getter and setter fields with safety annotations since the imported type itself isn't annotated (java.lang.Long doesn't have a safe or unsafe annotation). This way, if we alias a safe external long import, the alias factory method becomes `public static MyAlias of(@Safe long value)`, so that unsafe data cannot be passed in.

lsingh123:
That sounds like a plan - I'll open a separate PR for that change.

PR #1949, opened 2023-02-08
Remove unused jaxrs dependency from conjure-lib

carterkozak:
==COMMIT_MSG==
Remove unused jaxrs dependency from conjure-lib
==COMMIT_MSG==

## Possible downsides?
This has the potential to break consumers who depend on javax.ws.rs-api without declaring their own dependency, however jersey and feign support have been deprecated for several years at this point, so upgrades should be safe.


carterkozak:
major rev time.

bjlaub (approved):

PR #1955, opened 2023-02-10
Distinguish error logging between thrown Errors and ServiceExceptions

ash211:
## Before this PR
When seeing "Error handling request" in logs, it is difficult to distinguish if the problem is a thrown Error or a thrown ServiceException.

## After this PR
==COMMIT_MSG==
Distinguish error logging between thrown Errors and ServiceExceptions
==COMMIT_MSG==

In general, log messages from a class should be unique, so that someone seeing the message and comparing with source code can uniquely identify the location of the log.

### DIFF
--- a/changelog/@unreleased/pr-1955.v2.yml
+++ b/changelog/@unreleased/pr-1955.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Distinguish error logging between thrown Errors and ServiceExceptions
+  links:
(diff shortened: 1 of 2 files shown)


pkoenig10:
I frequently search for `Error handling request` when I want to find exceptions being thrown by my service. When starting my investigation, I don't actual know what kind of exceptions are being thrown. If different exceptions produce different error messages, it is going to be much more difficult to search for and find these errors.

Can you explain what problem you are trying to solve here? You can always determine the exception thrown by looking at the included stacktrace.

PR #1960, opened 2023-02-14
add integration tests that take advantage of safety annotations on external longs

lsingh123:
the generator still ignores the safety of external imports, so generated code won't annotate external imports


conjure-java-core/src/test/java/com/palantir/conjure/java/types/SafetyEvaluatorTest.java
```diff
@@ -130,7 +131,7 @@ void testMapSafeKey() {
                 .types(object)
                 .types(SAFE_ALIAS)
                 .build();
-        ConjureDefinitionValidator.validateAll(conjureDef);
+        ConjureDefinitionValidator.validateAll(conjureDef, SafetyDeclarationRequirements.ALLOWED);
```
lsingh123, line 134: this is because i bumped the conjure parser

conjure-java-core/src/test/resources/external-long-test-service.yml
```diff
@@ -0,0 +1,33 @@
...
+    package: com.palantir.product
+    default-auth: header
+    base-path: /external-long
+
+    endpoints:
+      testExternalLongArg:
+        http: POST /test
+        args:
+          externalLong:
+            type: DangerousLong
```
carterkozak: let's add args of Safe and unmarked external type imports while we're here. Possibly also args for aliases of those as well.

conjure-java-core/src/test/java/com/palantir/conjure/java/types/SafetyEvaluatorTest.java
carterkozak: Ah, I was wondering about that 👍

conjure-java-core/src/test/resources/external-long-test-service.yml
```diff
@@ -0,0 +1,58 @@
...
+      safety: safe
+    Long:
+      external:
+        java: java.lang.Long
+    DangerousLongAlias:
+      alias: DangerousLong
+    SafeExternalLongAlias:
+      alias: SafeExternalLong
+    LongAlias:
+      alias: Long
```
carterkozak:
```suggestion
  definitions:
    default-package: com.palantir.product
    objects:
      DangerousLongAlias:
        alias: DangerousLong
      SafeExternalLongAlias:
        alias: SafeExternalLong
      LongAlias:
        alias: Long
```

carterkozak (commented):
It doesn't look like `ExternalLongTestService.java` or `ExternalLongTestServiceEndpoints.java` has been generated once the yml is fixed up, a `./gradlew :conjure-java-core:test -Drecreate=true` should do the trick

conjure-java-core/src/test/resources/external-long-test-service.yml
lsingh123: oh oops, thanks for the catch

conjure-java-core/src/test/resources/external-long-test-service.yml
```diff
@@ -11,12 +11,15 @@ types:
...
-      alias: Long
+  definitions:
+    default-package: com.palantir.product
+    objects:
+      DangerousLongAliasEndpoint:
+        alias: DangerousLong
+      SafeExternalLongAliasEndpoint:
+        alias: SafeExternalLong
+      LongAliasEndpoint:
+        alias: Long
```
carterkozak: let's rename these to avoid the `Endpoint` suffix since it is very similar to some of the generated code.

carterkozak (approved):
Great stuff, thanks!

PR #1962, opened 2023-02-14
item setters now have safety annotations

lsingh123:
previously, item setters for collection types didn't have safety annotations, even for primitive collections. For example, a field of type List would generate the following setter:

```
            checkNotBuilt();
            return this;
        }
```

```
            checkNotBuilt();
            return this;
        }
```
The first commit on this branch adds a safety annotation to an object with a list field, so look at this diff to see the impact of this change on generated code: https://github.com/lsingh123/conjure-java/commit/80ad12785ffa61b42c592c681c5f9859b387f854


lsingh123:
I am not sure why all these prior commits are showing up on this branch

carterkozak:
It's a result of merging develop into this branch rather than rebasing this onto develop. The result should be the same, the diff looks correct

lsingh123:
yikes ok, i will make sure to rebase in the future. the diff is correct.

PR #1964, opened 2023-02-15
[stacked atop #1962] safety annotations item setters

carterkozak:
### [new changes diff link](https://github.com/palantir/conjure-java/compare/193fe21...c1215f3)
==COMMIT_MSG==
safety annotations item setters
==COMMIT_MSG==


conjure-java-core/src/integrationInput/java/com/palantir/product/BearerTokenAliasExample.java
```diff
@@ -40,7 +40,7 @@ public int hashCode() {
         return value.hashCode();
     }
 
-    public static BearerTokenAliasExample valueOf(String value) {
+    public static BearerTokenAliasExample valueOf(@DoNotLog String value) {
```
carterkozak, line 43: This is the result of using the _computed_ safety on `valueOf(String)` methods, rather than the direct safety.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -120,7 +119,7 @@ public static JavaFile generateAliasType(
             spec.addMethod(MethodSpec.methodBuilder("valueOf")
                     .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
                     .addParameter(ParameterSpec.builder(String.class, "value")
-                            .addAnnotations(safetyAnnotations)
+                            .addAnnotations(computedSafetyAnnotations)
```
carterkozak, line 122: This is the change I mentioned above, noticed while making changes.

conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/OptionalAlias.java
```diff
@@ -25,7 +25,6 @@ private OptionalAlias() {
     }
 
     @JsonValue
-    @Safe
     public Optional<@Safe String> get() {
```
carterkozak, line 28: No need to annotate both the method and return value, this deduplicates our safety annotations to the most specific one.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
lsingh123: I was thinking of making this change but I wasn't sure if this breaks the case where we're aliasing a type that declares its own safety (e.g. reference) - in which case we should be using safetyAnnotations, which will be empty, instead of computedSafetyAnnotations, which would resolve to the safety of the reference type?

lsingh123: Oh wait is this desired because value takes a string?

carterkozak: Exactly! The input via string-parsing is expected to have the same safety as the thing we're creating, but String.class isn't aware of log-safety.

PR #1965, opened 2023-02-15
item setters now have safety annotations

lsingh123:
previously, item setters for collection types didn't have safety annotations, even for primitive collections. For example, a field of type List would generate the following setter:

```
            checkNotBuilt();
            return this;
        }
```

```
            checkNotBuilt();
            return this;
        }
```
The first commit on this branch adds a safety annotation to an object with a list field, so look at this diff to see the impact of this change on generated code: https://github.com/palantir/conjure-java/pull/1965/files/fc14f9ff9aff3da3607050d6c4f0623fac228672..0e49f25f7e4fbd9b92681e5cf2e8eabaaf816e02


conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
```diff
@@ -189,19 +189,19 @@ public Builder from(ListExample other) {
         }
 
         @JsonSetter(value = "items", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
-        public Builder items(@Safe @Nonnull Iterable<String> items) {
+        public Builder items(@Nonnull Iterable<String> items) {
```
lsingh123: weird.. need to investigate

carterkozak: Hmm, I suspect `widenParameterIfPossible` may need to be updated to retain safety info

carterkozak (approved):

mpritham (approved):

PR #1966, opened 2023-02-15
external imports have safety now

lsingh123:
If an Alias, Field, or Argument is an external import type, use the underlying import's safety instead of the declared safety (which will be empty). Converts all usages of AliasDefinition.getSafety(), FieldDefinition.getSafety(), and ArgumentDefinition.getSafety() to the corresponding wrapper methods in SafetyUtils.java.

Tested locally against two large internal repos with complex conjure definitions.


conjure-java-core/src/integrationInput/java/com/palantir/product/SafeExternalLongExample.java
```diff
@@ -158,68 +165,68 @@ public Builder from(SafeExternalLongExample other) {
...
-        public Builder safeExternalLongList(long safeExternalLongList) {
+        public Builder safeExternalLongList(@Safe long safeExternalLongList) {
             checkNotBuilt();
             this.safeExternalLongList.add(safeExternalLongList);
             return this;
         }
 
         @JsonSetter(value = "safeExternalLongSet", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
-        public Builder safeExternalLongSet(@Nonnull Iterable<? extends Long> safeExternalLongSet) {
+        public Builder safeExternalLongSet(@Safe @Nonnull Iterable<? extends Long> safeExternalLongSet) {
```
lsingh123: ideally this would be `@Nonnull Iterable<@Safe ? extends Long>` but there is a [bug](https://github.com/square/javapoet/issues/959) in poet that makes that impossible. i think this is a pretty good workaround.

carterkozak: 👍 looks reasonable to me

conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
```diff
@@ -189,13 +189,13 @@ public Builder from(ListExample other) {
         }
 
         @JsonSetter(value = "items", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
-        public Builder items(@Nonnull Iterable<@Safe String> items) {
+        public Builder items(@Nonnull Iterable<String> items) {
```
lsingh123: oh no where'd you go

conjure-java-core/src/integrationInput/java/com/palantir/product/SafeExternalLongExample.java
```diff
@@ -158,68 +165,68 @@ public Builder from(SafeExternalLongExample other) {
...
+        public Builder safeExternalLongValue(@Safe long safeExternalLongValue) {
             checkNotBuilt();
             this.safeExternalLongValue = safeExternalLongValue;
             this._safeExternalLongValueInitialized = true;
             return this;
         }
 
         @JsonSetter(value = "optionalSafeExternalLong", nulls = Nulls.SKIP)
-        public Builder optionalSafeExternalLong(@Nonnull Optional<? extends Long> optionalSafeExternalLong) {
+        public Builder optionalSafeExternalLong(@Nonnull Optional<? extends @Safe Long> optionalSafeExternalLong) {
```
lsingh123, line 176: this is......technically correct i think

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -446,15 +449,17 @@ private static CodeBlock primitiveSafeEquality(
...
         return CodeBlocks.statement("return value.toString()");
     }
 
     private static CodeBlock primitiveSafeHashCode(TypeName aliasTypeName) {
-        if (aliasTypeName.isPrimitive()) {
-            return CodeBlocks.statement("return $T.hashCode(value)", aliasTypeName.box());
+        if (PrimitiveHelpers.isPrimitive(aliasTypeName)) {
+            return CodeBlocks.statement(
+                    "return $T.hashCode(value)",
+                    PrimitiveHelpers.box(aliasTypeName).withoutAnnotations());
```
lsingh123: before you could add safety to external longs, you were never going to end up with an aliasTypeName that is primitive for all intents and purposes (e.g. java.lang.Long) but is annotated. now, that situation is possible so we need to strip the annotation here to avoid the extremely weird `@Safe Long.hashCode(value)`

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -82,7 +85,7 @@ private BeanGenerator() {}
     /** The name of the singleton instance field generated for empty types. */
     private static final String SINGLETON_INSTANCE_NAME = "INSTANCE";
 
-    @SuppressWarnings("CyclomaticComplexity")
+    @SuppressWarnings({"CyclomaticComplexity", "MethodLength"})
```
lsingh123: did not want to add a refactor of this method onto this already large PR

conjure-java-core/src/main/java/com/palantir/conjure/java/util/PrimitiveHelpers.java
```diff
@@ -0,0 +1,68 @@
...
+            return type;
+        }
+    }
+
+    public static boolean isPrimitive(TypeName type) {
+        return getPrimitiveType(type).isPresent();
+    }
+
+    public static boolean isBoxedPrimitive(TypeName type) {
+        return type.withoutAnnotations().isBoxedPrimitive();
```
lsingh123: poet's isBoxedPrimitive method uses object equality but its isPrimitive method uses reference equality

conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
lsingh123: found it

conjure-java-core/src/integrationInput/java/com/palantir/product/SafeExternalLongExample.java
carterkozak: 👍 Looks good to me

conjure-java-core/src/main/java/com/palantir/conjure/java/util/SafetyUtils.java
```diff
@@ -0,0 +1,108 @@
...
+import com.palantir.conjure.spec.Type;
+import com.palantir.conjure.spec.TypeName;
+import java.util.Optional;
+
+public final class SafetyUtils {
+    private SafetyUtils() {}
+
+    // SafetyEvaluator uses its TypeDefinitionMap to resolve the safety of reference types, but these utils
+    // only evaluate the safety of external imports and primitives, so it's safe to use a placeholder
+    private static final SafetyEvaluator safetyEvaluator = new SafetyEvaluator(ImmutableMap.of());
```
carterkozak: As al alternative, we could make these methods non-static and add them to the SafetyEvaluator -- that may be more robust (or simply add the static methods to SafetyEvaluator so the safety evaluation pieces live in the same place)

conjure-java-core/src/main/java/com/palantir/conjure/java/util/PrimitiveHelpers.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.squareup.javapoet.AnnotationSpec;
+import com.squareup.javapoet.TypeName;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public final class PrimitiveHelpers {
+
+    private PrimitiveHelpers() {}
```
carterkozak: Pedantic naming nit: Rather than `Helper` or `Util` prefixes, we generally use plural of the target, e.g. `Primitives`.

conjure-java-core/src/main/java/com/palantir/conjure/java/util/SafetyUtils.java
lsingh123: I was trying to avoid plumbing through a SafetyEvaluator instance everywhere these utils are used, but it's probably a good idea to do so

carterkozak: I think we already have one plumbed in a bunch of places where it's not actually used, heh. If that turns out to require a great deal of plumbing, we could use static methods on SafetyEvaluator, and defer the plumbing to a follow-up commit. Up to you.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
```diff
@@ -78,6 +79,27 @@ public Optional<LogSafety> evaluate(Type type, Optional<LogSafety> declaredSafet
         return declaredSafety.or(() -> evaluate(type));
     }
 
+    public Optional<LogSafety> getUsageTimeSafety(ArgumentDefinition argument) {
```
carterkozak, line 87: We should add some javadoc to these methods since it's not entirely obvious when they should be used without the full context of this change

carterkozak:
Aside from the javadoc request, this looks good to me -- great work!

carterkozak (approved):

PR #1967, opened 2023-02-16
use our own boxing and unboxing logic instead of poet's

lsingh123:
poet's boxing logic is incompatible with annotated primitives, resulting in breaks like [this one](https://app.circle.palantir.build/pipelines/github/geotime/geotime/6903/workflows/27a4dde3-9468-46aa-9c8d-b85b2341079b/jobs/313949/steps), which is repro'd in the first commit on this branch. use our own boxing logic to avoid this error

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/util/PrimitiveHelpers.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/util/PrimitiveHelpers.java
@@ -1,0 +1,68 @@
+/*
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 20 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/util/PrimitiveHelpers.java
```diff
@@ -0,0 +1,68 @@
...
+        return getPrimitiveType(type).isPresent();
+    }
+
+    public static boolean isBoxedPrimitive(TypeName type) {
+        return type.withoutAnnotations().isBoxedPrimitive();
+    }
+
+    public static TypeName unbox(TypeName type) {
+        List<AnnotationSpec> annotations = type.annotations;
+        return type.withoutAnnotations().unbox().annotated(annotations);
```
lsingh123, line 59: this is fine because poet checks equality by object for boxed primitives but by reference for primitives

conjure-java-core/src/test/resources/example-types.yml
```diff
@@ -222,6 +222,9 @@ types:
           listAlias: ListAlias
           setAlias: SetAlias
           mapAlias: MapAliasExample
+          booleanField:
```
lsingh123, line 225: this is what was breaking in geotime

lsingh123: https://github.palantir.build/geotime/geotime/blob/develop/geotime-api/src/main/conjure/geotime-search.yml#L108-L111

carterkozak (approved):
Thanks for putting this together so quickly!

PR #1968, opened 2023-02-16
remove an extraneous annotation

lsingh123:
first commit repros the bug if you want to see the failure. tested locally on large repos with complex conjure apis

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/SafeLongAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/SafeLongAliasExample.java
@@ -4,24 +4,27 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.conjure.java.lib.SafeLong;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.Safe;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.processing.Generated;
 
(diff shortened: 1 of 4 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/SafeLongAliasExample.java
```diff
@@ -42,12 +45,12 @@ public int compareTo(SafeLongAliasExample other) {
         return value.compareTo(other.get());
     }
 
-    public static SafeLongAliasExample valueOf(String value) {
+    public static SafeLongAliasExample valueOf(@Safe String value) {
         return of(SafeLong.valueOf(value));
```
lsingh123, line 49: the bug is that this line used to read `of(@Safe SafeLong.valueOf(value));`, which the formatter didn't like

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -360,34 +360,35 @@ && hasValueOfFactory(reference.getExternalReference())
 
     @SuppressWarnings("checkstyle:cyclomaticcomplexity")
     private static CodeBlock valueOfFactoryMethodForPrimitive(PrimitiveType primitiveType, TypeName aliasTypeName) {
+        TypeName boxedTypeName = PrimitiveHelpers.box(aliasTypeName).withoutAnnotations();
```
carterkozak, line 363: Nice!

carterkozak (approved):
thanks!

PR #1970, opened 2023-02-24
Require non-null elements in collection body params

pkoenig10:
## Before this PR

Services generated using `conjure-undertow-annotations` allow null elements in top-level collection body parameters.

## After this PR

Services generated using `conjure-undertow-annotations` do not allow null elements in top-level collection body parameters.

This is similar to https://github.com/palantir/conjure-java/pull/1051, but for `conjure-undertow-annotations`.

Unlike the `nonNullCollections` flag, this behavior is not configurable. This matches the behavior of the built-in collection deserializers for query/header params, which do not allow null elements. If users want nullable elements, they should use `Optional`.

Unlike https://github.com/palantir/conjure-java/pull/1051, this should be safe to apply by default because it does not affect clients of the API, only servers.

## Possible downsides?

We may break services that were allowing requests with null elements. However it is very likely that the authors did not intended to allow null elements. Allowing null elements can results in data correctness issues in the future if the services do not perform their own validation. It is better to fail loudly in these cases than to silently allow bad input.

### DIFF
--- a/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
+++ b/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
@@ -64,6 +64,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import javax.annotation.processing.Generated;
 import javax.lang.model.element.Modifier;
 import org.immutables.value.Value;
@@ -179,8 +180,9 @@ public final class ConjureUndertowEndpointsGenerator {
                     CodeBlock deserializerFactory,
                     String deserializerFieldName,
                     SafeLoggingAnnotation safeLoggable) {
-                TypeName requestBodyType =
-                        def.argType().match(ArgTypeTypeName.INSTANCE).box();
+                TypeName requestBodyType = immutableCollection(
+                        def.argType().match(ArgTypeTypeName.INSTANCE).box());
+
                 additionalFields.add(ImmutableAdditionalField.builder()
                         .field(FieldSpec.builder(
                                         ParameterizedTypeName.get(ClassName.get(Deserializer.class), requestBodyType),
@@ -753,6 +755,26 @@ public final class ConjureUndertowEndpointsGenerator {
             default:
                 throw new SafeIllegalStateException("Illegal value", SafeArg.of("value", safeLoggable));
         }
+    }
+
+    private static final ClassName LIST_NAME = ClassName.get(List.class);
+    private static final ClassName IMMUTABLE_LIST_NAME = ClassName.get(ImmutableList.class);
+    private static final ClassName SET_NAME = ClassName.get(Set.class);
+    private static final ClassName IMMUTABLE_SET_NAME = ClassName.get(ImmutableSet.class);
+
+    private static TypeName immutableCollection(TypeName input) {
+        if (input instanceof ParameterizedTypeName) {
+            ParameterizedTypeName parameterized = (ParameterizedTypeName) input;
+            if (LIST_NAME.equals(parameterized.rawType)) {
+                return ParameterizedTypeName.get(
+                        IMMUTABLE_LIST_NAME, parameterized.typeArguments.toArray(new TypeName[0]));
+            } else if (SET_NAME.equals(parameterized.rawType)) {
+                return ParameterizedTypeName.get(
+                        IMMUTABLE_SET_NAME, parameterized.typeArguments.toArray(new TypeName[0]));
+            }
+        }
+
+        return input;
     }
 
     @Value.Immutable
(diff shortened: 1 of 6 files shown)


conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
```diff
@@ -755,6 +757,26 @@ private static Optional<CodeBlock> getSafeLogging(
...
+                return ParameterizedTypeName.get(
+                        IMMUTABLE_LIST_NAME, parameterized.typeArguments.toArray(new TypeName[0]));
+            } else if (SET_NAME.equals(parameterized.rawType)) {
+                return ParameterizedTypeName.get(
+                        IMMUTABLE_SET_NAME, parameterized.typeArguments.toArray(new TypeName[0]));
+            }
+        }
+
+        return input;
+    }
```
pkoenig10: I couldn't find a good place to put this shared logic, so I just copied it from `UndertowServiceHandlerGenerator`. It's fairly trivial and we may want these implementations to diverge in the future.

https://github.com/palantir/conjure-java/blob/260ed601c7faf034aa7cfba1dafc368078fa1795/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java#L392-L412

I'm happy to move this to a shared helper method if there's a place we feel is appropriate.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
```diff
@@ -755,6 +757,26 @@ private static Optional<CodeBlock> getSafeLogging(
...
 
+    private static final ClassName LIST_NAME = ClassName.get(List.class);
+    private static final ClassName IMMUTABLE_LIST_NAME = ClassName.get(ImmutableList.class);
+    private static final ClassName SET_NAME = ClassName.get(Set.class);
+    private static final ClassName IMMUTABLE_SET_NAME = ClassName.get(ImmutableSet.class);
+
+    private static TypeName immutableCollection(TypeName input) {
+        if (input instanceof ParameterizedTypeName) {
+            ParameterizedTypeName parameterized = (ParameterizedTypeName) input;
+            if (LIST_NAME.equals(parameterized.rawType)) {
```
carterkozak: Perhaps `Collection` and maybe `Iterable` should become `ImmutableList`?

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
carterkozak: :+1: I don't mind code duplication, it's much nicer than unnecessarily coupling. The edge cases are dramatically different here vs pure conjure.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
carterkozak: Maybe Map as well, but I'd hope that's less common and may not be necessary.

pkoenig10: I've added `Collection`, `Map`, and a couple more collection types.

`Iterable` feels a bit weird, I'd slightly prefer to not support that.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -405,6 +408,9 @@ private TypeName immutableCollection(TypeName input) {
             } else if (SET_NAME.equals(parameterized.rawType)) {
                 return ParameterizedTypeName.get(
                         IMMUTABLE_SET_NAME, parameterized.typeArguments.toArray(new TypeName[0]));
+            } else if (MAP_NAME.equals(parameterized.rawType)) {
+                return ParameterizedTypeName.get(
+                        IMMUTABLE_MAP_NAME, parameterized.typeArguments.toArray(new TypeName[0]));
```
pkoenig10, line 413: This appears to have been an oversight in https://github.com/palantir/conjure-java/pull/1051, so I've added it here.

pkoenig10: Does the concern from https://github.com/palantir/conjure-java/pull/1051#discussion_r473081559 still apply? If you've opted in to `nonNullCollections` it doesn't feel unreasonable to disallow null keys.

I think this is fairly unlikely to break people in practice.

carterkozak: We have a few edge cases that make it scarier, particularly around floating point values which lose precision into equality. Ideally we would validate it, but changing behavior behind the existing flag is riskier -- Lots of projects do lots of things with conujre, I suspect we would break at least a few of them. I don't mind a bit more risk in the annotation processor, but I'd prefer not to introduce that risk into our conjure codegen.

pkoenig10: Ack, will revert this.

It would be nice if there were some way for consumers to opt-in to this.

carterkozak (approved):
Thanks!

PR #1974, opened 2023-03-06
Fix alias-of-optional codegen with safety info

carterkozak:
## Before this PR
Previously aliases such as the newly added test:
```yaml
alias: optional<double>
safety:unsafe
```

Failed to compile with:
```
<identifier> expected
        this(java.util. @Unsafe OptionalDouble.empty());
```

## After this PR
==COMMIT_MSG==
Fix alias-of-optional codegen with safety info
==COMMIT_MSG==

## Possible downsides?
none

### DIFF
--- a/changelog/@unreleased/pr-1974.v2.yml
+++ b/changelog/@unreleased/pr-1974.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix alias-of-optional codegen with safety info
+  links:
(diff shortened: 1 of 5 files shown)


lsingh123 (approved):
nice catch! this would probably also fix the same bug for OptionalLong right?

carterkozak:
Yep!

mpritham (approved):

PR #1976, opened 2023-03-07
idea - do not merge: Refactoring `BeanGenerator` and `BeanBuilderGenerator`

mpritham:
## High level ideas
- `BeanGenerator::generateBeanType` is a long, and complex method
  - I think we can have a simpler method by delegating some of the logic to helper methods, and other modules 
- There exists `BeanBuilderGenerator` containing some methods used to generate builders, but still much of the staged builder generation logic is in `BeanGenerator`.
  - Staged builders are a specific type of builder, and I think we should move the stage builder generation logic into `BeanBuilderGenerator` or create a new file `BeanStagedBuilderGenerator`

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
@@ -93,17 +93,11 @@ public final class BeanGenerator {
             Options options) {
         ImmutableList<EnrichedField> fields = createFields(typeMapper, typeDef.getFields());
         ImmutableList<FieldSpec> poetFields = EnrichedField.toPoetSpecs(fields);
-        ImmutableList<EnrichedField> nonPrimitiveEnrichedFields =
-                fields.stream().filter(field -> !field.isPrimitive()).collect(ImmutableList.toImmutableList());
+        boolean hasFields = !fields.isEmpty();
 
         com.palantir.conjure.spec.TypeName prefixedName =
                 Packages.getPrefixedName(typeDef.getTypeName(), options.packagePrefix());
         ClassName objectClass = ClassName.get(prefixedName.getPackage(), prefixedName.getName());
-        ClassName builderClass = ClassName.get(objectClass.packageName(), objectClass.simpleName(), "Builder");
-        ClassName builderImplementation =
-                options.useStagedBuilders() && fields.stream().anyMatch(field -> !fieldShouldBeInFinalStage(field))
-                        ? ClassName.get(objectClass.packageName(), objectClass.simpleName(), "DefaultBuilder")
-                        : builderClass;
 
         // Use the fully-resolved/computed safety value for the top-level class, however
         // fields (setters/getters) are annotated with declared values because complex objects
@@ -113,41 +107,9 @@ public final class BeanGenerator {
 
         TypeSpec.Builder typeBuilder = TypeSpec.classBuilder(prefixedName.getName())
                 .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
-                .addAnnotations(safety)
-                .addFields(poetFields)
-                .addMethod(createConstructor(fields, poetFields))
-                .addMethods(createGetters(fields, typesMap, options, safetyEvaluator));
+                .addAnnotations(safety);
 
-        if (!poetFields.isEmpty()) {
-            boolean useCachedHashCode = useCachedHashCode(fields);
-            typeBuilder
-                    .addMethod(MethodSpecs.createEquals(objectClass))
-                    .addMethod(MethodSpecs.createEqualTo(objectClass, poetFields, useCachedHashCode));
-            if (useCachedHashCode) {
-                MethodSpecs.addCachedHashCode(typeBuilder, poetFields);
-            } else {
-                typeBuilder.addMethod(MethodSpecs.createHashCode(poetFields));
-            }
-        }
-
-        typeBuilder.addMethod(MethodSpecs.createToString(
-                        prefixedName.getName(),
-                        fields.stream().map(EnrichedField::fieldName).collect(Collectors.toList()))
-                .toBuilder()
-                .addAnnotations(safety)
-                .build());
-
-        if (poetFields.size() <= MAX_NUM_PARAMS_FOR_FACTORY) {
-            typeBuilder.addMethod(createStaticFactoryMethod(fields, objectClass, safetyEvaluator));
-        }
-
-        if (!nonPrimitiveEnrichedFields.isEmpty()) {
-            typeBuilder
-                    .addMethod(createValidateFields(nonPrimitiveEnrichedFields))
-                    .addMethod(createAddFieldIfMissing(nonPrimitiveEnrichedFields.size()));
-        }
-
-        if (poetFields.isEmpty()) {
+        if (!hasFields) {
             // Need to add JsonSerialize annotation which indicates that the empty bean serializer should be used to
             // serialize this class. Without this annotation no serializer will be set for this class, thus preventing
             // serialization.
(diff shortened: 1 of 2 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -113,125 +108,67 @@ public static JavaFile generateBeanType(
...
+            typeBuilder.addAnnotation(JsonSerialize.class).addField(createSingletonField(objectClass));
+            if (!options.strictObjects()) {
+                typeBuilder.addAnnotation(AnnotationSpec.builder(JsonIgnoreProperties.class)
+                        .addMember("ignoreUnknown", "$L", true)
+                        .build());
             }
         }
 
+        typeBuilder.addFields(poetFields);
+        addConstructor(typeBuilder, fields);
```
mpritham, line 126: Compared to `typeBuilder.addMethod(createConstructor)`, this approach actually seems worse: we have no guarantee that `addConstructor` doesn't add more than one method.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -113,125 +108,67 @@ public static JavaFile generateBeanType(
...
+        addGetters(
+                typeBuilder,
+                fields,
+                typesMap,
+                options.excludeEmptyOptionals(),
+                options.excludeEmptyCollections(),
+                options.useImmutableBytes(),
```
mpritham, line 133: Would prefer to make explicit which flags are being used.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -113,125 +108,67 @@ public static JavaFile generateBeanType(
...
+                safetyEvaluator);
+
+        if (hasFields) {
+            addEquals(typeBuilder, fields, objectClass);
```
mpritham, line 137: One idea is to have a method like `addEquals_IfExistsFields` which does the `hasFields` check internally. Given this is a simple check, it seems ok to do it at the caller.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -113,125 +108,67 @@ public static JavaFile generateBeanType(
...
-                                objectClass,
-                                builderClass,
-                                typeDef,
-                                typesMap,
-                                options,
-                                Optional.of(ClassName.get(
-                                        objectClass.packageName(), objectClass.simpleName(), builderInterface.name))));
-            }
+        if (hasFields) {
+            addBuilder(typeBuilder, objectClass, fields, typeMapper, safetyEvaluator, typeDef, typesMap, options);
```
mpritham, line 170: Motivation:
I don't think we expect to ever use `generateStagedInterfaces` without also generating all of the remaining methods. It seems reasonable to have one method do all of the required generation to add staged builders.
 
```java
typeBuilder
     .addTypes(interfaces)
     .addType(builderInterface)
     .addMethod(MethodSpec.methodBuilder("builder")
     .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
     .returns(interfacesAsClasses.get(0))
     .addStatement("return new DefaultBuilder()")
     .build())
.addType(BeanBuilderGenerator.generate(...));
```

I don't like the fact that `addBuilder` has side effects, but given the method is either in this class, or some other tightly coupled class, it seems ok.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -420,13 +482,21 @@ private static ImmutableList<EnrichedField> createFields(TypeMapper typeMapper,
...
-        Collection<FieldSpec> nonPrimitivePoetFields =
-                Collections2.filter(poetFields, f -> !Primitives.isPrimitive(f.type));
-        if (!nonPrimitivePoetFields.isEmpty()) {
-            builder.addStatement("$L", Expressions.localMethodCall("validateFields", nonPrimitivePoetFields));
+        Collection<EnrichedField> nonPrimitiveEnrichedFields = Collections2.filter(fields, f -> !f.isPrimitive());
+        if (!nonPrimitiveEnrichedFields.isEmpty()) {
+            builder.addStatement(
+                    "$L",
+                    Expressions.localMethodCall(
+                            "validateFields", EnrichedField.toPoetSpecs(nonPrimitiveEnrichedFields)));
```
mpritham, line 499: Not the point of this PR. Can ignore.

---

I removed `poetFields` as an argument, which requires me to call `EnrichedField.toPoetSpecs` and do an iteration over the `nonPrimitiveEnrichedFields` array. It's probably not a big performance hit(?), but I'm still leaning towards not introducing this change.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
carterkozak: Part of the issue here is that the builder was originally a single concrete `Builder` class enclosed within the bean type, however Staged builders require N builder interfaces, which are also nested within the bean type, not contained in the builder, so `BeanBuilderGenerator.generate` can no longer simply return a type. It could consume the outer typeBuilder instead, or spec out an object which represents N additions (a list of new typespecs, methods, fields, etc in a more pure way than mutating a builder instance).

mpritham: > consume the outer typeBuilder instead

Yep. I think that maps to having the `addBuilder` method here, be in `BeanBuilderGenerator`.

> spec out an object which represents N additions

I imagine this would look something like:
```java
public record GeneratedStatements (
    List<AnnotationSpec> annotations, 
    List<InterfaceSpec> interfaces, 
    // etc.
    ...) {}
```

then in our caller's code:
```java
stmts = BeanBuilderGenerator.generateStagedBuilder(...)
typeBuilder.addMethods(stmts.methods).addInterfaces(stmts.interfaces)...
```

If something like that is what you have in mind, I think we'd have to be careful about the insertion order of statements in `BeanBuilderGenerator`.

But I think it's a little annoying that we need to reflect the ordering of the statements in the generated code, by the call order of `typeBuilder.addXYZ`, and can't pass in our own ordering.

mpritham: Ah I guess we could have some union type `Union[AnnotationSpec, InterfaceSpec, ...]` which would allow me to just store one list instead of many. Still think passing in the `builder` is more straightforward.

> pass in our own ordering

thinking a bit more about what this might look like, maybe we could tag each poet spec with some string/enum/class, and provide a list of those string/enum/class as an ordering.

maybe:
```java
public record Statement<T> (T type, Union[AnnotationSpec, InterfaceSpec, ...] spec) {
}

public record GeneratedStatements<T> (
    List<Statement<T>> statements, 
    // maybe some other metadata
    ...) {}

public <StatementT> TypeBuilder.Builder generate(GeneratedStatements<StatementT> stmts, List<StatementT> order) {
    // stable sort stmts.statements according to the type's index in `order` 
    // for stmt : sortedStmts: add the statement to builder (unsure exactly how. maybe use a visitor) 
}
```

then in a generator class:
```java
// allow each "generator" class to tag statements with their own values
class MyGeneratedStatementType extends GeneratedStatementType {
    MY_ANNOTATION, 
    MY_INTERFACE,
    SOME_MODIFIER,
    //...
}

List<Statement<MyGeneratedStatementType>> stmts = new ArrayList<>();
stmts.add(Statement(MY_ANNOTATION, annotationSpec));
stmts.add(Statement(MY_INTERFACE, interfaceSpec))
generate(typeBuilder, stmts, List.of(MY_ANNOTATION, SOME_MODIFIER, MY_INTERFACE, ...)).build();
```

There are likely better ways to implement this, but I'm not sure how annoying the ordering depending on `typeBulider` call order is to others. Leaning towards the view that it's not worth introducing all this complexity.

PR #1978, opened 2023-03-07
Methods in staged builder interfaces are ordered deterministically

mpritham:
## Before this PR
Methods to be added to a staged builder interface were collected in a set. There are no guarantees on order when iterating a set. This could lead to different code being generated for the same definitions.

## After this PR
==COMMIT_MSG==
Collect methods to be added to a staged builder interface in a list
==COMMIT_MSG==


mpritham:
Ah, no changelog auto-generation because this is on a fork. Will re-open as a branch.

PR #1979, opened 2023-03-07
Methods in staged builder interfaces are ordered deterministically

mpritham:
## Before this PR
Methods to be added to a staged builder interface were collected in a set. JavaPoet emits method definitions in the order in which they are added in `addMethod(s)` calls. Sets are not ordered collections, and don't have guarantees of deterministic ordering when iterated over. This could lead to different code being generated for the same definitions.

## After this PR
==COMMIT_MSG==
Collect methods to be added to a staged builder interface in a list
==COMMIT_MSG==


carterkozak (approved):
Great find!

PR #1980, opened 2023-03-08
Refactor `BeanBuilderGenerator` to handle all builder generation logic

mpritham:
**Note**: Would like to merge #1981 before this PR.

## Before this PR
`BeanGenerator::generateBeanType` is a long, and complex method. While there exists `BeanBuilderGenerator` containing some methods used to generate builders, much of the staged builder generation code is in `BeanGenerator`.

## After this PR
Staged builders are a specific type of builder, and I think we should move all builder generation logic into `BeanBuilderGenerator`.

This PR:
- Moves all builder and staged builder generation logic in `BeanGenerator` into `BeanBuilderGenerator`
- Expose methods `BeanBuilderGenerator::addBuilder` and `BeanBuilderGenerator::addStagedBuilder`, which are called in `BeanGenerator`

==COMMIT_MSG==
Refactor BeanBuilderGenerator to handle all builder generation logic
==COMMIT_MSG==

## Possible downsides?
`BeanBuilderGenerator::addBuilder` and `BeanBuilderGenerator::addStagedBuilder` both have side effects (they modify the `TypeSpec.Builder` object passed in). It's possible that future developers introduce subtle errors because of this. However, given that `BeanGenerator` and `BeanBuilderGenerator` are tightly coupled (e.g. they share the `EnrichedField` structure), I think this it's reasonable to expect contributors to have enough context on the effects of these `add...` methods in `BeanGenerator`.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
@@ -102,7 +102,9 @@ public final class BeanBuilderGenerator {
     private static final String BUILT_FIELD = "_buildInvoked";
     private static final String CHECK_NOT_BUILT_METHOD = "checkNotBuilt";
     private static final String BUILDER_IMPLEMENTATION_NAME = "Builder";
+    /* The name of the interface extending all stage interfaces. */
     private static final String STAGED_BUILDER_INTERFACE_NAME = "Builder";
+    /* The name of the class implementing the interface extending all stage interfaces. */
     private static final String STAGED_BUILDER_IMPLEMENTATION_NAME = "DefaultBuilder";
 
     private final TypeMapper typeMapper;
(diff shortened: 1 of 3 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -115,53 +126,301 @@ private BeanBuilderGenerator(
...
+                .addSuperinterfaces(interfaceClassNames)
+                .build();
+
+        specBuilder
+                .addTypes(interfaces)
+                .addType(builderInterface)
+                .addMethod(MethodSpec.methodBuilder("builder")
+                        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+                        .returns(interfaceClassNames.get(0))
+                        .addStatement(String.format("return new %s()", STAGED_BUILDER_IMPLEMENTATION_NAME))
```
carterkozak: Javapoet supports templating along these lines:
```suggestion
                        .addStatement("return new $N()", STAGED_BUILDER_IMPLEMENTATION_NAME)
```

Docs in the readme: https://github.com/square/javapoet#javapoet

mpritham: nice!

mpritham: Nice!

carterkozak (approved):
Definitely easier to follow now, good work :-)

carterkozak:
This will need a rebase now that the sorting PR has merged

PR #1981, opened 2023-03-08
Staged Builders: Remove unneeded sort, and only sort arguments of static factory when using staged builders

mpritham:
## Before this PR
1. When creating the interfaces for the fields that require intermediate stages, [we sort the fields placing fields set in the final/"completed_" stage, last](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java#L278). This sort doesn't do anything since we filter out all fields in the final stage before passing the list of fields to this method.
2. When generating static factory methods, we sort the order in which fields are built (i.e. the order of `a, b, c` in `builder.buildA(a).buildB(b).buildC(c)`), placing collections and optional types last. As an existing comment implies, this is only required for staged builders.

## After this PR
- Remove the unneeded sorting of fields when constructing staged interfaces
- Only alter the order in which fields are built in the static factory method, when using staged builders.
- Add a test which uses staged builders to create an object with a collection field declared before a "plain" field. The generated static factory method builds the "plain" field before the collection field. 

==COMMIT_MSG==
Staged Builders: Remove unneeded sort, and only sort arguments of static factory when using staged builders
==COMMIT_MSG==

## Possible downsides?
.


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -557,7 +551,10 @@ private static MethodSpec createStaticFactoryMethod(
                     .addAnnotations(ConjureAnnotations.safety(safetyEvaluator.getUsageTimeSafety(field.conjureDef())))
                     .build()));
             // Follow order on adding methods on builder to comply with staged builders option if set
-            sortedEnrichedFields(fields).map(EnrichedField::poetSpec).forEach(spec -> {
+            Stream<EnrichedField> methodArgs = useStagedBuilders
+                    ? fields.stream().sorted(Comparator.comparing(BeanGenerator::fieldShouldBeInFinalStage))
```
mpritham, line 559: Since `fields` is an `ImmutableList`, I think `fields.stream()` would be an ordered stream, the sort would be stable, and we don't need `thenComparing(fields::indexOf)`.

carterkozak: I could go either way on this, relying on the input being an ordered stream is something that could go sideways unexpectedly in a refactor, but I'd hope we have enough test cases to catch it. Worth leaving a comment inline.

Avoiding the sort is less clear since increases branching within the method. We could pass in an enum comparator and always sort (even if the non-staged case uses a comparator which always returns zero), but at that point we might as well always sort for staged builders until we have a strict-staged option. What do you think?

conjure-java-core/src/integrationInput/java/com/palantir/product/MultipleFieldsOneFinalStage.java
```diff
@@ -0,0 +1,286 @@
...
+        return "MultipleFieldsOneFinalStage{mappedRids: " + mappedRids + ", token: " + token + ", optionalItem: "
+                + optionalItem + '}';
+    }
+
+    public static MultipleFieldsOneFinalStage of(
+            Map<ResourceIdentifier, String> mappedRids, OneField token, OneField optionalItem) {
+        return builder()
+                .token(token)
+                .mappedRids(mappedRids)
+                .optionalItem(Optional.of(optionalItem))
```
mpritham, line 107: order looks good.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
mpritham: > I could go either way on this, relying on the input being an ordered stream is something that could go sideways unexpectedly in a refactor, but I'd hope we have enough test cases to catch it. Worth leaving a comment inline.

Yeah that makes sense. Can drop a comment.

> We could pass in an enum comparator and always sort (even if the non-staged case uses a comparator which always returns zero)

We'd have to select which comparator to pass at the call site, which would be above in this class. 

> at that point we might as well always sort for staged builders until we have a strict-staged option. What do you think?

Yep. This PR currently always sorts for staged builders.

I intend to change this logic in a later PR:

I think the ordering we want here is the ordering required due to the stage interfaces, which is related to builders. I think it'd be nice to have this detail live in `BeanBuilderGenerator`, with a method like `getOrderedFieldsForBuilder(fields, options)`, (which would return the sorted list for non-strict staged builders, and just `fields` for other cases).

I think this method will make more sense when strict staged builders are introduced, or at least after the refactoring PR (#1980).

Would be curious about your thoughts on that approach.

carterkozak (approved):

PR #1987, opened 2023-03-12
Add @immutable to generated classes

schlosna:
## Before this PR
Generated classes were not annotated as `@Immutable`

## After this PR

==COMMIT_MSG==
Add @immutable to generated classes
==COMMIT_MSG==


schlosna:
closing out, unfortunately, we can't currently prove the beans' properties are transitively immutable

PR #1989, opened 2023-03-13
Fix regression causing union visitor primitive functions not to be used when safety is specified

carterkozak:
First commit reproduces the issue, second commit fixes it.

==COMMIT_MSG==
Fix regression which resulted in `IntFunction` and `DoubleFunction` not being used when log-safety information is provided
==COMMIT_MSG==


mpritham (approved):
lgtm 👍

PR #1992, opened 2023-03-15
Add strict staged builders

mpritham:
## Before this PR
Currently, Conjure has a `--useStagedBuilders` flag which requires all _required_ fields to be set, where fields of type list, set, map, and optional, are not _required_.

It is valuable to have the ability to enforce that _all_ fields (regardless of type) are set.

## After this PR
==COMMIT_MSG==
Add --useStrictStagedBuilders
==COMMIT_MSG==


conjure-java-core/src/integrationInput/java/com/palantir/product/StrictOneCollectionField.java
```diff
@@ -0,0 +1,150 @@
...
+        Builder from(StrictOneCollectionField other);
+    }
+
+    public interface Completed_StageBuilder {
+        StrictOneCollectionField build();
+    }
+
+    public interface Builder extends MyListStageBuilder, Completed_StageBuilder {
+        @Override
+        Builder myList(@Nonnull List<String> myList);
```
mpritham: There's an argument that this should take an `Iterable` parameter, and we're being too _strict_ with strict builders. I think it's ok to enforce that the provided argument is exactly the type defined in the definition. Let me know your thoughts.

conjure-java-core/src/integrationInput/java/com/palantir/product/StrictOneCollectionField.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        @Override
+        @JsonSetter(value = "myList", nulls = Nulls.SKIP)
+        public Builder myList(@Nonnull List<String> myList) {
+            checkNotBuilt();
+            this.myList = ConjureCollections.newArrayList(Preconditions.checkNotNull(myList, "myList cannot be null"));
+            return this;
+        }
+
```
mpritham, line 138: I've removed the auxiliary methods for lists (e.g. `addAllX`). I think it makes sense to enforce users that users not mutate the fields after construction when using strict staged builders. What do you think?

edit: and one auxiliary method for optional types, a value setter i.e. `Builder x(int value) {this.x = Optional.of(value);}`

conjure-java-core/src/integrationInput/java/com/palantir/product/StrictOneCollectionField.java
carterkozak: I don't think it matters what shape of iterable is provided, the important thing is that the data is set in some way (no reason to force folks to transform collection->list themselves)

conjure-java-core/src/integrationInput/java/com/palantir/product/StrictFourFields.java
```diff
@@ -0,0 +1,258 @@
...
+        @JsonSetter(value = "myList", nulls = Nulls.SKIP)
+        public Builder myList(@Nonnull List<String> myList) {
+            checkNotBuilt();
+            this.myList = ConjureCollections.newArrayList(Preconditions.checkNotNull(myList, "myList cannot be null"));
+            return this;
+        }
+
+        @Override
+        @JsonSetter("bearerTokenValue")
+        public Builder bearerTokenValue(@Nonnull BearerToken bearerTokenValue) {
```
mpritham, line 230: Unrelated to this change, but we don't allow `BearerToken`s to have any other safety value than `do-not-log`. Perhaps we should add the annotation here?

carterkozak: We don't need to declare log-safety for any type which itself has a type annotation describing log-safety (BearerToken has a `@DoNotLog` annotation declared at the class level)

mpritham: oh, didn't see that. Makes sense.

conjure-java-core/src/integrationInput/java/com/palantir/product/StrictOneCollectionField.java
mpritham: Fair. I think that tracks closer with "strict" meaning that every field is set.

Following the same reasoning, it sounds reasonable to add auxiliary methods, so long as the fields are set to something initially.

mpritham: Ok, after some more thought, it's probably best to keep these auxiliary methods. I think the optional value setter should be in the stage where the option is set though.

Will put up the diff to show exactly what I mean.

conjure-java-core/src/integrationInput/java/com/palantir/product/StrictFourFields.java
```diff
@@ -0,0 +1,318 @@
...
+
+    public interface BearerTokenValueStageBuilder {
+        OptionalItemStageBuilder bearerTokenValue(@Nonnull BearerToken bearerTokenValue);
+    }
+
+    public interface OptionalItemStageBuilder {
+        MappedRidsStageBuilder optionalItem(@Nonnull Optional<String> optionalItem);
+
+        MappedRidsStageBuilder optionalItem(@Nonnull String optionalItem);
+    }
```
mpritham, line 157: An option field's value setter is in the field builder's interface instead of in the final stage.

conjure-java-core/src/integrationInput/java/com/palantir/product/StrictFourFields.java
```diff
@@ -0,0 +1,318 @@
...
+    public interface Completed_StageBuilder {
+        StrictFourFields build();
+
+        Completed_StageBuilder addAllMyList(@Nonnull Iterable<String> myList);
+
+        Completed_StageBuilder myList(String myList);
+
+        Completed_StageBuilder putAllMappedRids(@Nonnull Map<ResourceIdentifier, String> mappedRids);
+
+        Completed_StageBuilder mappedRids(ResourceIdentifier key, String value);
```
mpritham: Adding the auxiliary methods for collections to the final stage.

conjure-java-core/src/integrationInput/java/com/palantir/product/StrictOneCollectionField.java
mpritham: Updated.

mpritham:
Aside: I see the build is failing on develop as well with `Could not find jakarta.ws.rs:jakarta.ws.rs-api`. I assuming this is a known issue, but can take a look if not.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -316,18 +359,58 @@ private static List<TypeSpec> generateStageInterfaces(
...
+                        .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
+                        .returns(Primitives.box(objectClass))
+                        .build())
+                .addMethods(fields.stream()
+                        .flatMap(field -> Streams.concat(
+                                Stream.of(generateMethodForFinalStageInterfaceField(
+                                        field, typeMapper, completedStageClass, safetyEvaluator)),
+                                generateAuxiliaryMethodsForFinalStageInterfaceField(
+                                        field, typeMapper, completedStageClass, safetyEvaluator)
+                                        .stream()))
```
mpritham: I want the [cons](https://en.wikipedia.org/wiki/Cons) operation in Java. The closest I found (without an import) was `<E> Lists.asList(E first, E[] rest)`.

carterkozak:
I think this should do the trick: https://github.com/palantir/conjure-java/pull/1993

Gradle up-to-dateness/caching likely allowed us to get into this state :-/

conjure-java-core/src/integrationInput/java/com/palantir/product/StrictFourFields.java
carterkozak: I think the auxiliary methods on the final stage are a little awkward for a couple reason:
- In most cases (field doesn't happen to be the last one), users will have to come back to a field they've already described to use these methods
- There's no way to use an auxiliary method to specify a few entries individually instead of using the full value setter

I might propose dropping the auxiliary methods from the final stage for now. We could support them with additional complex stages, but imo that would be easier to review in a follow-up if we decide it's helpful
```

mpritham: Yeah, this was my initial thought as well, but added them to keep the definition of "strict" mean that only "all fields are set" is enforced. I think dropping the auxiliary methods for collections makes sense here. 

I think the auxiliary value setter for optionals seems ok to keep, in the stage where the optional field is set. What do you think?

carterkozak: Totally agree, the convenience methods are great

mpritham: Updated the PR to remove aux methods for collections.

carterkozak (approved):
great stuff, thanks!

PR #2001, opened 2023-03-23
Nullable annotation for cause in ConjureErrors

deewar:
## Before this PR
the cause parameter in generated code was not annotated with a nullable annotation even though ServiceException accepts a nullable Throwable parameter.

## After this PR
Compile time code checks wont complain when exceptions.getCause() is passed in as a parameter.
==COMMIT_MSG==
Conjure errors have the "Throwable cause" parameter marked as Nullable
==COMMIT_MSG==


carterkozak (approved):
lgtm, thanks!

PR #2018, opened 2023-04-07
add sizes to traced encoding spans

bavardage:
## Before this PR
When debugging slow request deserialization/response serialization, you have to manually cross-reference request logs.

## After this PR
==COMMIT_MSG==
Include sizes of deserialized requests or serialized responses in the traced encoding spans.
==COMMIT_MSG==

## Possible downsides?
The adding the extra key/value to the immutable tags is a bit ugly


carterkozak:
I'd rather do something less invasive like this: #2019

bavardage:
cool that works too

PR #2030, opened 2023-04-25
Add RequestContext method to get request target

pkoenig10:
We use this logic in a couple different places in our internal authentication service. For example, when we receive an unauthenticated request we want to grab the current request target so we can redirect the user back to their original destination after they have authenticated.

It's easy to get this logic wrong. `HttpServerExchange` has a couple of methods that appear similar but have subtle differences. It's particularly important to know whether the returned value has been decoded or not.
- `getRequestURI` returns the raw request target, excluding the query string.
- `getRequestURL` returns the raw request URL, excluding the query string.
    - If the request target is not in absolute form, then it will construct the URL using the schema, host, and port of the request. This can result in URLs with things like IP addresses and is almost never what you want.
- `getRequestPath` return the decoded path component of the request target, excluding the query string
    - If you are storing this for a later redirect, using the decoded value is incorrect.

This PR adds a `requestTarget` method to `RequestContext` to centralize the correct implementation with the goal of making is easier for devs to use the correct value.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
@@ -61,6 +61,13 @@ final class ConjureContexts implements Contexts {
         }
 
         @Override
+        public String requestTarget() {
+            String requestUri = exchange.getRequestURI();
+            String queryString = exchange.getQueryString();
+            return queryString.isEmpty() ? requestUri : requestUri + "?" + queryString;
+        }
+
+        @Override
         public List<String> header(String headerName) {
             HeaderValues header = exchange.getRequestHeaders().get(headerName);
             return header == null ? ImmutableList.of() : Collections.unmodifiableList(header);
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
@@ -16,6 +16,7 @@
 
 package com.palantir.conjure.java.undertow.runtime;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableListMultimap;
 import com.palantir.conjure.java.undertow.lib.Contexts;
@@ -64,7 +65,7 @@ final class ConjureContexts implements Contexts {
         public String requestTarget() {
             String requestUri = exchange.getRequestURI();
             String queryString = exchange.getQueryString();
-            return queryString.isEmpty() ? requestUri : requestUri + "?" + queryString;
+            return Strings.isNullOrEmpty(queryString) ? requestUri : requestUri + "?" + queryString;
         }
 
         @Override
--- a/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
+++ b/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
@@ -30,6 +30,13 @@ import java.util.Optional;
 public interface RequestContext {
 
     /**
+     * Returns the <a href="https://www.rfc-editor.org/rfc/rfc7230#section-5.3">request target</a>.
+     *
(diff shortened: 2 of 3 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
```diff
@@ -29,6 +29,13 @@
  */
 public interface RequestContext {
 
+    /**
+     * Returns the <a href="https://www.rfc-editor.org/rfc/rfc7230#section-5.3">request target</a>.
+     *
+     * This includes the query string and is not decoded in any way.
+     */
+    String requestTarget();
```
pkoenig10, line 39: I decided to call this "request target", matching the terminology from [RFC 7230](https://www.rfc-editor.org/rfc/rfc7230#section-5.3), in an effort to avoid confusion. Terms like "request URI" feel too ambiguous.

The intention is for this to match exactly what is sent on the HTTP request line.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
```diff
@@ -60,6 +60,14 @@ private static final class ConjureServerRequestContext implements RequestContext
             this.requestArgHandler = requestArgHandler;
         }
 
+        @Override
+        public String requestTarget() {
+            String requestUri = exchange.getRequestURI();
+            String queryString = exchange.getQueryString();
+
+            return queryString.isEmpty() ? requestUri : requestUri + "?" + queryString;
```
pkoenig10: If the request target has `?` but no query string, then the returned value will not include the `?`.

For example, given a request target `foo/bar?`, this implementation will return `foo/bar`.

Undertow [doesn't differentiate](https://github.com/undertow-io/undertow/blob/a55aeac10ece93b172f9175b06145d804f4b97c1/core/src/main/java/io/undertow/server/HttpServerExchange.java#L207-L210) between an unset query string and an empty query string, so there's nothing we can do about this.

In practice this should be acceptable.

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
carterkozak: I don't have a strong opinion on this, I follow the rule of three for this sort of thing. It's a reasonable enough feature to support in the request context.

Let's annotate this method with `@Unsafe` since the result will contain query parameter values + path variables.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
```diff
@@ -60,6 +60,13 @@ private static final class ConjureServerRequestContext implements RequestContext
             this.requestArgHandler = requestArgHandler;
         }
 
+        @Override
+        public String requestTarget() {
+            String requestUri = exchange.getRequestURI();
+            String queryString = exchange.getQueryString();
+            return queryString.isEmpty() ? requestUri : requestUri + "?" + queryString;
```
carterkozak: Slight preference for the more defensive `Strings.isNullOrEmpty(` in case of future changes. We have similar (seemingly impossible) checks in our logging layouts where such cases are often triggered by incomplete mockito stubs.

```suggestion`
            return Strings.isNullOrEmpty(queryString) ? requestUri : requestUri + "?" + queryString;
```

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
pkoenig10: By that logic, should we annotate all header and query param methods with `@Unsafe`?

carterkozak: Maybe, but I'd think they're a bit more clear that the safety of the result is based on the key or otherwise dynamic/unsafe part of the request. In this instance, I think the risk is that folks add the result of an invocation because it seems like a helpful thing for debugging without understanding the risks.

The line is definitely fuzzy, though.

carterkozak (approved):
thanks!

PR #2040, opened 2023-05-19
Mark memoized hashcode as volatile

raiju:
## After this PR

==COMMIT_MSG==
Mark memoized hashcode as volatile
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
@@ -113,7 +113,10 @@ public final class MethodSpecs {
 
     public static void addCachedHashCode(TypeSpec.Builder typeBuilder, Collection<FieldSpec> fields) {
         FieldSpec field = FieldSpec.builder(
-                        TypeName.INT, JavaNameSanitizer.sanitize(MEMOIZED_HASH_CODE), Modifier.PRIVATE)
+                        TypeName.INT,
+                        JavaNameSanitizer.sanitize(MEMOIZED_HASH_CODE),
+                        Modifier.PRIVATE,
+                        Modifier.VOLATILE)
                 .build();
         typeBuilder.addField(field);
         typeBuilder.addMethod(createCachedHashCode(fields, field));
--- a/changelog/@unreleased/pr-2040.v2.yml
+++ b/changelog/@unreleased/pr-2040.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Mark memoized hashcode as volatile
+  links:
+  - https://github.com/palantir/conjure-java/pull/2040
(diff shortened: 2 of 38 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
```diff
@@ -113,7 +113,10 @@ public static MethodSpec createHashCode(Collection<FieldSpec> fields) {
 
     public static void addCachedHashCode(TypeSpec.Builder typeBuilder, Collection<FieldSpec> fields) {
         FieldSpec field = FieldSpec.builder(
-                        TypeName.INT, JavaNameSanitizer.sanitize(MEMOIZED_HASH_CODE), Modifier.PRIVATE)
+                        TypeName.INT,
+                        JavaNameSanitizer.sanitize(MEMOIZED_HASH_CODE),
+                        Modifier.PRIVATE,
+                        Modifier.VOLATILE)
```
schlosna, line 119: Chatted with @raiju for a bit about this and while I believe the previous implementation was correct (similar to [JDK `String::hashCode`](https://github.com/openjdk/jdk/blob/42948c04b90d3c01c22d00f684e7dc0129b66abd/src/java.base/share/classes/java/lang/String.java#L2354-L2374), making the field `volatile` removes the benign data race and given [Aleksey Shipilev's advice in "Close Encounters of The Java Memory Model Kind" `Wishful Thinking: Benign Races are Resilient`](https://shipilev.net/blog/2016/close-encounters-of-jmm-kind/#wishful-benign-is-resilient) Conjure objects' `hashCode` tend not to be as contended or critical as `String::hashCode`.

> Exploiting benign races is seldom profitable in usual code. In library code, it sometimes improves performance significantly, especially on non-TSO hardware where volatile reads are not cheap. But it takes utmost discipline to make the race really benign. When using benign races, it is your job to prove why race is actually benign.

schlosna (approved):
Change looks good to me, marked as `do not merge` as I'd like to get @carterkozak 's review as well before merging & releasing

conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java
carterkozak: In the race case, the failure mode is that we calculate the hash code twice. If we make the memoized hash volatile, we still have such a race, but with reduced likelihood in some scenarios.
The lack of a volatile modifier in this implementation was intentional, and it is my preference, I agree that it’s deceptively complex to understand. I don’t have a strong opinion though, do we have an example where we believe there’s an upside in practice to changing?

schlosna: @raiju was running some JCStress tests against a conjure object to validate correctness and the existing non-`volatile` `memoizedHashCode` passed as expected. The `volatile` is more expensive on aarch64 than x86/x64.

carterkozak: I think the critical piece here is that the JMM forbids word tearing.

schlosna: yeah, since these are writes to `int memoizedHashCode`, they're a [single write from JMM perspective, unlike `long` and `double`](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.7)

@raiju should we close this out without merging?

raiju: Sure, we can close it

PR #2042, opened 2023-05-25
RequestContext.peerCertificates returns List

pkoenig10:
There's no need to dictate a particular implementation in the `RequestContext` interface. We don't do this for the other methods here.

This is an ABI break, but I think it's unlikely to affect users because:
- There are very few users of this method
- These uses are almost certainly not in libraries

Searching our internal repos I don't see any uses of this method.


carterkozak (changes requested):
We already use guava types in this API (`ListenableFuture`). I find it helpful to use the type system to describe immutability rather than returning a standard `List`. Ideally the jdk would provide such a type.

PR #2069, opened 2023-07-17
Enable checkUnusedDependencies

pkoenig10:
Also migrates the remaining uses of JUnit4 to Junit5.


conjure-lib/build.gradle
```diff
@@ -22,11 +22,9 @@ dependencies {
     api 'com.palantir.safe-logging:preconditions'
     api 'com.palantir.tokens:auth-tokens'
     api 'com.palantir.conjure.java.api:errors'
-    api 'javax.annotation:javax.annotation-api'
```
pkoenig10, line 25: Eliminating this dependency is what motivated this PR.

build.gradle
```diff
@@ -86,6 +86,8 @@ subprojects {
     tasks.withType(Test) {
         systemProperty 'recreate', System.getProperty('recreate', 'false')
     }
+
+    tasks.check.dependsOn checkUnusedDependenciesMain, checkUnusedDependenciesTest
```
pkoenig10, line 90: If I use `checkUnusedDependencies` here, I get some inexplicable resolution errors. I don't feel like going down the Gradle rabbit hole today so this is a workaround to at least enforce this check on the configurations we care about.

```
* What went wrong:
Could not determine the dependencies of task ':conjure-java-core:compileIntegrationInputJava'.
> Could not resolve all task dependencies for configuration ':conjure-java-core:integrationInputCompileClasspath'.
   > Could not find com.palantir.conjure.java.runtime:conjure-java-retrofit2-client:.
     Required by:
         project :conjure-java-core
```

carterkozak (approved):

PR #2073, opened 2023-07-26
Simplify SafeLong implementation

carterkozak:
We use a simple final class rather than leveraging immutables unnecessarily.

It's possible that an unknown piece of code has subclassed the `abstract` SafeLong type, this is unlikely based on an internal search with no results, and immutables being a clear indicator that arbitrary subtypes are not expected.

Previously the ImmutableSafeLong type was public, there are three known direct references which I am in the process of removing. This introduces some risk, but the type is internal API and not meant to be consumed directly. No uses appear to be in core libraries, so removal should not be risky.

==COMMIT_MSG==
Simplify SafeLong implementation
==COMMIT_MSG==

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
@@ -19,11 +19,9 @@ package com.palantir.conjure.java.lib;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.Preconditions;
-import org.immutables.value.Value;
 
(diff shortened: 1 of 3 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
```diff
@@ -19,46 +19,68 @@
...
+        }
+        SafeLong safeLong = (SafeLong) other;
+        return longValue == safeLong.longValue;
     }
 
     @Override
-    public final int compareTo(SafeLong other) {
-        return Long.compare(longValue(), other.longValue());
+    public int hashCode() {
+        return 177573 + Long.hashCode(longValue);
```
carterkozak, line 84: Magic number borrowed from immutables for a consistent hashCode between the old implementation and this one.

bjlaub (approved):
seems reasonable to me

Issue #96, opened 2018-11-05
Foo.Of method makes all arguments mandatory, including optionals


bshaibu:
## What happened?

Adding an optional field to an object  definition in yml causes the generated object to require the field in its `of` method. This technically makes any optional field addition breaking.

e.g. going from 

```
      DuplicateRequest:
        fields:
          parentFolderId: optional<string>
```

to 

```
      DuplicateRequest:
        fields:
          parentFolderId: optional<string>
          foo: optional<string>
```

generates

```
@JsonDeserialize(builder = DuplicateRequest.Builder.class)
@Generated("com.palantir.conjure.java.types.BeanGenerator")
public final class DuplicateRequest {
 ...
    public static DuplicateRequest of(String parentFolderId, String foo) {
        return builder().parentFolderId(Optional.of(parentFolderId)).foo(Optional.of(foo)).build();
    }
}
```

A few things seem off wrong here:
* `of` shouldn't require optional fields - they should be optional or hidden via overloading
* adding an optional field should preserve the previous `of` methods in case consumers were already using them - removing them

## What did you want to happen?

Some combination of:
* The `of` method should require optionals for optional fields.
* The `of` method should be overloaded for all possible argument combinations given optional arguments
* The `of` method only takes required fields, ignoring optional ones so there's no chance of breaks
* No `of` method in the first place

I'm making the assumption that the method signature change here is a (small) break.

markelliot:
We've hotly debated this a couple of times, I think what I'd propose is:
1. we eliminate `of` method generation from the `-api` projects
2. (low confidence) we consider offering a `-utils` project that includes of methods and other convenience classes

iamdanfox:
To be clear, adding an optional field _is not a wire format break_.  Your issue is describing a compilation break that users of the `.of` fields might encounter.

I like the sound of making the `.of` factory take only required fields though, because if you're willing to add a new required field (i.e. impose a wire-break) then it's fine to impose a compilation break?

Note that this still doesn't give us 100% wire format compatibility though, as users are free to rename conjure types (e.g. `DuplicateRequest` -> `DuplicateItemRequest`) which is wire-format compatible but would obviously result in a compilation break.

pkoenig10:
[...]

PR #2087, opened 2023-08-16
Add flag to exclude static factory methods from generated objects

mpritham:
## Before this PR
Refs #96

## After this PR

==COMMIT_MSG==
Add flag to exclude static factory methods from generated objects
==COMMIT_MSG==

## Possible downsides?
Since the change is behind a flag it wouldn't unintentionally break already generated objects.


conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
```diff
@@ -163,6 +163,15 @@ default boolean jetbrainsContractAnnotations() {
         return false;
     }
 
+    /**
+     * If set to true, static factory methods ('of') will be excluded from generated objects with one or more fields.
+     * Note that for objects without any fields, this will still generate the static factory method.
+     */
+    @Value.Default
+    default boolean excludeStaticFactoryMethods() {
```
mpritham: Preferring this naming instead of `omitGeneratedOfMethods` because I think it's more descriptive.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -126,7 +126,8 @@ public static JavaFile generateBeanType(
                 .addAnnotations(safety)
                 .build());
 
-        if (poetFields.size() <= MAX_NUM_PARAMS_FOR_FACTORY) {
+        if (poetFields.isEmpty()
```
mpritham: Checkstyle is complaining about the cyclomatic complexity here. Will see if there is a nice refactor.

mpritham: Opened #2088.

conjure-java-core/src/test/resources/example-types-no-static-factory.yml
```diff
@@ -0,0 +1,20 @@
+types:
+  imports:
+    ExampleExternalReference:
+      base-type: string
+      external:
+        java: test.api.ExampleExternalReference
+  definitions:
+    default-package: com.palantir.product
+    objects:
+      EmptyExampleNoStaticFactory:
```
mpritham: nit: I think in this case we do have a static factory generated.
```suggestion
      EmptyExample:
```

conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
```diff
@@ -178,6 +178,15 @@ default boolean excludeDialogueAsyncInterfaces() {
...
     }
 
+    /**
+     * If set to true, static factory methods ('of') will be excluded from generated objects with one or more fields.
+     * Note that for objects without any fields, static factory methods will still be generated.
+     */
+    @Value.Default
+    default boolean excludeStaticFactoryMethodsForObjectsWithAtLeastOneField() {
+        return false;
+    }
```
carterkozak: That's quite a mouthful! Perhaps we could go with something like `preferObjectBuilders`?

carterkozak: This object is created based on command line args provided from `ConjureJavaCli` -- we need to mirror this change in that class for it to be used via gradle-conjure!

kunalrkak: Makes sense, updated!

sfackler:
FWIW, we changed the behavior in conjure-rust to generate these constructors off of only the required fields, so they're still there for simple objects but don't break source back compat as easily.

carterkozak (approved):
Thanks!

PR #2096, opened 2023-08-28
Add enum visitor builder

### LINKED ISSUE #1058: Enum Visitors having the same nice lambda builders as Unions
## What happened?
Expected to be able to use the lambda form of the visitor when looking at an enum, but turns out it's not generated.

Have the same lambda form available to us, the same way it is for union types.

mswintermeyer:
## Before this PR
Union types' visitor classes have a nice builder to generate them more concisely, but enum types do not

## After this PR
Addresses https://github.com/palantir/conjure-java/issues/1058

==COMMIT_MSG==
Enum types now have a visitor builder
==COMMIT_MSG==

## Possible downsides?
Implemented this just by copying the relevant methods over from UnionGenerator and adapting them so they fit the enum semantics and types. There's potential for code reuse in future, but seemed like more effort than needed right now since this functionality has been desired for over 3 years (see when the linked issue was created).

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
@@ -20,6 +20,8 @@ import com.fasterxml.jackson.annotation.JsonValue;
 import com.google.common.base.CaseFormat;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.PeekingIterator;
 import com.google.errorprone.annotations.Immutable;
 import com.palantir.conjure.java.ConjureAnnotations;
 import com.palantir.conjure.java.Options;
@@ -28,6 +30,8 @@ import com.palantir.conjure.java.util.Packages;
 import com.palantir.conjure.spec.EnumDefinition;
 import com.palantir.conjure.spec.EnumValueDefinition;
 import com.palantir.logsafe.Safe;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.squareup.javapoet.AnnotationSpec;
 import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.CodeBlock;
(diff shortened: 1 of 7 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/types/EnumTests.java
```diff
@@ -89,4 +91,11 @@ public String visitUnknown(String unknownValue) {
             return unknownValue;
         }
     }
+
+    private static final EnumExample.Visitor<String> VISITOR_FROM_BUILDER = EnumExample.Visitor.<String>builder()
+            .visitOne(() -> "one")
+            .visitTwo(() -> "two")
+            .visitOneHundred(() -> "one hundred")
+            .visitUnknown(unknownType -> unknownType)
```
ash211, line 105: please add test coverage for throwOnUnknown() as well

ash211 (commented):
Thank you for sending!  I've often wanted this myself as a fellow user.

I don't have approval permissions on this repo, but left a comment for a missing test case that I think the approvers would request.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
```diff
@@ -54,33 +63,44 @@ public final class EnumGenerator {
     private static final String VISIT_METHOD_NAME = "visit";
     private static final String VISIT_UNKNOWN_METHOD_NAME = "visitUnknown";
     private static final TypeVariableName TYPE_VARIABLE = TypeVariableName.get("T");
+    public static final String UNKNOWN_TYPE_PARAM_NAME = "unknownType";
+    static final String UNKNOWN_TYPE_NAME = "unknown";
```
mpritham, line 67: Do we want this to also be `private`?

carterkozak (approved):
looks great, thanks @mswintermeyer!

conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
carterkozak: Good catch, `private static final` would be better, I don't think there are uses in other classes. Similarly the `UNKNOWN_TYPE_PARAM_NAME` above should also be private.

carterkozak: (it's fine if this releases as is I'd be happy to approve a fixup later since this module isn't used as a library, so we don't have to worry about folks consuming internal fields)

PR #2100, opened 2023-08-30
completed has underscore in enum visitor builders

mswintermeyer:
## Before this PR
Because enums are UPPER_UNDERSCORE, we use different formatting logic than for unions, so we can remove their underscores. But that meant that we named the completed type CompletedStageVisitorBuilder instead of Completed_StageVisitorBuilder, so there could be errors if an enum actually had a value called COMPLETED.

## After this PR
==COMMIT_MSG==
Enum visitor builders now work when an enum value is named COMPLETED
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
@@ -63,9 +63,11 @@ public final class EnumGenerator {
     private static final String VISIT_METHOD_NAME = "visit";
     private static final String VISIT_UNKNOWN_METHOD_NAME = "visitUnknown";
     private static final TypeVariableName TYPE_VARIABLE = TypeVariableName.get("T");
-    public static final String UNKNOWN_TYPE_PARAM_NAME = "unknownType";
-    static final String UNKNOWN_TYPE_NAME = "unknown";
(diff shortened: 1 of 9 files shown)


mpritham:
Should we add a test that uses `Completed` as a value?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
```diff
@@ -577,17 +579,24 @@ private static String getVisitorMethodName(EnumValueDefinition definition) {
...
+        return enclosingClass.nestedClass(formatName(stageName, CaseFormat.UPPER_CAMEL) + STAGE_VISITOR_BUILDER);
     }
 
     private static String visitorFieldName(String value) {
-        return CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.LOWER_CAMEL, value) + "Visitor";
+        return formatName(value, CaseFormat.LOWER_CAMEL) + VISITOR_FIELD_NAME_SUFFIX;
+    }
+
+    private static String formatName(String value, CaseFormat camelCase) {
+        String formatted = CaseFormat.UPPER_UNDERSCORE.to(camelCase, value);
```
mpritham, line 595: hm I wonder what happens when we have `Completed` and `Completed_`

carterkozak (approved):
thanks for fixing!

PR #2108, opened 2023-09-28
Don't add incubating annotation to server interfaces

pkoenig10:
This is only necessary for client interfaces.

Adding it to server interfaces forces service to add a bunch of `@SuppressWarnings("IncubatingMethod")` annotations.

### DIFF
--- a/changelog/@unreleased/pr-2108.v2.yml
+++ b/changelog/@unreleased/pr-2108.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: The `@Incubating` annotation is no longer applied to server interfaces.
+  links:
+  - https://github.com/palantir/conjure-java/pull/2108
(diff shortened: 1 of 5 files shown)


carterkozak:
Personally I prefer to make it obvious when I'm calling into an incubating method. In many cases I intend to do so, but that's not always the case.
It doesn't look like we currently ignore invocations from within other incubating methods, only from within test code -- would it help your case if we ignored occurrences within incubating methods?

pkoenig10:
It would help in most cases. Although it is an anti-pattern, there are instance where server interfaces are used from within the service itself.

But I just don't think it makes sense to add `@Incubating` or `@Beta` annotations on internal interfaces. These annotations are used to signal to library consumers that an endpoint is unstable and may be changed. But that concern doesn't really apply to internal methods and just creates noise for devs.

carterkozak:
I have a hard time justifying making changes that remove potentially helpful context from generated code to facilitate an anti-pattern -- I think it's fair to require such anti-patterns to suppress the warning.

One of our large internal projects uses the incubating tag much more than anyone else, if they agree that incubating tags cause more friction than they're worth, I'll defer to consensus :-)

pkoenig10:
Alright this is a fair point. Our cases are mostly calling our service interfaces from delegating wrappers. I'll try putting up a PR to avoid flagging method invocations from incubating methods.

pkoenig10:
https://github.com/palantir/gradle-baseline/pull/2636

PR #2119, opened 2023-10-20
Handle path param validation regexes for undertow handlers

fawind:
## Before this PR

The Conjure spec technically allows simple regex validations in http path parameters such as `/foo/{bar:.+}` (see https://github.com/palantir/conjure/issues/297, [code](https://github.com/palantir/conjure/blob/a849637dadaa2bf86dd6525cc7ab3ca8fbefd736/conjure-core/src/main/java/com/palantir/conjure/defs/validator/HttpPathValidator.java#L41C1-L42)). However, when used to generate an Undertow endpoint handler, those would result in runtime errors when trying to access the path param.

Currently, an Undertow handler would look something like this:

```java
private static final class MyEndpoint implements HttpHandler, Endpoint {

    @Override
        Map<String, String> pathParams =
                exchange.getAttachment(PathTemplateMatch.ATTACHMENT_KEY).getParameters();
        String param = runtime.plainSerDe().deserializeString(pathParams.get("param"));
        // ...
    }

    @Override
    public String template() {
        return "/{param:.+}";
    }
}
```

Notice how it uses `pathParams.get("param")` while the path template still contains the regex validation. The template gets passed to undertow's [`PathTemplate`](https://github.com/undertow-io/undertow/blob/88553c2ca5727f8c09d653c8bb451afd6efb26f0/core/src/main/java/io/undertow/util/PathTemplate.java#L61), which doesn't support regex validations and instead handles the entire `param:.+` as the parameter name.

At runtime, this will result in an invalid argument exception, due to the requested parameter `param` being null (see test: https://github.com/palantir/conjure-java/commit/4610af8b4f3c7a5fe18decf629f63234a0e07bbf).

## After this PR

This PR adds a naive fix of ignoring path param validation regexes when generating Undertow handlers but I'm not sure if that's the best path forward.

If we don't want to support those validations, we could still allow them in the spec but drop them much earlier when parsing into endpoint definitions? There are a few usages in conjure definitions in the wild, so we can't drop them from the spec without a break.

==COMMIT_MSG==
Handle path param validation regexes for undertow handlers
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -16,6 +16,7 @@
 
 package com.palantir.conjure.java.services;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -53,6 +54,7 @@ import com.palantir.conjure.spec.EndpointDefinition;
 import com.palantir.conjure.spec.EndpointName;
 import com.palantir.conjure.spec.ExternalReference;
 import com.palantir.conjure.spec.HeaderAuthType;
+import com.palantir.conjure.spec.HttpPath;
 import com.palantir.conjure.spec.ListType;
 import com.palantir.conjure.spec.LogSafety;
 import com.palantir.conjure.spec.OptionalType;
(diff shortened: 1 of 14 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -329,7 +336,7 @@ private TypeSpec generateEndpointHandler(
                         .addModifiers(Modifier.PUBLIC)
                         .addAnnotation(Override.class)
                         .returns(String.class)
-                        .addStatement("return $1S", endpointDefinition.getHttpPath())
+                        .addStatement("return $1S", normalizeHttpPathTemplates(endpointDefinition.getHttpPath()))
```
fawind, line 339: This is the actual code change.

carterkozak:
At a high level, I think it would be better to fail fast if an input isn't fully supported rather than producing slightly unexpected behavior at runtime -- I could imagine a scenario where the path regex is used for security, perhaps to prevent certain path traversal attacks, where passing through the raw value may not be great. I think the malformed name including regex will not work, but currently I expect it fails consistently.

I would be in favor of causing the generator to fail eagerly when these patterns are detected. I thought we had officially deprecated the regex-in-pathparm feature, but perhaps that was never messaged out in the spec -- it was an intentional change not to support path param regex. We could write our own RoutingHandler which understands these patterns if  the lack of support is causing problems

fawind:
> At a high level, I think it would be better to fail fast if an input isn't fully supported rather than producing slightly unexpected behavior at runtime -- [...]

Makes sense! There are a few usages, but I think those only use Jersey endpoints or would otherwise have broken endpoints anyway. Can update the PR to fail Undertow generation when encountering these cases later!

PR #2128, opened 2023-11-01
Remove Usage of Retrofit-clients in Tests

bmarcaur:
Moves from `retrofit` to `dialogue` for testing


conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -470,38 +437,22 @@ public void testBinaryServerSideFailureAfterFewBytesSent() {
...
-                                    }
-                                }));
-                    } catch (UncheckedExecutionException e) {
-                        throw e.getCause();
-                    }
-                })
+        assertThatThrownBy(() -> binaryClient.postBinaryThrows(
+                        AuthHeader.valueOf("authHeader"),
+                        1024 * 1024,
+                        BinaryRequestBody.of(new ByteArrayInputStream(data))))
```
carterkozak: I don't think this is equivalent. SequenceInputStream perhaps?

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -286,15 +274,12 @@ public void testContentLengthSet() throws Exception {
...
     public void testBinaryPost() throws Exception {
         byte[] expected = "Hello, World".getBytes(StandardCharsets.UTF_8);
-        ResponseBody response = binaryClient
-                .postBinary(
-                        AuthHeader.valueOf("authHeader"),
-                        RequestBody.create(MediaType.parse("application/octet-stream"), expected))
-                .get();
-        assertThat(response.contentType()).isEqualTo(MediaType.parse("application/octet-stream"));
-        assertThat(response.bytes()).isEqualTo(expected);
+        InputStream response = binaryClient.postBinary(
```
carterkozak: try-with-resources

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -389,75 +374,57 @@ public void testOptionsOnGetIncludesHead() throws IOException {
         assertThat(con.getHeaderField(HttpHeaders.ALLOW)).isEqualTo("OPTIONS, GET, HEAD");
     }
 
-    @Test
-    public void testUnknownContentType() {
-        assertThatThrownBy(() -> {
```
carterkozak, line 394: We should retain this test, but use `HttpURLConnection` instead of retrofit (verifying the response status code, and SerializableError response body)

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -389,75 +374,57 @@ public void testOptionsOnGetIncludesHead() throws IOException {
...
+        assertThatThrownBy(response::readAllBytes).isInstanceOf(IOException.class);
     }
 
     @Test
+    @SuppressWarnings("MustBeClosedChecker")
     public void testBinaryServerSideFailureAfterFewBytesSent() {
         assertThatThrownBy(() -> {
                     try {
-                        Futures.getUnchecked(binaryClient.getBinaryFailure(AuthHeader.valueOf("authHeader"), 1));
+                        binaryClient.getBinaryFailure(AuthHeader.valueOf("authHeader"), 1);
```
carterkozak: with a small modification we can avoid suppressions
```suggestion
                        binaryClient.getBinaryFailure(AuthHeader.valueOf("authHeader"), 1).close();
```

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -389,75 +374,57 @@ public void testOptionsOnGetIncludesHead() throws IOException {
...
-        assertThat(response.get().string()).isEqualTo("Hello World!");
+        assertThat(new String(response.get().readAllBytes(), StandardCharsets.UTF_8))
+                .isEqualTo("Hello World!");
     }
 
     @Test
+    @SuppressWarnings("MustBeClosedChecker")
     public void testBinaryServerSideFailureAfterManyBytesSent() {
-        ResponseBody response = Futures.getUnchecked(binaryClient.getBinaryFailure(
+        InputStream response = binaryClient.getBinaryFailure(
```
carterkozak: try-with-resources

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
bmarcaur: I will restore 👍🏼

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -470,38 +454,27 @@ public void testBinaryServerSideFailureAfterFewBytesSent() {
...
-                                }));
-                    } catch (UncheckedExecutionException e) {
-                        throw e.getCause();
-                    }
-                })
+        assertThatThrownBy(() -> binaryClient.postBinaryThrows(
+                        AuthHeader.valueOf("authHeader"), 1024 * 2, new BinaryRequestBody() {
+                            @Override
+                            public void write(OutputStream sink) throws IOException {
+                                for (int i = 0; i < 1024 * 1; i++) {
```
carterkozak: Any reason we reduced this from 1gb to 1mb? Idea is that this test should send more data than can fit in system buffers -- 1mb isn't sufficient.

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
bmarcaur: I am not sure I understand the purpose of this test. You can send it less or more and it always throws the same thing. We aren't testing if it stops consuming before it sends too much data. Basically I cannot make this thing fail to show that I have restored its behavior with my implementation. I checked out develop, twiddled with the numbers, and I could not make it fail. Should I be testing that its severing the stream before all the copies are written to it?

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
bmarcaur: Debug remnant, it is restored in my next commit.

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -390,74 +375,73 @@ public void testOptionsOnGetIncludesHead() throws IOException {
...
+        assertThat(client.path(AuthHeader.valueOf("bearer"), expected)).isEqualTo(expected);
     }
 
     @Test
     public void testSpaceInPathParam() {
         String expected = "foo bar";
         assertThat(client.path(AuthHeader.valueOf("bearer"), expected)).isEqualTo(expected);
-        assertThat(Futures.getUnchecked(retrofitClient.path(AuthHeader.valueOf("bearer"), expected)))
-                .isEqualTo(expected);
+        assertThat(client.path(AuthHeader.valueOf("bearer"), expected)).isEqualTo(expected);
```
schlosna: curious, what are the duplicate checks here? should one be testing the synchronous service and the other async service?

conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceEteTest.java
```diff
@@ -164,16 +162,15 @@ public void java_url_client_receives_unprocessable_entity_with_null_body() throw
 
     @Test
     public void test_optionalBinary_present() throws IOException {
-        Optional<ResponseBody> response =
-                Futures.getUnchecked(binary.getOptionalBinaryPresent(AuthHeader.valueOf("authHeader")));
+        Optional<InputStream> response = binaryClient.getOptionalBinaryPresent(AuthHeader.valueOf("authHeader"));
         assertThat(response).isPresent();
-        assertThat(response.get().string()).isEqualTo("Hello World!");
+        assertThat(new String(response.get().readAllBytes(), StandardCharsets.UTF_8))
+                .isEqualTo("Hello World!");
```
schlosna: nit: TWR

```suggestion
        try (InputStream is = response.get()) {
            assertThat(new String(is.readAllBytes(), StandardCharsets.UTF_8))
                .isEqualTo("Hello World!");
        }
```

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -390,74 +375,73 @@ public void testOptionsOnGetIncludesHead() throws IOException {
...
 
     @Test
     public void testBinaryOptionalPresentResponse() throws Exception {
-        Optional<ResponseBody> response =
-                Futures.getUnchecked(binaryClient.getOptionalBinaryPresent(AuthHeader.valueOf("authHeader")));
+        Optional<InputStream> response = binaryClient.getOptionalBinaryPresent(AuthHeader.valueOf("authHeader"));
         assertThat(response).isPresent();
-        assertThat(response.get().string()).isEqualTo("Hello World!");
+        assertThat(new String(response.get().readAllBytes(), StandardCharsets.UTF_8))
+                .isEqualTo("Hello World!");
```
schlosna: nit: TWR

```suggestion
        try (InputStream is = response.get()) {
            assertThat(new String(is.readAllBytes(), StandardCharsets.UTF_8))
                .isEqualTo("Hello World!");
        }
```

versions.lock
```diff
@@ -102,26 +102,20 @@ com.google.auto.value:auto-value-annotations:1.7.4 (1 constraints: 640a29b9)
...
+com.palantir.conjure.java.runtime:refresh-utils:7.66.0 (1 constraints: c918989e)
 com.palantir.safe-logging:preconditions-assertj:3.6.0 (1 constraints: 0b050c36)
 com.palantir.tracing:tracing-jersey:6.18.0 (1 constraints: 47199dbb)
-com.palantir.tracing:tracing-okhttp3:6.18.0 (1 constraints: b814e190)
 com.palantir.websecurity:dropwizard-web-security:1.1.0 (1 constraints: 0405f335)
-com.squareup.okhttp3:okhttp:3.14.1 (3 constraints: fe226150)
+com.squareup.okhttp3:okhttp:3.14.1 (1 constraints: 3b05423b)
 com.squareup.okio:okio:1.17.2 (1 constraints: 850cc309)
-com.squareup.retrofit2:converter-jackson:2.4.0 (1 constraints: 051a63f2)
-com.squareup.retrofit2:retrofit:2.4.0 (3 constraints: 8330aa5e)
```
schlosna, line 124: 🚀

versions.lock
```diff
@@ -102,26 +102,20 @@ com.google.auto.value:auto-value-annotations:1.7.4 (1 constraints: 640a29b9)
...
-com.palantir.conjure.java.runtime:okhttp-clients:7.66.0 (1 constraints: 421ae10c)
-com.palantir.conjure.java.runtime:refresh-utils:7.66.0 (2 constraints: 0a33eabc)
+com.palantir.conjure.java.runtime:conjure-java-legacy-clients:7.66.0 (2 constraints: 071ecd43)
+com.palantir.conjure.java.runtime:refresh-utils:7.66.0 (1 constraints: c918989e)
 com.palantir.safe-logging:preconditions-assertj:3.6.0 (1 constraints: 0b050c36)
 com.palantir.tracing:tracing-jersey:6.18.0 (1 constraints: 47199dbb)
-com.palantir.tracing:tracing-okhttp3:6.18.0 (1 constraints: b814e190)
 com.palantir.websecurity:dropwizard-web-security:1.1.0 (1 constraints: 0405f335)
-com.squareup.okhttp3:okhttp:3.14.1 (3 constraints: fe226150)
+com.squareup.okhttp3:okhttp:3.14.1 (1 constraints: 3b05423b)
```
schlosna, line 117: 🚀

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
bmarcaur: Yea that is an oversight

versions.lock
bmarcaur: This is a lie, retrofit is [still used](https://github.com/palantir/conjure-java/blob/271476d6274741543c87f9e84305f3c59ba5be30/conjure-java-core/build.gradle#L111) it just seems that GCV doesn't understand that configuration. But do not despair. I did the research internally and we do not have a single (active) repo that generates the `api-retrofit` module so this will be removed in a follow up.

versions.lock
bmarcaur: This is likely never going to 0, but its ok because it is a `testImplementation` configuration only.

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -468,40 +460,31 @@ public void testBinaryServerSideFailureAfterFewBytesSent() {
     @Test
     @Timeout(20)
     public void testBinaryServerSideFailureAfterFewBytesReceived() {
-        byte[] data = new byte[1024];
+        int chunkSize = 1024;
+        int expectedChunks = 1024; // 1 MB
+        int chunksToSend = 1024 * 64; // 64 MB
```
bmarcaur: Still needs attention. This is systemically slower than the previous implementation and if I were to match this with 1GB instead of 64MB it would blow the 20 second timeout budget. I was not able to figure out why on my first try.

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -466,42 +458,33 @@ public void testBinaryServerSideFailureAfterFewBytesSent() {
     }
 
     @Test
-    @Timeout(20)
+    @Timeout(40)
```
bmarcaur, line 461: I am intentionally cheating here, and I plan to fix this in a FLUP. Lots of back and forth with @carterkozak we do not think this is a bug nor a regression, but actually an interaction between two testing suites. Which I will fix in a FLUP.

carterkozak (approved):

PR #2133, opened 2023-11-15
Deconflict Jersey Undertow Tests

bmarcaur:
## Before this PR
We observed in #2128 that a specific Undertow test was an order of magnitude slower when run in a suite with the Jersey variants of the test. We suspect their is some conflict between the two. It cannot be the port, but its possible that its some odd lock contention underneath?

## After this PR
Runs them concurrently, but runs them on different ports so they are mutually exclusive.

## Possible downsides?
All I might have done is make them not run serially and made this a heisenbug in that depending on which tests run first it could fail. This also might be fixed but we are on a super ancient version of dropwizard and the comment says dont upgrade it soooo.

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceEteTest.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceEteTest.java
@@ -63,12 +63,11 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.io.TempDir;
 import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
-import org.junit.jupiter.api.parallel.ResourceLock;
 
 @Execution(ExecutionMode.CONCURRENT)
 @ExtendWith(DropwizardExtensionsSupport.class)
-@ResourceLock("port:8080")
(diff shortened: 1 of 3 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceEteTest.java
```diff
@@ -63,12 +63,11 @@
 import org.junit.jupiter.api.io.TempDir;
 import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
-import org.junit.jupiter.api.parallel.ResourceLock;
 
 @Execution(ExecutionMode.CONCURRENT)
 @ExtendWith(DropwizardExtensionsSupport.class)
-@ResourceLock("port:8080")
```
carterkozak, line 70: I'd have kept this since we do need to guarantee nothing else uses port 8080, but in practice there are no other uses of `@ResourceLock` in this project.

carterkozak (approved):

conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyServiceEteTest.java
bmarcaur: Fair enough

PR #2135, opened 2023-11-16
Remove Retrofit Client Generation

bmarcaur:
## Before this PR
Conjure supported generating Retrofit clients

## After this PR
Internally we have completely moved off of their usage 😄


changelog/@unreleased/pr-2135.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
```
carterkozak: might as well call this a break and major rev. version numbers are inexpensive :-)

carterkozak (approved):

PR #2137, opened 2023-11-16
Reauthor Timing Sensitive Test to Use Handlers

bmarcaur:
## Before this PR
`testBinaryResponseClientDisconnect` flakes ever so often, likely because of a timing issue

## After this PR
Now it is implemented using callbacks so that it is no longer timing sensitive

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
@@ -22,7 +22,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.github.stefanbirkner.systemlambda.SystemLambda;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
@@ -30,7 +29,6 @@ import com.google.common.net.HttpHeaders;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.UncheckedExecutionException;
-import com.google.common.util.concurrent.Uninterruptibles;
 import com.palantir.conjure.defs.Conjure;
 import com.palantir.conjure.java.api.errors.RemoteException;
 import com.palantir.conjure.java.api.errors.SerializableError;
@@ -72,7 +70,6 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.time.Duration;
 import java.time.Instant;
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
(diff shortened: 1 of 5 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -486,7 +488,7 @@ public void testBinaryResponseClientDisconnect() throws Exception {
                     .close();
             // Unfortunately there's no great way to tell when the server has finished processing
             // our request.
-            Uninterruptibles.sleepUninterruptibly(Duration.ofMillis(500));
+            Uninterruptibles.sleepUninterruptibly(Duration.ofMillis(100));
```
carterkozak: I think this test was implemented before CojureHandler allowed custom ExceptionHandler implementations to be provided in the builder. Let's move this single test into a new class which sets up the undertow server using a delegating wrapper around the exception handler, that way we can use a countdownLatch after the exception handler has been called instead of an arbitrary sleep or retry.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ClientDisconnectTest.java
```diff
@@ -0,0 +1,67 @@
...
+    private Undertow server;
+    private CountDownLatch latch = new CountDownLatch(1);
+
+    @BeforeEach
+    public void before() {
+        server = Undertow.builder()
+                .addHttpListener(12345, "localhost")
+                .setHandler(new ConjureExceptionHandler(
+                        _exchange -> Uninterruptibles.sleepUninterruptibly(Duration.ofMillis(50)),
+                        (_e, _t) -> latch.countDown()))
```
bmarcaur: This doesnt work, its not actually concurrent so the client isn't disconnecting while the handler is working. WIP

carterkozak: The default exception handler implementation logs the line we're looking for. We still must delegate to that exception handler prior to counting down the latch.

carterkozak: We still need to use the `ConjureHandler` and register a custom wrapping `ExceptionHandler` with that conjure handler ~~in order to use the server-side async codepath~~ (disregard the async piece, I had this confused with another test -- the interaction between the exception handler and the standard handler is still pretty important imo)

bmarcaur: Thanks for the pointers!

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionDisconnectTest.java
```diff
@@ -0,0 +1,78 @@
...
+    private static CountDownLatch latch = new CountDownLatch(1);
+
+    @BeforeAll
+    public static void before() {
+        server = Undertow.builder()
+                .addHttpListener(12345, "localhost")
+                .setHandler(new ConjureExceptionHandler(
+                        _e -> {
+                            Thread.sleep(Duration.ofMillis(50).toMillis());
+                            throw new IOException();
```
bmarcaur, line 46: I tried doing a simple write to the stream so that when it is closed it throws, but then it got mad at me for writing on the IO thread. So then I just figured I would throw, and then it got mad that I never setBlocking this test kinda sucks I need to think about it

bmarcaur: I think I might also be losing sight as to what I am testing, right now it seems that I mostly testing that `ConjureExceptions` write to INFO which this is an elaborate way of doing so. But making a more real setup with conjure and dialogue means I cannot actually get in to the handler chain because `ConjureExceptions` eats the exception and logs.

PR #2143, opened 2023-12-08
No enum visitor builders for large enums

mswintermeyer:
## After this PR

==COMMIT_MSG==
Do not add visitor builders for enums with >100 elements, so we avoid issues with generated strings being too long
==COMMIT_MSG==


conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
```diff
@@ -68,6 +69,7 @@ public final class EnumGenerator {
     private static final String COMPLETED = "completed_";
     private static final String STAGE_VISITOR_BUILDER = "StageVisitorBuilder";
     private static final String VISITOR_FIELD_NAME_SUFFIX = "Visitor";
+    private static final int MAX_VALUES_FOR_STAGED_BUILDER = 100;
```
ash211, line 72: A test like https://github.com/palantir/conjure-java/pull/2144/files#diff-51f01f53d48ae8f797c053fac34a3e9978aa4dcffe00c6bb8cf08b92b5d0e33a may be useful to confirm, in addition to inspection of the generated files

conjure-java-core/src/integrationInput/java/test/prefix/com/palantir/product/LargeEnumExample.java
```diff
@@ -0,0 +1,1109 @@
...
+
+        E99,
+
+        E100,
+
+        UNKNOWN
+    }
+
+    @Generated("com.palantir.conjure.java.types.EnumGenerator")
+    public interface Visitor<T> {
```
carterkozak, line 904: In our internal example, do you know if the visitor is used? I wonder if we should avoid generating the visitor as well, since it'll also be a bit unwieldy.

mswintermeyer: The visitor is not used. However, there's always a chance that some other usage still could use visitors, so I'm hesitant to introduce a back-compat break that'd remove visitors entirely

carterkozak: Happy to start small, we can always chip away more later if that turns out to be helpful.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/EnumGenerator.java
carterkozak: I can go either way on reflective tests you describe. They improve code coverage metrics, but the diff must also make it through code review where we expect the generated code output to receive at least as much scrutiny as the codegen changes.

carterkozak (approved):

PR #2162, opened 2024-01-08
use ExternalLoggingFactory on DropwizardAppExtension

bjlaub:
## Before this PR
Dropwizard's `DefaultLoggingFactory` always resets the root logging level to `DEBUG` via the `reset()` method, which is invoked via `DropwizardTestSupport`. Unfortunately due to https://issues.apache.org/jira/browse/HTTPCLIENT-2313 this can cause other test classes to fail if they run after `JerseyServiceEteTest`. To work around that, just use `ExternalLoggingFactory` which is more representative of real-world environments anyway.

## After this PR

==COMMIT_MSG==
use ExternalLoggingFactory on DropwizardAppExtension
==COMMIT_MSG==

## Possible downsides?
Some tests no longer pass. I'm deleting them since I think they're inconsequential.


carterkozak (dismissed):

carterkozak:
👍

bmarcaur (dismissed):

carterkozak:
Tests are failing with:
```
Caused by: org.gradle.api.GradleException: Found 1 implicit dependencies - consider adding the following explicit dependencies to 'conjure-java-core/build.gradle', or avoid using classes from these jars:
    dependencies {
        testImplementation 'io.dropwizard:dropwizard-logging'
    }
 ```

carterkozak:
👍

felixdesouza:
👍

bmarcaur (approved):

PR #2168, opened 2024-01-16
LazilyInitializedEncoding overrides overloads with Endpoint

carterkozak:
Previously this would drop the Endpoint arg, and funnel calls through the original method.

==COMMIT_MSG==
LazilyInitializedEncoding overrides overloads with Endpoint
==COMMIT_MSG==

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/LazilyInitializedEncoding.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/LazilyInitializedEncoding.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.java.undertow.runtime;
 
 import com.google.common.base.Suppliers;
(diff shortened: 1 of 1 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/LazilyInitializedEncoding.java
```diff
@@ -91,5 +107,10 @@ private static final class LazilyInitializedDeserializer<T> implements Deseriali
         public T deserialize(InputStream input) throws IOException {
             return delegate.get().deserialize(input);
         }
+
+        @Override
+        public String toString() {
+            return "LazilyInitializedDeserializer{" + delegate + '}';
+        }
```
carterkozak, line 114: drive-by: added toString to these as best practice for debugging

mpritham (approved):

PR #2193, opened 2024-02-20
Documenting ordering on conjure-undertow-annotations inferred deserialization

ergo14:
Documenting behavior from https://github.com/palantir/conjure-java/blob/3a425c144ae6293fb6bd8f67d9dea21c8edaadac/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java#L406-L410

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

PR #2194, opened 2024-02-20
Fix inferred deserialization docs

pkoenig10:
Looks like we missed the `fromString` case in https://github.com/palantir/conjure-java/pull/2193.


ergo14 (approved):
my bad, thanks 👍🏼

carterkozak (approved):

PR #2212, opened 2024-02-29
SafeLoggable exception for invalid UUID values

ash211:
## Before this PR
Parsing a too-long UUID string fails with an IllegalArgumentException which does not include the invalid value:

```
java.lang.IllegalArgumentException: UUID string too large
        at java.base/java.util.UUID.fromString1(UUID.java:264)
        at java.base/java.util.UUID.fromString(UUID.java:258)
        at com.palantir.example.BlockInternalId.valueOf(BlockInternalId.java:45)
```

[openjdk/jdk source](https://github.com/openjdk/jdk/blob/998d0baab0fd051c38d9fd6021628eb863b80554/src/java.base/share/classes/java/util/UUID.java#L270)

See https://pl.ntr/2l6 for an internal example stacktrace.

This makes it difficult to debug, as there is no log of the value, even at unsafe level, to assist with debugging.

## After this PR
==COMMIT_MSG==
SafeLoggable exception for invalid UUID values
==COMMIT_MSG==

## Possible downsides?
Potentially there is some performance impact from this change for parsing invalid UUID aliases, or even valid UUIDs.  I think it's likely slight -- just wrapped in a try block for the valid case, and creating a wrapping exception for the invalid case.

### DIFF
--- a/changelog/@unreleased/pr-2212.v2.yml
+++ b/changelog/@unreleased/pr-2212.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: SafeLoggable exception for invalid UUID values
+  links:
+  - https://github.com/palantir/conjure-java/pull/2212
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/UuidAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/UuidAliasExample.java
@@ -3,6 +3,8 @@ package com.palantir.product;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.UnsafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.UUID;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -38,7 +40,11 @@ public final class UuidAliasExample {
     }
 
     public static UuidAliasExample valueOf(String value) {
-        return of(UUID.fromString(value));
+        try {
+            return of(UUID.fromString(value));
+        } catch (IllegalArgumentException e) {
+            throw new SafeIllegalArgumentException("Unable to parse as UUID", e, UnsafeArg.of("input", value));
+        }
     }
 
     @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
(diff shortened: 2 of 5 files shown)


carterkozak:
@ash211 I'm not sure this should be necessary/worth the additional generated exception handling.

In the worst case, the stack trace points to the point in the jdk uuid parser which throws the error, making it clear what the cause was. Not the best devx, but workable.
However, unless a service is misconfigured (missing the constant-strings javaagent), the exception message constant will be considered safe by our infrastructure, so it's not clear to me that making this change would buy us anything.

carterkozak:
Ah, I misread the problem you're solving -- not that the exception isn't safeloggable, but that we don't have a way to understand the bad input.
I think we can also leverage the conjure log-safety context to log bad inputs as `SafeArg` instead of `UnsafeArg`. However, we must also be careful not to include anything tagged `do-not-log` as `UnsafeArg`.

ash211:
Sorry, I should've made the PR title/description more clear.  You're right that I'm trying to solve "what was the invalid input?" and not "the thrown exception doesn't inherit from SafeLoggable"

Being able to promote UnsafeArgs to SafeArgs where possible would be a nice extension!  In the particular case that prompted this the user had access to unsafe logs (on an IL), but I acknowledge that's not the common case.

In the meantime, I was thinking it's safe to log at UnsafeArg level, since that's what `ResourceIdentifier.of(String)` [does](https://github.com/palantir/resource-identifier/blob/d769f8578ada3d1b88395ab4c125f57694971663/src/main/java/com/palantir/ri/ResourceIdentifier.java#L329), and it can receive any arbitrary String inputs.

Maybe we can use the conjure log-safety context (not entirely sure what that is) to improve both `UuidAliasExample.valueOf` and `ResourceIdentifier.valueOf`.  Do you have any pointers on how to do this?  Happy to FLUP with that if you can give a bit of guidance.

carterkozak:
@ash211 The computed/combined log-safety is evaluated here: https://github.com/palantir/conjure-java/blob/2ad5a433bf2355461253a0888ce7eccc27ec7cd3/conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java#L76C29-L76C43

It's used to set the type-level safety annotations on the class.

> In the meantime, I was thinking it's safe to log at UnsafeArg level, since that's what ResourceIdentifier.of(String) [does](https://github.com/palantir/resource-identifier/blob/d769f8578ada3d1b88395ab4c125f57694971663/src/main/java/com/palantir/ri/ResourceIdentifier.java#L329), and it can receive any arbitrary String inputs.

I think that may be a bit questionable, though probably fine in practice because resource-id has a more specific intent than uuid.

If we go down this path, I think we should attempt to solve both Alias.valueOf(String) (for all aliases which produce such a method, not exclusively UUID), and the PlainSerDe deserializers for header/query/path params. I'm not sure we can do anything quite as helpful for jackson deserialization, but it's a start.

PR #2213, opened 2024-03-01
Java generator supports ignoring external type imports

carterkozak:
## Before this PR
Generating remote APIs for projects which use the deprecated external type import feature ranges from tricky to impossible.
See https://github.com/palantir/gradle-conjure/issues/1370

## After this PR
==COMMIT_MSG==
Java generator supports ignoring external type imports using the `externalFallbackTypes` generator option
==COMMIT_MSG==

### DIFF
--- a/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
+++ b/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
@@ -223,6 +223,12 @@ public final class ConjureJavaCli implements Runnable {
                 description = "Union visitors expose the values of unknowns in addition to their types.")
         private boolean unionsWithUnknownValues;
 
+        @CommandLine.Option(
+                names = "--externalFallbackTypes",
+                defaultValue = "false",
+                description = "Java external type imports are generated using their fallback type.")
+        private boolean externalFallbackTypes;
+
         @SuppressWarnings("unused")
         @CommandLine.Unmatched
         private List<String> unmatchedOptions;
@@ -253,7 +259,7 @@ public final class ConjureJavaCli implements Runnable {
                 if (config.generateDialogue()) {
                     generatorBuilder.add(new DialogueServiceGenerator(config.options()));
                 }
-                new GenerationCoordinator(executor, generatorBuilder.build())
+                new GenerationCoordinator(executor, generatorBuilder.build(), config.options())
                         .emit(conjureDefinition, config.outputDirectory());
             } catch (IOException e) {
                 throw new SafeRuntimeException("Error parsing definition", e);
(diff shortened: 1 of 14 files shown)


conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -223,6 +223,12 @@ public static final class GenerateCommand implements Runnable {
                 description = "Union visitors expose the values of unknowns in addition to their types.")
         private boolean unionsWithUnknownValues;
 
+        @CommandLine.Option(
+                names = "--externalFallbackTypes",
+                defaultValue = "false",
+                description = "Java external type imports are generated using their fallback type.")
+        private boolean externalFallbackTypes;
```
ash211, line 230: This option is kind of like type erasure to the base/fallback type, and is all or nothing right now.

In the future we might want something in-between, for example to erase `com.palantir.CustomThing` to a conjure `string`, while keeping `java.time.ZoneId` as an external import because that class is more widely available.

No action now, but wanted to mention this in case it changes the way the options are arranged now.

ash211 (commented):
Thanks for working on this!  I'm excited to try this out once it's available for internal use cases

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
carterkozak: I did consider something special with java standard-library types, but opted for more of an absolutist view to align with the code we produce for other languages like typescript, golang, and python.

carterkozak:
I've cut you a release candidate from this branch, [8.9.0-rc1](https://github.com/palantir/conjure-java/pull/2213/commits/c3cf022541784e3a1a1234ef776fa13bf7c65efb) should be available in ~20 minutes or so.

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
ash211: That makes sense to me.  If there are special cases for Java->Java, then it makes everything non-Java second class in a way.  We aim to make all supported languages equally supported, so pruning benefits of same-language conjure calls is a way to do that.

ash211:
Thank you!  Initial take is this is looking good.  Conjure generation+compile passes now, and I'm working through adapting the rest of the code to the new packages.

Changes I made to pick this up:

`versions.props`
```
com.palantir.conjure.java:* = 8.9.0-rc1
```

`remote-api/build.gradle`
```gradle
conjure {
    java {
        addFlag 'objects'
        addFlag 'dialogue'
        useStagedBuilders = true
        externalFallbackTypes = true  // <-- new
    }
}
```

bjlaub (approved):
lgtm, and makes sense, thanks for the clean code!

carterkozak:
🤖

PR #2222, opened 2024-03-12
Add deprecated forRemoval attribute to server-side code

esword:
## After this PR

==COMMIT_MSG==
Add deprecated forRemoval attribute to server-side code
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
@@ -187,10 +187,15 @@ public final class JerseyServiceGenerator implements Generator {
                     .build());
         }
 
-        endpointDef
-                .getDeprecated()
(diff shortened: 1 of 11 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/JerseyServiceGenerator.java
```diff
@@ -187,10 +187,15 @@ private MethodSpec generateServiceMethod(
                     .build());
         }
 
-        endpointDef
-                .getDeprecated()
-                .ifPresent(
-                        deprecatedDocsValue -> methodBuilder.addAnnotation(ClassName.get("java.lang", "Deprecated")));
+        endpointDef.getDeprecated().ifPresent(_deprecatedValue -> {
```
esword, line 190: this code is now duplicated in 3 places - undertow, jersey, and dialogue.  Happy to add it to a util class if we want.

PR #2225, opened 2024-03-13
Fix externalFallbackTypes to also apply to endpoint return types

ash211:
FLUP to https://github.com/palantir/conjure-java/pull/2213

## Before this PR
Enabling `externalFallbackTypes` would erase external import types to their base type in most places, but not in the return type of endpoints.

## After this PR
==COMMIT_MSG==
Fix externalFallbackTypes to also apply to endpoint return types
==COMMIT_MSG==

Changes in this PR:
1. Change ExternalImportFilter to also erase types for EndpointDefinition returns
2. Enable UndertowServiceGenerator for the test.  For fakes, I am using conjure-java-local not just to generate clients, but also to generate the server-side interface.
3. Change `ExternalLong` in the test conjure file to point to an unknown reference. The classes in src/integrationInput are compiled as part of `check`, so this will help ensure these continue to compile in the future in the face of missing local classes.

## Possible downsides?
None known

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/external/ServiceUsingExternalTypesBlocking.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/external/ServiceUsingExternalTypesBlocking.java
@@ -14,7 +14,6 @@ import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Serializer;
 import com.palantir.dialogue.TypeMarker;
 import com.palantir.logsafe.Safe;
-import java.lang.Long;
 import java.lang.Override;
 import java.lang.String;
 import java.util.List;
@@ -26,7 +25,7 @@ import javax.annotation.processing.Generated;
 public interface ServiceUsingExternalTypesBlocking {
     /** @apiNote {@code PUT /external/{path}} */
     @ClientEndpoint(method = "PUT", path = "/external/{path}")
-    Map<Long, Long> external(@Safe String path, @Safe List<String> body);
+    Map<String, String> external(@Safe String path, @Safe List<String> body);
 
     /** Creates a synchronous/blocking client for a ServiceUsingExternalTypes service. */
     static ServiceUsingExternalTypesBlocking of(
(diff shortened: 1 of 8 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/ExternalImportFilter.java
```diff
@@ -127,6 +127,7 @@ private EndpointDefinition filter(EndpointDefinition definition) {
                 // Remove markers, which are required to be external type imports
                 .markers(ImmutableList.of())
                 .args(Lists.transform(definition.getArgs(), this::filter))
+                .returns(definition.getReturns().map(this::filter).map(TypeWithSafety::type))
```
ash211, line 130: core of the change

conjure-java-core/src/test/java/com/palantir/conjure/java/types/ExternalTypeFallbackTests.java
```diff
@@ -56,7 +57,10 @@ public void testExternalFallbackTypes() throws IOException {
         ConjureDefinition def = Conjure.parse(ImmutableList.of(new File("src/test/resources/external-types.yml")));
         List<Path> files = new GenerationCoordinator(
                         MoreExecutors.directExecutor(),
-                        ImmutableSet.of(new ObjectGenerator(options), new DialogueServiceGenerator(options)),
+                        ImmutableSet.of(
+                                new ObjectGenerator(options),
+                                new DialogueServiceGenerator(options),
+                                new UndertowServiceGenerator(options)),
```
ash211, line 63: I also use this for fakes, so want test coverage on it as well

conjure-java-core/src/integrationInput/java/com/palantir/product/external/ServiceUsingExternalTypesBlocking.java
```diff
@@ -14,7 +14,6 @@
 import com.palantir.dialogue.Serializer;
 import com.palantir.dialogue.TypeMarker;
 import com.palantir.logsafe.Safe;
-import java.lang.Long;
```
ash211, line 17: the fix's effect (removing this import)

carterkozak (approved):
Thanks

PR #2259, opened 2024-03-21
When `nonNullCollections` is enabled, null-check values added to collections

mpritham:
## Before this PR
When the `nonNullCollections` flag is enabled, the generated Conjure definitions do not check that null values cannot be added to collections.

## After this PR
==COMMIT_MSG==
When `nonNullCollections` is enabled, null-check values added to collections
==COMMIT_MSG==

## Possible downsides?
N/A

## Alternative considered
In the generated object, we could hold a collection as a reference to a builder for a relevant implementation of Guava's [ImmutableCollection](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableCollection.html) interface. The `ImmutableCollection` interface null-checks values, and we would not need to extend `ConjureCollections`. This approach is more complicated due to the divergence in implementation from objects generated with and without the `nonNullCollections` flag.


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -819,13 +838,20 @@ private MethodSpec createItemSetter(
...
+    private MethodSpec createMapSetter(EnrichedField enriched, boolean override, boolean nonNullCollections) {
+        MethodSpec.Builder builder = BeanBuilderAuxiliarySettersUtils.createMapSetterBuilder(
+                        enriched, typeMapper, builderClass)
                 .addAnnotations(ConjureAnnotations.override(override))
-                .addCode(verifyNotBuilt())
-                .addStatement("this.$1N.put(key, value)", enriched.poetSpec().name)
-                .addStatement("return this")
-                .build();
+                .addCode(verifyNotBuilt());
+        if (nonNullCollections) {
```
mpritham: oops

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -45,6 +45,20 @@ public static <T> void addAll(Collection<T> addTo, Iterable<? extends T> element
         }
     }
 
+    @SuppressWarnings("unchecked")
+    public static <T> void addAllNonNull(Collection<T> addTo, Iterable<? extends T> elementsToAdd) {
+        Preconditions.checkNotNull(elementsToAdd, "elementsToAdd cannot be null");
+        if (elementsToAdd instanceof Collection) {
+            // This special-casing allows us to take advantage of the more performant
+            // ArrayList#addAll method which does a single System.arraycopy.
+            addTo.addAll((Collection<T>) elementsToAdd);
```
carterkozak: Doesn't this allow null elements to be added?

mpritham: Yes. And so does the `Map#putAll`

mpritham:
Should be covered by other work to support primitive arrays.

PR #2264, opened 2024-03-22
Expose a TypeMarker factory to wrap an arbitrary Type

carterkozak:
## Before this PR
It was impossible to make a reusable serializer/deserializer factory for a generic wrapper type, a new de/serializer would need to be implemented for each component, passing the concrete `TypeMarker`. While likely safer, it's not ideal to repeat unnecessary cruft like this.

## After this PR
==COMMIT_MSG==
Expose a TypeMarker factory to wrap an arbitrary Type
==COMMIT_MSG==

Downsides: This method doesn't provide compile-time safety, as the returned `TypeMarker` cannot have generic type info (result is `TypeMarker<?>` because `Type` itself doesn't provide type info). However, this is reasonable because it can simplify code used by the annotation processor, and doesn't represent anything more dangerous than cast suppressions folks already use.

### DIFF
--- a/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/TypeMarker.java
+++ b/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/TypeMarker.java
@@ -21,6 +21,7 @@ import com.palantir.logsafe.SafeArg;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
+import java.util.Objects;
 
 /**
  * Captures generic type information.
@@ -49,6 +50,10 @@ public abstract class TypeMarker<T> {
                 SafeArg.of("typeVariable", type));
     }
 
+    private TypeMarker(Type type) {
+        this.type = Preconditions.checkNotNull(type, "Type is required");
+    }
+
     public final Type getType() {
         return type;
     }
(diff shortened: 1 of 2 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/TypeMarker.java
```diff
@@ -57,4 +62,29 @@ public final Type getType() {
...
+        }
+
+        @Override
+        public boolean equals(Object other) {
+            if (other instanceof WrappingTypeMarker) {
+                WrappingTypeMarker otherMarker = (WrappingTypeMarker) other;
+                return Objects.equals(getType(), otherMarker.getType());
+            }
+            return false;
+        }
```
pkoenig10, line 88: Should these be moved to `TypeMarker`? I don't see any reason why we wouldn't want this implementation for all `TypeMarker` instances.

I guess maybe it's a little weird to consider different anonymous types equal, but that feel like the correct behavior for `TypeMarker`.

carterkozak: I went back and forth on this. We definitely want to allow caches to work correctly for these, where in the usual anonymous class creation path we’d get the same instance and didn’t need to worry about it.
Equals and hashcode are pretty special in that they should take implementation details into account, so I don’t think it would be correct to make them final in the base class.

I think in a follow-up id like to update our caches to create a copy of TypeMarker instances used as keys using TypeMarker.of(marker.getType()) to avoid holding onto encapsulating classes

pkoenig10 (approved):

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/TypeMarker.java
carterkozak: https://github.com/palantir/conjure-java/pull/2265

PR #2267, opened 2024-03-22
Support providing runtime to param decoders

pkoenig10:
Param decoders often want to leverage the methods provided by `PlainSerDe` in order to get consistent handling of invalid inputs but are forced to instantiate their own local copy of `UndertowRuntime`. While this works today, it could be a bit dangerous in the future if the `PlainSerDe` becomes configurable and the behavior of the server's `UndertowRuntime` does not match the behavior of this local `UndertowRuntime`.

```java
class FooParamDecoder implements CollectionParamDecoder<Foo> {

    private static final PlainSerDe PLAIN_SER_DE =
            ConjureUndertowRuntime.builder().build().plainSerDe();

    @Override
    public Foo decode(Collection<String> values) {
        return PLAIN_SER_DE.deserializeComplex(values, FooParamDecoder::decode);
    }

    private decode(String value) { ... }
}
```

### DIFF
--- a/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/Instantiables.java
+++ b/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/Instantiables.java
@@ -18,8 +18,6 @@ package com.palantir.conjure.java.undertow.processor.data;
 
 import com.google.auto.common.MoreTypes;
 import com.google.common.collect.MoreCollectors;
-import com.palantir.logsafe.SafeArg;
-import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.TypeName;
 import javax.lang.model.element.Element;
@@ -32,18 +30,14 @@ final class Instantiables {
 
     private Instantiables() {}
 
+    static CodeBlock instantiate(TypeMirror mirror) {
+        return instantiate(MoreTypes.asDeclared(mirror));
+    }
+
     /**
      * Produces a {@link CodeBlock} of instantiation code for the provided type.
      * This supports both no-arg class constructors and single-value enum singletons.
      */
-    static CodeBlock instantiate(TypeMirror mirror) {
-        DeclaredType declaredType = MoreTypes.asDeclared(mirror);
-        if (declaredType == null) {
-            throw new SafeIllegalStateException("TypeMirror is not a DeclaredType", SafeArg.of("type", mirror));
-        }
-        return instantiate(declaredType);
-    }
-
     static CodeBlock instantiate(DeclaredType declaredType) {
         TypeElement typeElement = (TypeElement) declaredType.asElement();
         if (typeElement.getKind() == ElementKind.ENUM) {
--- a/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
+++ b/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.java.undertow.processor.data;
 
 import com.google.auto.common.MoreElements;
+import com.google.auto.common.MoreTypes;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -25,7 +26,9 @@ import com.palantir.conjure.java.lib.SafeLong;
 import com.palantir.conjure.java.undertow.annotations.DefaultParamDecoder;
 import com.palantir.conjure.java.undertow.annotations.Handle;
 import com.palantir.conjure.java.undertow.annotations.ParamDecoders;
+import com.palantir.conjure.java.undertow.lib.Endpoint;
 import com.palantir.conjure.java.undertow.lib.RequestContext;
+import com.palantir.conjure.java.undertow.lib.UndertowRuntime;
 import com.palantir.conjure.java.undertow.processor.data.ParameterType.SafeLoggingAnnotation;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.Safe;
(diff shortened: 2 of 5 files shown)


conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
```diff
@@ -349,6 +354,32 @@ private CodeBlock getCollectionParamDecoder(
...
+                        if (parameterTypeName.equals(ClassName.get(UndertowRuntime.class))) {
+                            args.add(CodeBlock.of("runtime"));
+                        } else if (parameterTypeName.equals(ClassName.get(Endpoint.class))) {
+                            args.add(CodeBlock.of("this"));
+                        } else {
+                            return null;
+                        }
+                    }
+                    return CodeBlock.of("new $T($L)", TypeName.get(declaredType), CodeBlock.join(args, ", "));
+                })
```
pkoenig10, line 379: We could require all constructors to have both `UndertowRuntime` and `Endpoint` parameters in a particular order, but:
- This would require users to write constructors with unused parameters, which is a bit strange.
- If we ever support additional parameters in the future, we'd have to do something like this anyways.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/Instantiables.java
```diff
@@ -32,18 +30,14 @@ final class Instantiables {
...
+
     /**
      * Produces a {@link CodeBlock} of instantiation code for the provided type.
      * This supports both no-arg class constructors and single-value enum singletons.
      */
-    static CodeBlock instantiate(TypeMirror mirror) {
-        DeclaredType declaredType = MoreTypes.asDeclared(mirror);
-        if (declaredType == null) {
-            throw new SafeIllegalStateException("TypeMirror is not a DeclaredType", SafeArg.of("type", mirror));
-        }
```
pkoenig10, line 43: `MoreTypes.asDeclared` throws an `IllegalArgumentException` if the given `TypeMirror` is not a `DeclaredType`. It will never return `null`.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
```diff
@@ -349,6 +354,32 @@ private CodeBlock getCollectionParamDecoder(
...
+                .filter(Objects::nonNull)
+                .findFirst();
```
pkoenig10: I opted to just select the first matching constructor that we find, rather than try to prioritize them in some way. This is consistent with the behavior of similar methods here and almost certainly will be acceptable in real-world use.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
carterkozak: 👍 it keeps things simple for folks implementing these

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
```diff
@@ -349,6 +354,32 @@ private CodeBlock getCollectionParamDecoder(
                 });
     }
 
+    private Optional<CodeBlock> getParamDecoderConstructor(DeclaredType declaredType) {
+        TypeElement typeElement = (TypeElement) declaredType.asElement();
+        // new T(String)
+        return typeElement.getEnclosedElements().stream()
+                .filter(element -> element.getKind() == ElementKind.CONSTRUCTOR)
```
carterkozak, line 361: possibly worth excluding private elements here

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
```diff
@@ -349,6 +354,32 @@ private CodeBlock getCollectionParamDecoder(
                 });
     }
 
+    private Optional<CodeBlock> getParamDecoderConstructor(DeclaredType declaredType) {
+        TypeElement typeElement = (TypeElement) declaredType.asElement();
+        // new T(String)
```
carterkozak: Was this intentional?

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
carterkozak: We could produce the (first?) constructor with `max(num-parameters)` of the available options for more obvious behavior -- or possibly better still, require exactly zero or one matches, failing compilation otherwise (so it's obvious that the ctor choice is ambiguous)

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
pkoenig10: Nope

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
pkoenig10: Ah, yes.

The checks in `getUnknownDecoderFactoryFunction` require public constructors/methods. I originally had that here and then the tests failed because I didn't use public constructors in my test cases. I figured that was unnecessarily restrictive and removed it.

I think this in consistency is desirable. There's probably a meaningful difference between a factory method used to instantiate a value and a factory methods used to instantiate a decoder. We probably want to require public elements for the former but it's okay to allow package private elements for the later.

In a follow up, it might be good to add a similar check to `Instantiables`. Right now it throws if there are multiple enum values and generated invalid code if there is not a non-private no-args constructor. At the very least we should add a check, although it would be better to refactor this a bit and report errors.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java
pkoenig10: I've also replaced `.findFirst()` with `.collect(MoreCollectors.toOptional()` in the methods used in `getUnknownDecoderFactoryFunction`. Those filters are specific enough where we should never have more than one element, but always better to enforce the expectation rather than assume it.

carterkozak (approved):
Looks great, thanks!

PR #2274, opened 2024-04-01
Implement optimization for Primitive lists

Shibi-bala:
## Before this PR

Conjure holds primitive arrays as boxed lists. Instead, we can switch to primitive array lists that are backed by primitive typed arrays (double[]) rather than lists (ArrayList<Double>).

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
```diff
@@ -231,7 +232,14 @@ public Builder primitiveItems(int primitiveItems) {
         @JsonSetter(value = "doubleItems", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
         public Builder doubleItems(@Nonnull Iterable<Double> doubleItems) {
             checkNotBuilt();
-            this.doubleItems = ConjureCollections.newArrayList(
+            this.doubleItems = ConjureCollections.newDoubleArrayList(
+                    Preconditions.checkNotNull(doubleItems, "doubleItems cannot be null"));
+            return this;
+        }
+
+        public Builder doubleItems(@Nonnull double[] doubleItems) {
```
Shibi-bala: Adding a new builder methods to allow the setter to create directly from a primitiveArray

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,144 @@
...
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        elements[size] = 0;
+        size--;
+
+        return oldValue;
+    }
+
+    @Override
+    public Iterator<Double> iterator() {
+        // TODO: Implement as the default one doesn't do deletion
```
Shibi-bala: Doesn't the AbstractList Itr class have a remove method?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -71,4 +72,22 @@ public static <T> LinkedHashSet<T> newLinkedHashSet(Iterable<? extends T> iterab
         }
         return set;
     }
+
+    // explicitly need to return mutable list for generated builders
+    public static DoubleArrayList newDoubleArrayList(Iterable<Double> iterable) {
```
Shibi-bala, line 76: Decided to put all the new primitive array list methods here to minimize complexity in the codegen classes

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
raiju: It does; got myself confused.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,144 @@
...
+            resizeIfNecessary(1);
+            // Move over all entries
+            System.arraycopy(elements, index, elements, index + 1, size - index - 1);
+        }
+
+        elements[index] = toAdd;
+    }
+
+    @Override
+    public boolean remove(Object obj) {
```
raiju: We shouldn't need this one (`AbstractCollection` already has an implementation)

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,144 @@
...
+    private int size = 0;
+    private double[] elements;
+
+    public DoubleArrayList() {
+        this(16);
+    }
+
+    private void resizeIfNecessary(int toAdd) {
+        if (size + toAdd > elements.length) {
+            elements = Arrays.copyOf(elements, Math.max(size + toAdd, elements.length * 2));
```
raiju: Probably want to match `ArrayList`'s approach (see `ensureCapacity` & `grow`)

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,144 @@
...
+
+    public DoubleArrayList(Collection<Double> collection) {
+        elements = new double[collection.size()];
+        Double[] array = collection.toArray(Double[]::new);
+        size = collection.size();
+        if (size != 0) {
+            for (int idx = 0; idx < size; idx++) {
+                elements[idx] = array[idx];
+            }
+        }
```
raiju: What is the toArray for? Why not just iterate through the collection?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    public DoubleArrayList(double[] doubleArray) {
+        elements = new double[doubleArray.length];
+        size = doubleArray.length;
+        if (size != 0) {
+            elements = Arrays.copyOf(doubleArray, size);
+        }
```
raiju: This allocates twice. The first allocation is thrown away. 
```suggestion
        size = doubleArray.length;
        elements = Arrays.copyOf(doubleArray, size);
```

build.gradle
```diff
@@ -45,7 +45,7 @@ apply plugin: 'com.palantir.failure-reports'
 
 javaVersions {
     libraryTarget = 11
-    runtime = 21
+    runtime = 17
```
raiju: Surprised to find this in this PR?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -917,4 +979,22 @@ public Boolean visitUnknown(String unknownType) {
             }
         });
     }
+
+    private static final class CollectionType {
+        private final Class<?> clazz;
+        private final boolean isPrimitive;
```
raiju: `isPrimitive` isn't a great name. We're trying to describe the fact that the list class isn't a generic list but a specialized one.

build.gradle
Shibi-bala: This was temp bc I was running into errors with java 21 downloading the jdk. Turns out vpn was blocking some cert during installation to gradle-jdks

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public Iterator<Double> iterator() {
+        // TODO: Implement as the default one doesn't do deletion
+        return super.iterator();
+    }
+
+    @Override
+    public void clear() {
+        for (int i = 0; i < size; i++) {
+            elements[i] = 0;
```
Shibi-bala: We shouldn't need to set elements to 0 when clearing. This should be done with `size=0`

conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
carterkozak: Let's do one thing at a time -- We should implement this as an optimization without impacting the public api of generated types.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
Shibi-bala: I was between that and `isParameterized`. But then started using the same flag for creating an auxiliary setter. So, I renamed it to `isPrimitive`. Maybe there should be two flags. Anyway @carterkozak mentioned leaving this aux. setter optimization for this PR so I'll rename to `isParameterized`

Shibi-bala: [4b0daa1](https://github.com/palantir/conjure-java/pull/2274/commits/4b0daa13de1cf4ff0d2524d957d6712291a07af5)

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -577,7 +600,12 @@ private EnrichedField createField(FieldName fieldName, FieldDefinition field) {
                 ConjureAnnotations.withSafety(typeMapper.getClassName(type), safetyEvaluator.getUsageTimeSafety(field));
         FieldSpec.Builder spec = FieldSpec.builder(typeName, JavaNameSanitizer.sanitize(fieldName), Modifier.PRIVATE);
         if (type.accept(TypeVisitor.IS_LIST) || type.accept(TypeVisitor.IS_SET) || type.accept(TypeVisitor.IS_MAP)) {
-            spec.initializer("new $T<>()", type.accept(COLLECTION_CONCRETE_TYPE));
+            CollectionType collectionType = type.accept(COLLECTION_CONCRETE_TYPE);
+            if (collectionType.isParameterized()) {
+                spec.initializer("new $T()", collectionType.getClazz());
```
Shibi-bala: Technically, if null collections are allowed, we should fallback to ArrayList @carterkozak. Any idea on the number of clients using this option?

Shibi-bala: This handles that 4da29cb77ca56cb4c2ba9594516e1fbdd9454957

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,167 @@
...
+    @SuppressWarnings("ThrowError")
+    private int newCapacity(int minCapacity) {
+        int oldCapacity = elements.length;
+        // increase by 50%
+        int newCapacity = oldCapacity + (oldCapacity >> 1);
+        if (newCapacity - minCapacity <= 0) {
+            if (elements.length == 0) {
+                return Math.max(10, minCapacity);
+            }
+            if (minCapacity < 0) {
```
raiju, line 49: What does this mean?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private void resizeIfNecessary(int toAdd) {
+        if (size + toAdd > elements.length) {
+            elements = Arrays.copyOf(elements, newCapacity(size + toAdd));
+        }
+    }
+
+    @SuppressWarnings("ThrowError")
+    private int newCapacity(int minCapacity) {
```
raiju: Can we guard against overflowing `minCapacity`? (We could enforce this within `resizeIfNecessary`; although at that point I'd suggest inlining the code so we don't get accidental assumptions.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,167 @@
...
+    private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
+    private int size = 0;
+    private double[] elements;
+
+    public DoubleArrayList() {
+        this(16);
+    }
+
+    private void resizeIfNecessary(int toAdd) {
+        if (size + toAdd > elements.length) {
```
raiju: This breaks when `size + toAdd` > 2^31

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,167 @@
...
+    @SuppressWarnings("ThrowError")
+    private static int hugeCapacity(int minCapacity) {
+        if (minCapacity < 0) {
+            throw new OutOfMemoryError();
+        }
+        return (minCapacity > MAX_ARRAY_SIZE) ? Integer.MAX_VALUE : MAX_ARRAY_SIZE;
+    }
+
+    public DoubleArrayList(int initialCapacity) {
+        elements = new double[initialCapacity];
```
raiju: The Java `ArrayList` library uses a neat trick: they allocate a single static empty array so empty lists end up costing an object ptr & an integer.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    public void add(int index, Double toAdd) {
+        checkIndexInAddRange(index);
+        resizeIfNecessary(1);
+        if (index == size) {
+            resizeIfNecessary(1);
+            size++;
+        } else {
+            resizeIfNecessary(1);
```
raiju:
```suggestion
            size++;
        } else {
```

Accidental remains of a refactor?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,167 @@
...
+        return addAll(size, collection);
+    }
+
+    @Override
+    public Double remove(int index) {
+        Objects.checkIndex(index, size);
+        double oldValue = elements[index];
+        if (index != size - 1) {
+            System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        }
```
raiju: Do we want to zero out `elements[size-1]` here?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,167 @@
...
+        size--;
+
+        return oldValue;
+    }
+
+    @Override
+    public void clear() {
+        size = 0;
```
raiju: Same question, do we want to zero out all data?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,167 @@
...
+            offset++;
+        }
+        size += numToAdd;
+        return true;
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends Double> collection) {
+        return addAll(size, collection);
+    }
```
raiju: AbstractList already provides this

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
Shibi-bala: No, size should enforce this. Logic shouldn't be value dependent otherwise 0 arrays would break things.

raiju: Not from a correctness perspective. Call it applied paranoia 😄

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
```diff
@@ -0,0 +1,167 @@
...
+
+package com.palantir.conjure.java.lib;
+
+import java.util.AbstractList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.RandomAccess;
+
+public final class DoubleArrayList extends AbstractList<Double> implements RandomAccess {
```
Shibi-bala: @carterkozak would you like this to be internal or made public? Currently the implementation is pretty trimmed down to mostly what is accessible through the generated code.

If ur team is cool owning maintenance and thinks this will be useful in other places, I'm happy to own flushing out the implementation to optimize all the other methods.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
Shibi-bala: Cool! [86ef3c1](https://github.com/palantir/conjure-java/pull/2274/commits/86ef3c1173395569d70efa2dfabcaba99aa77a73)

Shibi-bala: Cool! 86ef3c11

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
Shibi-bala:
```
public boolean addAll(int index, Collection<? extends E> c) {
    rangeCheckForAdd(index);
    boolean modified = false;
    for (E e : c) {
        add(index++, e);
        modified = true;
    }
    return modified;
}
```

This is the implementation I see from AbstractCollection which AbstractList uses. Better of using the DoubleArrayList implementation that only does one array resize and check.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
carterkozak: > would you like this to be internal or made public?

This should be package private and only accessible as a `List<Double>` result from the public factory methods in `ConjureCollections`.

>  Currently the implementation is pretty trimmed down to mostly what is accessible through the generated code.

Do you have documentation describing what it does and doesn't support?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
Shibi-bala:
```
if (minCapacity < 0) {
    throw new OutOfMemoryError();
}
```

Should check for overflows. This method is only scoped to add/addAll uses so the only way it overflows is if the array tries to grow over INT_MAX. In that case we throw OutOfMemoryError()

Shibi-bala: This mimics arraylist behavior

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
Shibi-bala: Good catch, moved the check to Out of memory check to resize if necessary as well.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
Shibi-bala: yes 🤦

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
Shibi-bala: This is how the class is accessed from generated code. So minimum required functionality is `add`, `addAll`, `constructor`, and `get`. On top of that remove, clear, and set are implemented, but won't ever really be used. AbstractList gives us `iterator` and `indexOf`

```
@JsonSetter(value = "doubleItems", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
public Builder doubleItems(@Nonnull Iterable<Double> doubleItems) {
    checkNotBuilt();
    this.doubleItems = ConjureCollections.newDoubleArrayList(
            Preconditions.checkNotNull(doubleItems, "doubleItems cannot be null"));
    return this;
}

public Builder addAllDoubleItems(@Nonnull Iterable<Double> doubleItems) {
    checkNotBuilt();
    ConjureCollections.addAll(
            this.doubleItems, Preconditions.checkNotNull(doubleItems, "doubleItems cannot be null"));
    return this;
}

public Builder doubleItems(double doubleItems) {
    checkNotBuilt();
    this.doubleItems.add(doubleItems);
    return this;
}
```

Shibi-bala: The codegen'd getter returns List<Double>
```
@JsonProperty("doubleItems")
public List<Double> getDoubleItems() {
    return this.doubleItems;
}
```

conjure-lib/src/main/java/com/palantir/conjure/java/lib/DoubleArrayList.java
Shibi-bala: For posterity we discussed if data could leak on serialization. But the object mapper iterates through objects in a collection respecting the `.size()` method. So this won't be a problem

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/DoubleArrayList.java
```diff
@@ -0,0 +1,183 @@
...
+    private double[] elements;
+
+    public DoubleArrayList() {
+        this(16);
+    }
+
+    private void resizeIfNecessary(int toAdd) {
+        int minCapacity = size + toAdd;
+        if (minCapacity < 0) {
+            throw new OutOfMemoryError();
```
Shibi-bala: These OutOfMemory errors were pulled from the ArrayList implementation. Kept it like this for consistency but also not opposed to using a safe runtime exception or something.

carterkozak:
In terms of sequencing, I think we should split this up into a few parts:

1) new generic list factory methods ConjureCollections which do not allow null values + update codegen to use them. note that `list<any>` should perhaps be allowed to contain null, as `List.of(Optional.empty())` serializes to `[null]` which deserializes as an array containing null.
2) new specialized list factory methods for primitives in ConjureCollections.java, however these methods return the same types as in [1], but without generic inputs. Codegen can be updated to use these where possible
3) We can replace the types returned by the primitive-specialized methods with more efficient collection implementations without impacting behavior because we already have validation against nulls.

this sequencing is helpful because it enforces consistent non-nullness for all conjure collections, rather than based on specific types being used.

Shibi-bala:
> new generic list factory methods ConjureCollections which do not allow null values + update codegen to use them

@carterkozak Just to clarify this, should these new non-null ConjureCollections methods only be used if options.nonNullCollections is true. Otherwise, this is going to cause a break right?

carterkozak:
That's correct

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/DoubleArrayList.java
```diff
@@ -0,0 +1,184 @@
...
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import java.util.AbstractList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.RandomAccess;
+
+public final class DoubleArrayList extends AbstractList<Double> implements RandomAccess {
```
ash211, line 28: were using the GNU Trove4J classes considered here?  https://trove4j.sourceforge.net/javadocs/gnu/trove/list/array/TDoubleArrayList.html  Might want to mention in this class's comment why we aren't using those.

sfackler: Hasn't trove been abandonware for the last decade?

ash211: Ah yes indeed.  Not a decade, but close.  https://bitbucket.org/trove4j/trove/commits/d4433e656b9eba5b1ed713ffe8e15660af37dfb4

Maybe https://eclipse.dev/collections/ then?  I meant this more of a "why build our own?" question than "why not X?"

carterkozak: fwiw I think eclipse-collections may be reasonable here as an `implementation` dependency of conjure-java, so long as the eclipse-collection implementation types are never exposed directly. This way we're free to change the underlying implementation as we wish. I believe `new BoxedMutableDoubleList(new DoubleArrayList())` provides the behavior we're looking for.

Shibi-bala: Considered using a OSS implementation of this but some of my first choices were Trove/FastUtils but aren't maintained much. Highbury coincidently ran into a classloader deadlock with eclipse-collections right at the time of writing this 😓. Anyway, due to how small the api needs to be for Conjure access paths, I figured writing it up myself was good.

Conjure gen code only requires an add, addAll, and constructor fwiw. All pretty straightforward to implement, at least up till Java 21.

PR #2275, opened 2024-04-04
Avoid generating union builders for large unions

carterkozak:
## Before this PR
all unions are generated with a visitor builder.

## After this PR
This is equivalent to #2143, but for unions rather than exclusively enums.
==COMMIT_MSG==
Avoid generating union builders for large unions. Particularly large enums result in too much bytecode, and in some cases cause us to run against Java limits
==COMMIT_MSG==

## Possible downsides?
Possible that folks are using builders for existing large unions, however this isn't terribly likely, and most failures should be caught at compile time in the unexpected case this is hit.


mpritham (approved):

ash211:
> Possible that folks are using builders for existing large unions, however this isn't terribly likely, and most failures should be caught at compile time in the unexpected case this is hit.

🙋 that's me!

Internal search for `r:/eddie f:ValidationErrorHelper.java` shows the location, and failing PR is `foundry/eddie/pull/17087`.  It's a conjure union of ~175 values.

What's the recommended fix?  Should we switch from the visitor builder to the far more verbose visitor subclass?

PR #2279, opened 2024-04-08
Implement non null conjure collection [Primitive Arrays Pt. 1]

Shibi-bala:
## Before this PR

This adds null value checking when using the nonNullCollections flag. In a subsequent PR, I'll move primitive lists to use a different type-specific collection factory method

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -673,17 +658,29 @@ private MethodSpec createCollectionSetter(String prefix, EnrichedField enriched,
...
-                        "this.$1N = $2T.new$3T($4L)",
-                        spec.name,
-                        ConjureCollections.class,
-                        type.accept(COLLECTION_CONCRETE_TYPE),
-                        Expressions.requireNonNull(
-                                spec.name, enriched.fieldName().get() + " cannot be null"));
+                CollectionType collectionType = getCollectionType(type);
+                if (collectionType.useNonNullFactory()) {
+                    return CodeBlocks.statement(
+                            "this.$1N = $2T.newNonNull$3T($4L)",
```
Shibi-bala: prefixing with `newNonNull` and adding the null checks to ConjureCollections seems like the simplest way to do this apart from creating a NonNullArrayList class.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -917,4 +914,68 @@ public Boolean visitUnknown(String unknownType) {
...
+
+    private CollectionType getCollectionType(Type type) {
+        return type.accept(new DefaultTypeVisitor<>() {
+            @Override
+            public CollectionType visitList(ListType _value) {
+                return _value.getItemType().accept(new DefaultTypeVisitor<>() {
+                    @Override
+                    public CollectionType visitDefault() {
+                        return new CollectionType(
+                                ArrayList.class, options.nonNullCollections() && !isOptionalInnerType(type));
```
Shibi-bala: referencing https://github.com/palantir/conjure-java/pull/2274#issuecomment-2040347409, seems any type with optional in it will need a collection that can hold nulls.

conjure-java-core/src/integrationInput/java/com/palantir/product/ManyFieldExample.java
```diff
@@ -298,7 +298,8 @@ public Builder optionalItem(@Nonnull String optionalItem) {
         @JsonSetter(value = "items", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
         public Builder items(@Nonnull Iterable<String> items) {
             checkNotBuilt();
-            this.items = ConjureCollections.newArrayList(Preconditions.checkNotNull(items, "items cannot be null"));
+            this.items =
+                    ConjureCollections.newNonNullArrayList(Preconditions.checkNotNull(items, "items cannot be null"));
```
carterkozak: Let's add a unit test which asserts a NPE is thrown when we execute the following:

```java
CovariantListExample.builder().items(Collections.singletonList(null));
CovariantListExample.builder().addAllItems(Collections.singletonList(null));
CovariantListExample.builder().items((String) null);
```

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -59,6 +59,24 @@ public static <T> ArrayList<T> newArrayList(Iterable<? extends T> iterable) {
...
+                Preconditions.checkNotNull(item, "iterable cannot contain null elements");
+            }
+            return new ArrayList<>((Collection<T>) iterable);
+        }
+        ArrayList<T> list = new ArrayList<>();
+        for (T item : iterable) {
+            Preconditions.checkNotNull(item, "iterable cannot contain null elements");
+            list.add(item);
+        }
+        return list;
```
carterkozak: Thoughts on structuring this as a delegation to the existing `newArrayList`, then checking the result for nulls?

The current implementation involves a race where the input may be mutated to include null values between iteration over the `Collection` case, and `new ArrayList(Collection)`

carterkozak: We should include methods for Set as well

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
carterkozak: I don't think that's true, a `list<optional<string>>` should never include null, otherwise we'd have some `optional.empty` elements and some `null` elements, which wouldn't make a lot of sense!

carterkozak: (but we should add tests to keep us honest :-)

Shibi-bala: This is current codegen setter for list\<any\> and nonNullCollections
```
public Builder items(@Nonnull Iterable<?> items) {
    checkNotBuilt();
    this.items = ConjureCollections.newArrayList(Preconditions.checkNotNull(items, "items cannot be null"));
    return this;
}
```
From the annotations, jackson won't let us set deser nulls. So I think it is fine if this setter also uses `newNonNullArrayList`

carterkozak: 👍

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
Shibi-bala: 👍

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -45,6 +45,23 @@ public static <T> void addAll(Collection<T> addTo, Iterable<? extends T> element
         }
     }
 
+    @SuppressWarnings("unchecked")
+    public static <T> void addAllAndCheckNonNull(Collection<T> addTo, Iterable<? extends T> elementsToAdd) {
```
Shibi-bala, line 51: There are currently two cases for the underlying collection for `addTo`:

For ArrayList, we can use ensureCapcaity to eagerly grow the backing array to avoid multiple copies. Thus, the add(e) call that is done in a for loop only results in array assignments.

For HashSet, the existing addAll implementation already just takes the incoming Iterable and repeatedly calls add(e). So this case stays the same.

For the future primitive array classes, it'll be a good idea to follow a similar api

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -917,4 +929,46 @@ public Boolean visitUnknown(String unknownType) {
             }
         });
     }
+
+    private CollectionType getCollectionType(Type type) {
+        return type.accept(new DefaultTypeVisitor<>() {
+            @Override
+            public CollectionType visitList(ListType _value) {
+                return new CollectionType(ArrayList.class, options.nonNullCollections());
```
Shibi-bala: I think it is interesting to think about when we can enforce non-null on the backing conjure collections (constructor, addAll, add).

Adding nonNullCollections means adding the `contentNulls = Nulls.FAIL` block to the json deser annotation. This means jackson will not ever deserialize anything with a null in it. This includes collections that had nulls, or were of type collection<any> and had Optional.empty() in them (something strongly typed to collection<optional<T>> will serde properly as seen in tests).

This means the deserialized object will Never have nulls in it if the nonNullCollections is turned on. I think one conjure guarantee is that anything that serializes, must deserialize. Having nulls in a collection with this flag would break that rule, so I think it makes sense to enforce nonNull-ness on all collections when this is on.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -71,4 +94,57 @@ public static <T> LinkedHashSet<T> newLinkedHashSet(Iterable<? extends T> iterab
         }
         return set;
     }
+
+    public static <T> List<T> newList() {
+        return new ArrayList<>();
+    }
+
+    public static <T> List<T> newList(Iterable<? extends T> iterable) {
```
Shibi-bala, line 100: Deprecated old methods in favor of more generic factory methods that allow us to change backing list implementation

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -576,8 +560,13 @@ private EnrichedField createField(FieldName fieldName, FieldDefinition field) {
         TypeName typeName =
                 ConjureAnnotations.withSafety(typeMapper.getClassName(type), safetyEvaluator.getUsageTimeSafety(field));
         FieldSpec.Builder spec = FieldSpec.builder(typeName, JavaNameSanitizer.sanitize(fieldName), Modifier.PRIVATE);
-        if (type.accept(TypeVisitor.IS_LIST) || type.accept(TypeVisitor.IS_SET) || type.accept(TypeVisitor.IS_MAP)) {
-            spec.initializer("new $T<>()", type.accept(COLLECTION_CONCRETE_TYPE));
+        if (type.accept(TypeVisitor.IS_LIST) || type.accept(TypeVisitor.IS_SET)) {
+            spec.initializer(
+                    "$1T.new$2L()",
```
Shibi-bala, line 564: Calling factory methods from ConjureCollections means we don't need javapoet type($T) to be added to the import list. This causes edge cases when the collection here is the only usage. I think using an internal collection type may be easier since we can then just make the name correspond to the ConjureCollections factory method.

carterkozak: I don't follow -- can you elaborate on the edge case you're describing?

Shibi-bala: So the old initializer was `$1T.new$2T()` where $2 was passed a class, like ArrayList. The codegen adds any types added with $T to the list of deps to import. When we switched to using conjure collections factories like `ConjureCollections.newList()` this works but breaks when the name of the factory method uses a class that isn't used anywhere else and is referred to with the parent class. Ie newDoubleArrayList -> List<Double>. This would add DoubleArrayList as a import even though it isn't need. So maybe not exactly an edge case, but something that broke with the old logic.

carterkozak: Ah I see, seems like the old logic was a bit incorrect/fragile :-]

conjure-java-core/src/test/java/com/palantir/conjure/java/NonNullCollectionsTest.java
```diff
@@ -0,0 +1,74 @@
...
+public class NonNullCollectionsTest {
+    private static final ObjectMapper objectMapper = ObjectMappers.newClientObjectMapper();
+
+    @Test
+    public void throwsNpe() {
+        // ListExample is using code generated with the nonNullCollections flag set to true. Thus, we should not be able
+        // to add a null to the collection in any way.
+        assertThatExceptionOfType(NullPointerException.class)
+                .isThrownBy(() ->
+                        ListExample.builder().items(Collections.singleton(null)).build());
```
carterkozak: slight preference to extract `Collections.singleton(null)` to above -- that way if an excavator ever attempts to replace it with `Set.of(null)`, the test throws a NPE at that point, so it's not ambiguous where things fail!

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollectionType.java
```diff
@@ -0,0 +1,32 @@
...
+    private final String collectionName;
+
+    ConjureCollectionType(String collectionName) {
+        this.collectionName = collectionName;
+    }
+
+    public String getCollectionName() {
+        return collectionName;
+    }
+}
```
carterkozak: I don't think this type should exist -- conjure-lib is the core API required by our generated code, and this seems like more of a utility for codegen. Codegen should be largely self-contained

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -45,7 +47,26 @@ public static <T> void addAll(Collection<T> addTo, Iterable<? extends T> element
...
+        }
+        for (T element : elementsToAdd) {
+            Preconditions.checkNotNull(element, "elementsToAdd cannot contain null elements");
+            addTo.add(element);
+        }
+    }
+
+    // Prefer to use newList(iterable)
     // explicitly need to return mutable list for generated builders
+    @Deprecated
```
carterkozak: Please do not deprecate these, it can cause folks who upgrade conjure-lib prior to upgrading codegen to get a bunch of compiler warnings, making it harder to read the output.
It takes about ~6 months before we can realistically deprecate

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -71,4 +94,54 @@ public static <T> LinkedHashSet<T> newLinkedHashSet(Iterable<? extends T> iterab
...
+    public static <T> List<T> newList(Iterable<? extends T> iterable) {
+        Preconditions.checkNotNull(iterable, "iterable cannot be null");
+        if (iterable instanceof Collection) {
+            return new ArrayList<>((Collection<T>) iterable);
+        }
+        List<T> list = new ArrayList<>();
+        for (T item : iterable) {
+            list.add(item);
+        }
+        return list;
```
carterkozak: No reason to duplicate code:
```suggestion
        return newArrayList(iterable);
```

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -71,4 +94,54 @@ public static <T> LinkedHashSet<T> newLinkedHashSet(Iterable<? extends T> iterab
...
+    public static <T> Set<T> newSet(Iterable<? extends T> iterable) {
+        Preconditions.checkNotNull(iterable, "iterable cannot be null");
+        if (iterable instanceof Collection) {
+            return new LinkedHashSet<>((Collection<T>) iterable);
+        }
+        Set<T> set = new LinkedHashSet<>();
+        for (T item : iterable) {
+            set.add(item);
+        }
+        return set;
```
carterkozak: We should delegate here as well

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
Shibi-bala: 👍

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollectionType.java
Shibi-bala: that makes sense moved to BeanBuilderGenerator

carterkozak:
Tagging @schlosna, since you've worked in `ConjureCollections.java` as well, in case you'd like to provide feedback here.

This change generally looks good to me. In terms of sequencing, sometimes I like to stage this sort of thing into a feature branch before merging, that way if we get to the specialized collections and find we wish we'd done something a little different, we can still change earlier parts without worrying about them being released/used in prod already. Up to you.

carterkozak (approved):

PR #2282, opened 2024-04-09
Add @CheckReturnValue to build() methods

### LINKED ISSUE #2281: Add @CheckReturnValue onto generated conjure object methods
## What happened?

A product I maintain recently had a bug where we instantiated a conjure object in Java source and then did nothing with it.  ErrorProne has a [@CheckReturnValue](https://errorprone.info/bugpattern/CheckReturnValue) pattern meant to help with this.  Let's apply it and try to squash this category of bug entirely.

Add `@CheckReturnValue` to the generated `MyObject.builder()` and `MyObject.Builder.build()` methods.  When paired with the error prone rule, this will catch any issues where callers call one of those methods but then discard the result.  This will manifest as a build failure when they pick up the new version of conjure-java.

ash211:
Fixes https://github.com/palantir/conjure-java/issues/2281

## Before this PR
It's possible for users to call `MyConjureObject.builder().x(1).y(2).build();` and not assign that to a variable.  This is almost certainly a mistake, and calling code should either remove the instantiation of that object, or assign to a variable / return the value.

## After this PR
==COMMIT_MSG==
Add @CheckReturnValue to build() methods
==COMMIT_MSG==

[@CheckReturnValue](https://errorprone.info/bugpattern/CheckReturnValue)

## Possible downsides?
Any violations of the error prone rule will manifest as a build failure when downstream repos pick up the new version of conjure-java.


ash211:
We had a private discussion about how many methods to put `@CheckReturnValue` on.  You can imagine really going ham on this and sprinkling them over tons of methods.  "constructors" like builder(), build(), from(), of(), valueOf(), and empty().  Also getters like get(), myField(), toString(), hashCode().  But not setters like myField(val) since they can be chained in a fluent style or not.

Carter's comment:
> It's difficult to make a call where these annotations should be applied, and where it doesn't necessarily make sense. Adding them will cause some level of rollout friction on existing but non-critical bugs, and increase the load we put on the java compiler. If we can prevent bugs from shipping, this is reasonable, but it depends on the cost vs rate we expect bugs to appear vs their criticality... all of which are difficult to measure!

We ended up deciding to take a small incremental step and add to just the `build()` method, since that's the one that was involved in the bug that triggered this PR.

carterkozak (approved):

PR #2285, opened 2024-04-11
Implement unparameterized collection factories [Primitive Arrays Pt. 2] 

Shibi-bala:
## Before this PR

Creating factory methods for primitive collections that change codegen while keeping backing ArrayList

Replaced by https://github.com/palantir/conjure-java/pull/2306 to add changelog

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
```diff
@@ -165,9 +167,9 @@ public static final class Builder {
 
         private List<@Safe String> items = new ArrayList<>();
 
-        private List<Integer> primitiveItems = new ArrayList<>();
+        private List<Integer> primitiveItems = new IntegerArrayList();
 
-        private List<Double> doubleItems = new ArrayList<>();
+        private List<Double> doubleItems = new DoubleArrayList();
```
Shibi-bala: This is what the final codegen will look like. We should now be able to change the implementation of DoubleArrayList to use a primitive array.

Shibi-bala: new DoubleArrayList() should be replaced with ConjureCollections factory method

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -71,4 +94,94 @@ public static <T> LinkedHashSet<T> newLinkedHashSet(Iterable<? extends T> iterab
...
+        Set<T> set = newSet(iterable);
+        for (T item : set) {
+            Preconditions.checkNotNull(item, "iterable cannot contain null elements");
+        }
+
+        return set;
+    }
+
+    // This method returns a list that can't handle nulls. Do not use this unless the nonNullCollections flag is set
+    public static List<Double> newDoubleArrayList() {
```
Shibi-bala: Should be able to just replace these with instantiations of the eclipse-collection implementation

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -83,30 +84,15 @@
...
-
     private static final String BUILT_FIELD = "_buildInvoked";
     private static final String CHECK_NOT_BUILT_METHOD = "checkNotBuilt";
     private static final String BUILDER_IMPLEMENTATION_NAME = "Builder";
     /* The name of the interface extending all stage interfaces. */
     private static final String STAGED_BUILDER_INTERFACE_NAME = "Builder";
     /* The name of the class implementing the interface extending all stage interfaces. */
     private static final String STAGED_BUILDER_IMPLEMENTATION_NAME = "DefaultBuilder";
+    private static final boolean USE_NON_NULL_COLLECTION_FACTORY = true;
+    private static final boolean DO_NOT_USE_NON_NULL_COLLECTION_FACTORY = false;
```
carterkozak: Let's define an enum for readability throughout

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -920,4 +939,80 @@ public Boolean visitUnknown(String unknownType) {
             }
         });
     }
+
+    private CollectionType getCollectionType(Type type) {
+        return type.accept(new DefaultTypeVisitor<>() {
+            @Override
+            public CollectionType visitList(ListType value) {
+                if (!options.nonNullCollections()) {
+                    return new CollectionType(ConjureCollectionType.LIST, DO_NOT_USE_NON_NULL_COLLECTION_FACTORY);
```
carterkozak: not-non-null double negative is a bit tricky to read, perhaps `NULLABLE_COLLECTION_FACTORY`?

Shibi-bala: 👍

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
carterkozak: Why do we have both `newDoubleArrayList` and `newNonNullDoubleArrayList` neither of which support nulls?

carterkozak: My preference is to explicitly declare `NonNull` in the method name. We should not declare the `ArrayList` implementation detail, so something like `public static List<Double> newNonNullDoubleList()` is what I'd go for here

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -940,38 +940,101 @@ public Boolean visitUnknown(String unknownType) {
...
+                            }
+                        });
+                    }
+                });
             }
 
             @Override
             public CollectionType visitSet(SetType _value) {
-                return new CollectionType(ConjureCollectionType.SET, options.nonNullCollections());
+                if (options.nonNullCollections()) {
```
carterkozak: Let's add a new temporary "experimental" feature flag to opt into codegen using the new methods, that way we can allow dependencies to roll out for a bit before folks start to use them. After a couple weeks we can remove the flag and make it the default, that way rollout shouldn't break folks who violate jar dependencies 😓

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
Shibi-bala: fixed bf56b6e08901238154e793c981ea82543431d030

also opted to create `newNonNullSet` and `newNonNullList` to make the code a little simpler (these just instantiate a List, but the add, addAll methods will do the validation...) + can replace these later with a validating collection...

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
Shibi-bala: good idea 4ef8ad246583c75959a4ee27f5c2cfca32fdf881

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -126,4 +135,43 @@ public static <T> Set<T> newNonNullSet(Iterable<? extends T> iterable) {
 
         return set;
     }
+    // This method returns a list that can't handle nulls. Do not use this unless the nonNullCollections flag is set
```
carterkozak, line 139: odd missing whitespace above this, I'm not sure why the formatter didn't insert a newline

PR #2288, opened 2024-04-18
Use eclipse collections in implementation [Primitive Arrays Pt.3]

Shibi-bala:
## Before this PR

replaced by https://github.com/palantir/conjure-java/pull/2307

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -0,0 +1,89 @@
...
+package com.palantir.conjure.java.lib.internal;
+
+import com.palantir.logsafe.exceptions.SafeUnsupportedOperationException;
+import java.util.AbstractList;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.RandomAccess;
+import org.eclipse.collections.impl.list.mutable.primitive.DoubleArrayList;
+
+public final class ConjureDoubleList extends AbstractList<Double> implements RandomAccess {
```
Shibi-bala, line 26: This is weird. eclipse-collections doesn't extend AbstractList<> so I need to use this delegating pattern. I guess this takes 8 more bytes... FastUtils does extend this (https://github.com/karussell/fastutil/blob/master/src/it/unimi/dsi/fastutil/doubles/DoubleArrayList.java#L69), but its no longer maintained... Ideally the implementation already extends this class so we can instantiate directly in conjureCollections. This feels really close to writing our own impl.

Curious on your thoughts about if its worth pivoting to us writing the implementation where we throw on all unused code paths (everything except add, addAll) will prevent unintentionally picking up a inefficient default impl.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public Double get(int index) {
+        Objects.checkIndex(index, delegate.size());
+        return delegate.get(index);
+    }
+
+    @Override
+    public boolean addAll(int _index, Collection<? extends Double> _collection) {
+        throw new SafeUnsupportedOperationException("This operation is unsupported");
```
Shibi-bala, line 62: This is also weird. The API for eclipse collections provides `boolean addAllAtIndex(int var1, DoubleIterable var2)`, but LazyDoubleIterable is in the api package which could cause classloader bug in java < 21. There are workarounds, like implementing our own DoubleIterable, but given this method is never accessed, I left the exception here.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureBooleanList.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public boolean addAll(int _index, Collection<? extends Boolean> _collection) {
+        throw new SafeUnsupportedOperationException("This operation is unsupported");
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends Boolean> collection) {
+        for (boolean element : collection) {
```
Shibi-bala, line 68: No ensureCapacity for BooleanArrayList for some reason. Maybe an oversight?

Shibi-bala (commented):
@carterkozak left a few questions. Eclipse-collections is weird.... No tests yet. Want to decide what implementation strategy is first

PR #2291, opened 2024-04-23
Alias types use memoized hashCode for collections

schlosna:
## Before this PR

Generated alias types did not memoize `hashCode`, potentially leading to expensive hash code computations when the aliased type is an immutable collection or map.

A hypothetical example `CompositeKey` alias for `Map<Field, Value>` currently generates:

```
@Generated("com.palantir.conjure.java.types.AliasGenerator")
public final class CompositeKey {
    private static final CompositeKey EMPTY = new CompositeKey();

    private final Map<Field, Value> value;

    private CompositeKey(@Nonnull Map<Field, Value> value) {
        this.value = Preconditions.checkNotNull(value, "value cannot be null");
    }

    private CompositeKey() {
        this(Collections.emptyMap());
    }

    @JsonValue
    public Map<Field, Value> get() {
        return value;
    }

    @Override
    public String toString() {
        return value.toString();
    }

    @Override
    public boolean equals(@Nullable Object other) {
        return this == other
                || (other instanceof CompositeKey && this.value.equals(((CompositeKey) other).value));
    }

    @Override
    public int hashCode() {
        return value.hashCode();
    }

    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
    public static CompositeKey of(@Nonnull Map<Field, Value> value) {
        return new CompositeKey(value);
    }

    public static CompositeKey empty() {
        return EMPTY;
    }
}
```

## After this PR

==COMMIT_MSG==
Alias types use memoized hashCode for collections
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2291.v2.yml
+++ b/changelog/@unreleased/pr-2291.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Alias types use memoized hashCode for collections
+  links:
+  - https://github.com/palantir/conjure-java/pull/2291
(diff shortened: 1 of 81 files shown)


carterkozak (approved):
lgtm, thanks!

PR #2295, opened 2024-04-29
`Double` Alias types with log safety equality fix

felixdesouza:
## Before this PR
After #2291 merged, some tests started failing in a specific case of an aliased `Double` that had safety annotations on.

The `equalTo` method was now generating something like this:
```diff
-Double.doubleToLongBits(this.value) == Double.doubleToLongBits(((Float) other).value))
+this.value == other.value
```

https://github.com/palantir/conjure-java/blob/7891d80363baa9843cc75b39b8a04d1430be1a2e/conjure-java-core/src/main/java/com/palantir/conjure/java/types/MethodSpecs.java#L93-L97

The type we pass in is `ConjureAnnotations.withSafety(<underlying type>)` and if you have any safety annotations, you end up with a `TypeName.Double` *with* annotations on top which are not equal to `TypeName.Double`, hence why we're hitting the second block.

This was the confusing part as `AliasDoubleExample` was actually generating the correct output.

## After this PR

==COMMIT_MSG==
Ensure that Aliased `Double` types with specified log safety have a correct `equalTo`.
==COMMIT_MSG==

* added a new [example type](https://github.com/palantir/conjure-java/commit/5a266de4c717707f653935b8811657bca46672f3) and you can see the incorrect generated code.

* added a convenience method, `Primitives#isDouble` so don't have to remember to remove annotations etc.

### DIFF
--- a/changelog/@unreleased/pr-2295.v2.yml
+++ b/changelog/@unreleased/pr-2295.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Ensure that Aliased `Double` types with specified log safety have a
+    correct `equalTo`.
+  links:
+  - https://github.com/palantir/conjure-java/pull/2295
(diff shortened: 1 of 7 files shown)


schlosna (approved):

carterkozak (approved):
Thanks for fixing this!

PR #2306, opened 2024-05-12
Implement unparameterized collection factories [Primitive Arrays Pt. 2]

Shibi-bala:
## Before this PR

Creating factory methods for primitive collections that change codegen while keeping backing ArrayList

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -940,38 +948,101 @@ public Boolean visitUnknown(String unknownType) {
...
+                            }
+                        });
+                    }
+                });
             }
 
             @Override
             public CollectionType visitSet(SetType _value) {
-                return new CollectionType(ConjureCollectionType.SET, options.nonNullCollections());
+                if (options.nonNullCollections() && options.primitiveOptimizedCollections()) {
```
carterkozak: Does `options.primitiveOptimizedCollections()` have any impact for sets?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -940,38 +948,101 @@ public Boolean visitUnknown(String unknownType) {
     private CollectionType getCollectionType(Type type) {
         return type.accept(new DefaultTypeVisitor<>() {
             @Override
-            public CollectionType visitList(ListType _value) {
-                return new CollectionType(ConjureCollectionType.LIST, options.nonNullCollections());
+            public CollectionType visitList(ListType value) {
+                if (!options.nonNullCollections() || !options.primitiveOptimizedCollections()) {
+                    return new CollectionType(
+                            ConjureCollectionType.LIST, ConjureCollectionNullHandlingMode.NULLABLE_COLLECTION_FACTORY);
+                }
```
carterkozak: This reverts some of the behavior from the previous PR, doesn't it? I think some of the previous invocations of `ConjureCollections.newNonNullList(original)` will be reverted to `ConjureCollections.newList(original)` losing null checks we already have

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
Shibi-bala: Yes, it will use newNonNullSet() instead of newSet(). Maybe not the most correct name in this case, but was going for wrapping all the codegen changes in this one flag.

Shibi-bala: Functionally those two methods just call `new Set<>()`, but can be modified in the future with some validating set if necessary

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
Shibi-bala: ah true 3490ad8f

carterkozak: I think our goal in the short term should be that removing `.primitiveOptimizedCollections(true)` from `ObjectGeneratorTests` produces no diff when comparing `conjure-java-core/src/integrationInput` with develop. It's a bit of a pain to implement because the goal isn't so much absolute corredtness, but the ability to opt into/out of changes to the codegen.
The risk of removing our checks is that some projects have updated tests to adopt new conjure versions, and those tests may fail again in the opposite direction if we effectively revert our null-checking logic. In a week or two we can remove the new feature flag, and get rid of that cruft from this codebase, but I think a smooth rollout is worth it

Shibi-bala: added slightly different codegen in this PR
https://github.com/palantir/conjure-java/pull/2306/files#diff-af480f5a1b91af7a42236746b4697dc2bb4f932e3d3dbafb8e6032d1d6efe0b8R567-R576

Shibi-bala: I can put this change behind the feature flag

Shibi-bala: 583b0e07

carterkozak: I think you pushed to your fork. If you have two remotes, you can remove your fork and rename the other remote to the name you're most familiar with

```bash
git remote remove myforkname
git remote rename palantirremotename myforkname
```
So in my case, it usually takes this form because I usually call my forked-source remote `upstream` and my the remote I push to `origin`:

```bash
git remote remove origin
git remote rename upstream origin
```

Shibi-bala: ah yeah i did, oops

carterkozak (approved):

PR #2307, opened 2024-05-12
Use eclipse collections in implementation [Primitive Arrays Pt.3] 

Shibi-bala:
## Before this PR

Replace generic list implementations with primitive optimized collections using eclipse collections.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -0,0 +1,89 @@
...
+package com.palantir.conjure.java.lib.internal;
+
+import com.palantir.logsafe.exceptions.SafeUnsupportedOperationException;
+import java.util.AbstractList;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.RandomAccess;
+import org.eclipse.collections.impl.list.mutable.primitive.DoubleArrayList;
+
+public final class ConjureDoubleList extends AbstractList<Double> implements RandomAccess {
```
Shibi-bala: This is weird. eclipse-collections doesn't extend AbstractList<> so I need to use this delegating pattern. I guess this takes 8 more bytes... FastUtils does extend this (https://github.com/karussell/fastutil/blob/master/src/it/unimi/dsi/fastutil/doubles/DoubleArrayList.java#L69), but its no longer maintained... Ideally the implementation already extends this class so we can instantiate directly in conjureCollections. This feels really close to writing our own impl.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public Double get(int index) {
+        Objects.checkIndex(index, delegate.size());
+        return delegate.get(index);
+    }
+
+    @Override
+    public boolean addAll(int _index, Collection<? extends Double> _collection) {
+        throw new SafeUnsupportedOperationException("This operation is unsupported");
```
Shibi-bala: This is also weird. The API for eclipse collections provides `boolean addAllAtIndex(int var1, DoubleIterable var2)`, but LazyDoubleIterable is in the api package which could cause classloader bug in java < 21. There are workarounds, like implementing our own DoubleIterable, but given this method is never accessed, I left the exception here.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureBooleanList.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public boolean addAll(int _index, Collection<? extends Boolean> _collection) {
+        throw new SafeUnsupportedOperationException("This operation is unsupported");
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends Boolean> collection) {
+        for (boolean element : collection) {
```
Shibi-bala: No ensureCapacity for BooleanArrayList for some reason. Maybe an oversight?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
carterkozak: Why are we providing our own implementations instead of using eclipse-collections types directly?

Shibi-bala: See this https://github.com/palantir/conjure-java/pull/2307#discussion_r1597703482

carterkozak: I'm not sure I understand how `AbstractList` is relevant, though

Shibi-bala:
```
public static List<Double> newNonNullDoubleList() {
    return new DoubleArrayList();
}
```

won't work because DoubleArrayList doesn't extend AbstractList<Double> or implement List<Double>

carterkozak: I don't think it needs to -- shouldn't we implement the collection factories using something along these lines in order to expose eclipse-collections using standard java collection facades?

```java
public static List<Double> newNonNullDoubleList() {
    return new BoxedMutableDoubleList(new DoubleArrayList());
}
```

Shibi-bala: Whats the benefit of this approach versus using a delegate instantiated in the parent's constructor? As I understand these all will do the same thing?

carterkozak: `return new BoxedMutableDoubleList(new DoubleArrayList());` uses classes that we don't have to build and maintain ourselves

Shibi-bala: Where do you see BoxedMutableDoubleList? I can't find it, so I assumed this would be what ConjureDoubleList.java is right now

Shibi-bala: Seems like that isn't functionality provided by eclipse-collections yet https://github.com/eclipse/eclipse-collections/issues/1408

carterkozak: Ah it is implemented on the tip of the development branch, but has not yet been released in 12

Shibi-bala: oof that makes sense y i was confused. I was looking at their 11.1 release. Last release was in 2022 so idk when this will come out. Keep current pr as is and bump version when its out?

carterkozak: I think that’s reasonable, I think a couple minor tweaks will make things a bit easier:
1. Let’s javadoc the plan to use eclipse-collections 12 wrappers, since we don’t know how far out that will be
2. Let’s make our classes similar in shape/structure to the eclipse-collections api so that the transition won’t require much code change aside from deleting our implementations.
3. Our collection implementations should be package private, and only created (without exposing type information) through the ConjureCollections factories

Shibi-bala: done https://github.com/palantir/conjure-java/pull/2307/commits/e8c2f78e85390ea9435c427d5ae7db2b0aa484ef

eclipse-collections implementation for reference: https://github.com/eclipse/eclipse-collections/blob/master/eclipse-collections-code-generator/src/main/resources/impl/list/mutable/boxedPrimitiveList.stg

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureBooleanList.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public void clear() {
+        throw new SafeUnsupportedOperationException("This operation is unsupported");
+    }
+
+    @Override
+    public Boolean set(int _index, Boolean _element) {
+        throw new SafeUnsupportedOperationException("This operation is unsupported");
+    }
```
carterkozak: Let's implement these, otherwise code which appears reasonable may fail at runtime:
```java
    @Override
    public Boolean remove(int index) {
        return delegate.removeAtIndex(index);
    }

    @Override
    public void clear() {
        delegate.clear();
    }

    @Override
    public Boolean set(int index, Boolean element) {
        return delegate.set(index, element);
    }
```

Shibi-bala: 605102cd

carterkozak (approved):
sorry for the delay reviewing this!

PR #2312, opened 2024-05-21
Add primitive array setter optimization [Primitive Arrays Pt.4]

Shibi-bala:
## Before this PR

Add back optimization removed here: https://github.com/palantir/conjure-java/pull/2274#discussion_r1546622698

This will allow clients to just track primitive arrays internally which will be a nice perf improvement.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
@@ -154,9 +154,7 @@ public final class ConjureCollections {
 
     // This method returns a list that can't handle nulls. Do not use this unless the nonNullCollections flag is set
     public static List<Double> newNonNullDoubleList(double[] doubles) {
-        double[] conjureCopy = new double[doubles.length];
-        System.arraycopy(doubles, 0, conjureCopy, 0, doubles.length);
-        return new ConjureDoubleList(DoubleArrayList.newListWith(conjureCopy));
+        return new ConjureDoubleList(DoubleArrayList.newListWith(doubles.clone()));
     }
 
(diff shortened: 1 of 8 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -140,11 +141,22 @@ public static <T> Set<T> newNonNullSet(Iterable<? extends T> iterable) {
         return set;
     }
 
+    /**
```
Shibi-bala, line 144: Oops don't know how I accidentally left the comment block down there :(

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -217,4 +245,8 @@ public static List<SafeLong> newNonNullSafeLongList(Iterable<SafeLong> iterable)
 
         return safeLongList;
     }
+
+    private static boolean safeLongCheck(long value) {
```
Shibi-bala, line 250: duplicated check from SafeLong. Not sure if its worth moving this to be public

carterkozak:
I’m currently ooto, but as I mentioned in our last discussion, I don’t want to expand the generated types public api. We have gone to great lengths to avoid and remove code using arrays over well-encapsulated immutable types. Before examining this, it would be helpful to have examples with benchmarks of what we’re trying to resolve, as well as an analysis of the other options which might achieve the desired outcome.

raiju:
@carterkozak I modified the methods to copy the array on input, which should protect our expectations around mutability. While it will cost some memory/speed, it's quite a bit more efficient than the boxed list version. Would this be ok to merge?

PR #2317, opened 2024-06-05
Enable primitive collection optimization by default

Shibi-bala:
## Before this PR

primitive collection changes should have rolled out so lets flip the flag. It is still gatekept by the `nonNullCollections` flag

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/types/ObjectGeneratorTests.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/types/ObjectGeneratorTests.java
@@ -25,6 +25,7 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.palantir.conjure.defs.Conjure;
 import com.palantir.conjure.java.GenerationCoordinator;
 import com.palantir.conjure.java.Options;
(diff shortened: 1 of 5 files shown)


carterkozak (commented):
I think we want to remove the flag entirely here

conjure-java-core/src/test/java/com/palantir/conjure/java/types/ObjectGeneratorTests.java
```diff
@@ -56,6 +58,8 @@ public final class ObjectGeneratorTests {
 
     @Test
     public void testObjectGenerator_allExamples() throws IOException {
+        List<SafeLong> longs = new ArrayList<>();
+        assertThat(longs).hasSize(0);
```
carterkozak: was this intentional?

Shibi-bala: oops. I had that to mess around in the debugger. Didn't notice it when I pushed

carterkozak (dismissed):

carterkozak (approved):

PR #2338, opened 2024-08-07
[HackWeek] Generated builders understand default field values from IR

cflems:
## Before this PR
Context: https://github.com/palantir/conjure/pull/1615

## After this PR
Hooks into the new IR definition including defaults in order to generate bean builders that understand these default values. Since these builders are used by Jackson to deserialize wire and storage types, this enables `null` or missing field values to be interpreted for backwards compatibility.

==COMMIT_MSG==
Generated builders understand default field values from IR
==COMMIT_MSG==


cflems:
A couple things. One, this depends on the PR in [conjure](https://github.com/palantir/conjure), so right now it's on a mavenLocal version and the checks are going to fail until the upstream is merged. Two, this presents an opportunity to do something very useful by way of separating wire-format deserializing builders from userland builders. This would allow deserialization to maintain backwards compatibility without presenting the same options to developers, forcing them to write code conforming to updated object formats and eliminating an entire class of bugs that comes along with insufficiently strict builders.

PR #2341, opened 2024-08-15
Add test for Cache-Control override

pkoenig10:
It looks like we intended to write a test for this case, but simply forgot.


carterkozak (approved):
Thanks!

PR #2351, opened 2024-09-05
Singleton beans use a JsonCreator with mode=PROPERTIES

carterkozak:
See the discussion here:
https://github.com/FasterXML/jackson-databind/issues/4688#issuecomment-2330537868

Ideally we can avoid deserialization failures for existing code, but it's better to align our usage with the maintainers vision either way :-)

==COMMIT_MSG==
Singleton beans use a JsonCreator with mode=PROPERTIES
==COMMIT_MSG==


bjlaub:
👍

PR #2352, opened 2024-09-05
conjure-undertow streamed binary response bodies protect against early closure

carterkozak:
There's a common pattern where folks use a try-with-resources block on the response stream. While it's generally good practice to close our resources, the webserver is responsible for this one, and closing it in a finally block can prevent us from relaying errors to clients! By blocking close calls to the delegate, we can allow folks to close all the things, while providing accurate failure information to clients :-]

==COMMIT_MSG==
conjure-undertow streamed binary response bodies protect against early closure
==COMMIT_MSG==

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/UnclosableOutputStream.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/UnclosableOutputStream.java
@@ -67,6 +67,6 @@ final class UnclosableOutputStream extends OutputStream {
 
     @Override
     public String toString() {
-        return "CloseShieldingOutputStream{" + delegate + '}';
+        return "UnclosableOutputStream{" + delegate + '}';
     }
 }
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/UnclosableOutputStream.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/UnclosableOutputStream.java
@@ -1,0 +1,72 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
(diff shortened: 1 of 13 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/UnclosableOutputStream.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private void assertOpen() throws IOException {
+        if (closeCalled) {
+            throw new SafeIoException("Stream is closed");
+        }
+    }
+
+    @Override
+    public String toString() {
+        return "CloseShieldingOutputStream{" + delegate + '}';
```
mpritham: nit: 
```suggestion
        return "UnclosableOutputStream{" + delegate + '}';
```

carterkozak: Naming is hard, apparently so is renaming...
```suggestion
        return "UnclosableOutputStream{" + delegate + '}';
```
(using my own suggestion to avoid confusing policy-bot)

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/UnclosableOutputStream.java
```diff
@@ -0,0 +1,72 @@
...
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.exceptions.SafeIoException;
+import java.io.IOException;
+import java.io.OutputStream;
+
+/** Helper stream used in {@link ConjureBodySerDe} to make error propagation easier. */
+final class UnclosableOutputStream extends OutputStream {
+
+    private final OutputStream delegate;
+    private boolean closeCalled;
```
bjlaub, line 28: any chance this needs to be `volatile`? i don't think so, but just figured i'd double check.

bjlaub:
👍 lgtm

PR #2362, opened 2024-09-12
add support for triggering callback after auth token has been set

bjlaub:
## Before this PR
Runtime will try to parse and inject an auth token, but provides no mechanism for consumers to apply any extra business logic based on the token value before a handler is actually invoked.

## After this PR
add support for triggering callback after auth token has been set

This allows consumers to hook into the AuthorizationExtractor encapsulated by ConjureUndertowRuntime and make early decisions about request processing based on the extracted token.

The expected use case is to be able to e.g. implement rate limiting or other business logic using data within a JWT that can be done without any extra validation, but should happen just before actual request handlers are invoked.

==COMMIT_MSG==
add support for triggering callback after auth token has been set
==COMMIT_MSG==

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
@@ -27,9 +27,15 @@ import com.palantir.conjure.java.undertow.lib.MarkerCallback;
 import com.palantir.conjure.java.undertow.lib.PlainSerDe;
 import com.palantir.conjure.java.undertow.lib.UndertowRuntime;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.tokens.auth.AuthHeader;
+import com.palantir.tokens.auth.BearerToken;
+import com.palantir.tokens.auth.UnverifiedJsonWebToken;
+import io.undertow.server.HttpServerExchange;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
+import java.util.function.BiConsumer;
 
 /** {@link ConjureUndertowRuntime} provides functionality required by generated handlers. */
 public final class ConjureUndertowRuntime implements UndertowRuntime {
@@ -46,7 +52,8 @@ public final class ConjureUndertowRuntime implements UndertowRuntime {
                 builder.encodings.isEmpty()
                         ? ImmutableList.of(Encodings.json(), Encodings.smile(), Encodings.cbor())
                         : builder.encodings);
-        this.auth = new ConjureAuthorizationExtractor(plainSerDe());
+        this.auth = new OnSetRequestTokenAuthExtractor(
+                new ConjureAuthorizationExtractor(plainSerDe()), builder.onSetRequestToken);
         this.exceptionHandler = builder.exceptionHandler;
         this.markerCallback = MarkerCallbacks.fold(builder.paramMarkers);
         this.async = new ConjureAsyncRequestProcessing(builder.asyncTimeout, builder.exceptionHandler);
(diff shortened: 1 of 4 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -132,8 +141,51 @@ public Builder requestArgHandler(RequestArgHandler value) {
             return this;
         }
 
+        @CanIgnoreReturnValue
+        public Builder onSetRequestToken(BiConsumer<HttpServerExchange, Optional<UnverifiedJsonWebToken>> callback) {
```
carterkozak: Let's model this after the `public Builder requestArgHandler(RequestArgHandler value) {` method above and define a custom interface. Perhaps something like:
```java
public Builder jsonWebTokenHandler(JsonWebTokenHandler value) {
```

A custom `JsonWebTokenHandler` interface should be a bit easier for folks to understand at a glance, and we can javadoc it with relevant info.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -132,8 +141,51 @@ public Builder requestArgHandler(RequestArgHandler value) {
...
+            onSetRequestToken = callback;
+            return this;
+        }
+
         public ConjureUndertowRuntime build() {
             return new ConjureUndertowRuntime(this);
         }
     }
+
+    private static final class OnSetRequestTokenAuthExtractor implements AuthorizationExtractor {
```
carterkozak: We can pass our handler into the ConjureAuthorizationExtractor, no need to reimplement AuthorizationExtractor here, this way we can avoid dealing with `Attachments.UNVERIFIED_JWT`

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
```diff
@@ -132,8 +141,51 @@ public Builder requestArgHandler(RequestArgHandler value) {
...
+            Preconditions.checkNotNull(token, "jwt in exchange attachment is null");
+            onSetRequestToken.accept(exchange, token);
+            return result;
+        }
+
+        @Override
+        public void setRequestToken(HttpServerExchange exchange, Optional<UnverifiedJsonWebToken> token) {
+            delegate.setRequestToken(exchange, token);
+            // may throw
+            onSetRequestToken.accept(exchange, token);
```
carterkozak: Let's add test coverage including one where the custom handler throws, to ensure this work

bjlaub: sure thing, added

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
bjlaub: fixed

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureUndertowRuntime.java
bjlaub: fixed

carterkozak (approved):
thanks!

carterkozak:
wat -- changelog isn't enforced in this repo!? I think the automation has messed the config up a bit

PR #2386, opened 2024-10-18
[Feature] Add Serialization Optimization for Primitive Collection Types

bmarcaur:
## Before this PR
When Jackson attempted to serialize a primitive collection type, e.g. `ConjureDoubleList`, it utilized Jackon's default serializers, specifically `IndexedListSerializer`. This makes sense when you remember that `ConjureDoubleList` [does not have an implemented serializer](https://github.com/palantir/conjure-java/blob/17ac7361a1d16261dfa119e3ef307b64d8361b58/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java#L29), so at best its going to be treated like List<Double> at time of serialization. This causes unnecessary boxing no matter how optimally you this collection stores the primitive types. You can see evidence of these points in the debugging screenshot below:

image

## After this PR
Adds serialization paths that avoid boxing and directly leverage Jackson primitive serialization routines.

~I purposely didn't implement this feature for `ConjureBooleanList` because it does not have a simple serialized format that benefits from primitive optimization. i.e. it expands a single bit out to `true` or `false` over the wire. Frankly, this type feels like it was added for completeness and not for usefulness. A dense `boolean[]` is abstractly a bit mask and if that is the intent should have its own type. I think the existence of this class is just maintenance overhead at this point.~ This has been removed https://github.com/palantir/conjure-java/pull/2390

## Possible downsides?
Creates an alternative path for serialization that needs to match the default serialization paths.

### DIFF
--- a/conjure-java-core/src/test/resources/primitive-collections.yml
+++ b/conjure-java-core/src/test/resources/primitive-collections.yml
@@ -1,0 +1,8 @@
+types:
+  definitions:
+    default-package: com.palantir.product
+    objects:
+      PrimitiveExample:
+        fields:
+          ints: list<integer>
+          doubles: list<double>
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
@@ -16,10 +16,11 @@
 
 package com.palantir.conjure.java.lib.internal;
 
+import com.fasterxml.jackson.annotation.JsonValue;
 import java.util.AbstractList;
 import java.util.Collection;
 import java.util.RandomAccess;
-import org.eclipse.collections.impl.list.mutable.primitive.DoubleArrayList;
+import org.eclipse.collections.api.list.primitive.MutableDoubleList;
 import org.eclipse.collections.impl.utility.Iterate;
 
 /**
@@ -27,9 +28,9 @@ import org.eclipse.collections.impl.utility.Iterate;
  * a BoxedMutableDoubleList will be released. Once available, ConjureDoubleList should be replaced with that.
  */
 final class ConjureDoubleList extends AbstractList<Double> implements RandomAccess {
-    private final DoubleArrayList delegate;
+    private final MutableDoubleList delegate;
 
-    ConjureDoubleList(DoubleArrayList delegate) {
+    ConjureDoubleList(MutableDoubleList delegate) {
         this.delegate = delegate;
     }
 
@@ -68,5 +69,16 @@ final class ConjureDoubleList extends AbstractList<Double> implements RandomAcce
     @Override
     public Double set(int index, Double element) {
         return delegate.set(index, element);
+    }
+
+    ConjureDoubleList asUnmodifiable() {
+        return new ConjureDoubleList(delegate.asUnmodifiable());
+    }
+
+    // Cannot be named 'toArray' as that conflicts with the #toArray in AbstractList
+    // This is a serialization optimization that avoids boxing, but does copy
+    @JsonValue
+    double[] jacksonSerialize() {
+        return delegate.toArray();
     }
 }
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureIntegerList.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureIntegerList.java
@@ -16,10 +16,11 @@
 
 package com.palantir.conjure.java.lib.internal;
 
+import com.fasterxml.jackson.annotation.JsonValue;
 import java.util.AbstractList;
 import java.util.Collection;
 import java.util.RandomAccess;
-import org.eclipse.collections.impl.list.mutable.primitive.IntArrayList;
+import org.eclipse.collections.api.list.primitive.MutableIntList;
 import org.eclipse.collections.impl.utility.Iterate;
 
 /**
@@ -27,9 +28,9 @@ import org.eclipse.collections.impl.utility.Iterate;
  * a BoxedMutableIntList will be released. Once available, ConjureIntegerList should be replaced with that.
  */
 final class ConjureIntegerList extends AbstractList<Integer> implements RandomAccess {
-    private final IntArrayList delegate;
+    private final MutableIntList delegate;
 
-    ConjureIntegerList(IntArrayList delegate) {
+    ConjureIntegerList(MutableIntList delegate) {
         this.delegate = delegate;
     }
 
@@ -68,5 +69,16 @@ final class ConjureIntegerList extends AbstractList<Integer> implements RandomAc
     @Override
     public Integer set(int index, Integer element) {
         return delegate.set(index, element);
+    }
+
+    ConjureIntegerList asUnmodifiable() {
+        return new ConjureIntegerList(delegate.asUnmodifiable());
+    }
+
+    // Cannot be named 'toArray' as that conflicts with the #toArray in AbstractList
+    // This is a serialization optimization that avoids boxing, but does copy
+    @JsonValue
+    int[] jacksonSerialize() {
+        return delegate.toArray();
     }
 }
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureSafeLongList.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureSafeLongList.java
@@ -16,11 +16,12 @@
 
 package com.palantir.conjure.java.lib.internal;
 
+import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.conjure.java.lib.SafeLong;
 import java.util.AbstractList;
 import java.util.Collection;
 import java.util.RandomAccess;
-import org.eclipse.collections.impl.list.mutable.primitive.LongArrayList;
+import org.eclipse.collections.api.list.primitive.MutableLongList;
 import org.eclipse.collections.impl.utility.Iterate;
 
 /**
@@ -28,9 +29,9 @@ import org.eclipse.collections.impl.utility.Iterate;
  * with SafeLongs.
  */
 final class ConjureSafeLongList extends AbstractList<SafeLong> implements RandomAccess {
-    private final LongArrayList delegate;
+    private final MutableLongList delegate;
 
-    ConjureSafeLongList(LongArrayList delegate) {
+    ConjureSafeLongList(MutableLongList delegate) {
         this.delegate = delegate;
     }
 
@@ -69,5 +70,16 @@ final class ConjureSafeLongList extends AbstractList<SafeLong> implements Random
     @Override
     public SafeLong set(int index, SafeLong element) {
         return SafeLong.of(delegate.set(index, element.longValue()));
+    }
+
+    ConjureSafeLongList asUnmodifiable() {
+        return new ConjureSafeLongList(delegate.asUnmodifiable());
+    }
+
+    // Cannot be named 'toArray' as that conflicts with the #toArray in AbstractList
+    // This is a serialization optimization that avoids boxing, but does copy
+    @JsonValue
+    long[] jacksonSerialize() {
+        return delegate.toArray();
     }
 }
--- a/changelog/@unreleased/pr-2386.v2.yml
+++ b/changelog/@unreleased/pr-2386.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: '[FR] Add Serialization Optimization for Primitive Collection Types'
+  links:
+  - https://github.com/palantir/conjure-java/pull/2386
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveExample.java
@@ -1,0 +1,181 @@
+package com.palantir.product;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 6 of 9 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -69,4 +80,22 @@ public void clear() {
...
+    }
+
+    static final class Deserializer extends JsonDeserializer<ConjureDoubleList> {
+        @Override
+        public ConjureDoubleList deserialize(JsonParser parser, DeserializationContext _ctxt) throws IOException {
+            // Avoid making a copy of the value from jackson
+            double[] doubles = parser.readValueAs(double[].class);
+            return new ConjureDoubleList(new DoubleArrayList(doubles));
+        }
+    }
```
carterkozak: I don't think these are used anywhere in practice -- Jackson isn't  provided the `ConjureDoubleList` type as a deserialization target, only `List<Double>`. On the serialization side, the constructed objects use `Collections.unmodifiableList` to wrap these specialized types, losing this specialization.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureSafeLongList.java
```diff
@@ -70,4 +81,24 @@ public void clear() {
...
+            gen.writeArray(longs, 0, longs.length);
+        }
+    }
+
+    static final class Deserializer extends JsonDeserializer<ConjureSafeLongList> {
+        @Override
+        public ConjureSafeLongList deserialize(JsonParser parser, DeserializationContext _ctxt) throws IOException {
+            // Avoid making a copy of the value from jackson
+            long[] longs = parser.readValueAs(long[].class);
+            return new ConjureSafeLongList(new LongArrayList(longs));
```
raiju: Maybe fine, but I think this will move the failure of unsafe longs from parse time to access time (and maybe allow proxying of unsafe longs depending on the exact conjure setup)

bmarcaur: O ew, nah, it needs to be upfront. Will incorporate when I have a moment to address feedback.

bmarcaur: I will address this when I reintroduce deserialization see: https://github.com/palantir/conjure-java/pull/2386/files#r1813038405

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
bmarcaur: I removed deserialization for now, will reintroduce later.

conjure-lib/src/test/java/com/palantir/conjure/java/lib/internal/ConjureCollectionsTest.java
```diff
@@ -82,4 +86,20 @@ public void newNonNullSafeLongList() {
...
+
+    @Test
+    public void testSerializationRoundtrip() throws JsonProcessingException {
+        List<Double> expectedDoubles = List.of(1.1, 2.2, 3.3);
+        testSerialization(ConjureCollections.newNonNullDoubleList(expectedDoubles), expectedDoubles);
+        List<Integer> expectedIntegers = List.of(1, 2, 3);
+        testSerialization(ConjureCollections.newNonNullIntegerList(expectedIntegers), expectedIntegers);
+        List<SafeLong> expectedSafeLongs = List.of(SafeLong.of(1L), SafeLong.of(2L), SafeLong.of(3L));
+        testSerialization(ConjureCollections.newNonNullSafeLongList(expectedSafeLongs), expectedSafeLongs);
+    }
```
carterkozak: serialization of the `ConjureCollections.newNonNull*List` isn't representative of usage within generated objects. Generated objects wrap these with `Collections.unmodifiableList`, so the type of the list that Jackson perceives is `java.util.Collections$UnmodifiableList` or `java.util.Collections$UnmodifiableRandomAccessList`, and the custom serializer isn't called.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -69,4 +75,13 @@ public void clear() {
     public Double set(int index, Double element) {
         return delegate.set(index, element);
     }
+
+    static final class Serializer extends JsonSerializer<ConjureDoubleList> {
+        @Override
+        public void serialize(ConjureDoubleList val, JsonGenerator gen, SerializerProvider _serializer)
+                throws IOException {
+            double[] doubles = val.delegate.toArray();
+            gen.writeArray(doubles, 0, doubles.length);
```
carterkozak: Does this impact our use of jackson-databind configuration options like excluding empty collections? https://github.com/palantir/conjure-java/blob/7ed0bdac3e25f5f567325cbf2ae580168f2cc99b/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java#L212-L218

carterkozak: I suspect we may be better off adding a `@JsonValue` annotation on the `toArray` method. A custom unmodifiable wrapper will need to be implemented to provide a `@JsonValue` annotated `toArray` method in a place that is visible to jackson (the outermost type)

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -16,20 +16,21 @@
...
+import org.eclipse.collections.api.list.primitive.MutableDoubleList;
 import org.eclipse.collections.impl.utility.Iterate;
 
 /**
  * ConjureDoubleList is a boxed list wrapper for the eclipse-collections DoubleArrayList. In eclipse-collections 12,
  * a BoxedMutableDoubleList will be released. Once available, ConjureDoubleList should be replaced with that.
  */
 final class ConjureDoubleList extends AbstractList<Double> implements RandomAccess {
-    private final DoubleArrayList delegate;
+    private final MutableDoubleList delegate;
```
bmarcaur, line 31: Changing this type so that I can wrap both `DoubleArrayList` and `UnmodifiableDoubleList` which share this interface as the lowest common denominator.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureIntegerList.java
```diff
@@ -16,20 +16,21 @@
...
+import org.eclipse.collections.api.list.primitive.MutableIntList;
 import org.eclipse.collections.impl.utility.Iterate;
 
 /**
  * ConjureIntegerList is a boxed list wrapper for the eclipse-collections IntArrayList. In eclipse-collections 12,
  * a BoxedMutableIntList will be released. Once available, ConjureIntegerList should be replaced with that.
  */
 final class ConjureIntegerList extends AbstractList<Integer> implements RandomAccess {
-    private final IntArrayList delegate;
+    private final MutableIntList delegate;
```
bmarcaur, line 31: https://github.com/palantir/conjure-java/pull/2386/files#r1842676255

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureSafeLongList.java
```diff
@@ -16,21 +16,22 @@
...
+import org.eclipse.collections.api.list.primitive.MutableLongList;
 import org.eclipse.collections.impl.utility.Iterate;
 
 /**
  * ConjureSafeLongList is a boxed list wrapper for the eclipse-collections LongArrayList. This handles boxing/unboxing
  * with SafeLongs.
  */
 final class ConjureSafeLongList extends AbstractList<SafeLong> implements RandomAccess {
-    private final LongArrayList delegate;
+    private final MutableLongList delegate;
```
bmarcaur, line 32: https://github.com/palantir/conjure-java/pull/2386/files#r1842676255

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -69,4 +70,15 @@ public void clear() {
...
+
+    public ConjureDoubleList asUnmodifiable() {
+        return new ConjureDoubleList(delegate.asUnmodifiable());
+    }
+
+    // Cannot be named 'toArray' as that conflicts with the #toArray in AbstractList
+    // This is a serialization optimization that avoids boxing, but does copy
+    @JsonValue
+    double[] jacksonSerialize() {
+        return delegate.toArray();
```
bmarcaur, line 82: This is how we avoid boxing out the door (I checked the Jackson code paths manually, it does infact operate on the array), while also respecting the various Jackson flags we use like `excludeEmptyCollections`. It does make a copy, but implementing a version that avoided this copy was very complex as it required leaking details of various Jackson settings.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureIntegerList.java
```diff
@@ -69,4 +70,15 @@ public void clear() {
...
+
+    public ConjureIntegerList asUnmodifiable() {
+        return new ConjureIntegerList(delegate.asUnmodifiable());
+    }
+
+    // Cannot be named 'toArray' as that conflicts with the #toArray in AbstractList
+    // This is a serialization optimization that avoids boxing, but does copy
+    @JsonValue
+    int[] jacksonSerialize() {
+        return delegate.toArray();
```
bmarcaur, line 82: https://github.com/palantir/conjure-java/pull/2386/files#r1842680511

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureSafeLongList.java
```diff
@@ -70,4 +71,15 @@ public void clear() {
...
+
+    public ConjureSafeLongList asUnmodifiable() {
+        return new ConjureSafeLongList(delegate.asUnmodifiable());
+    }
+
+    // Cannot be named 'toArray' as that conflicts with the #toArray in AbstractList
+    // This is a serialization optimization that avoids boxing, but does copy
+    @JsonValue
+    long[] jacksonSerialize() {
+        return delegate.toArray();
```
bmarcaur, line 83: https://github.com/palantir/conjure-java/pull/2386/files#r1842680511

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanGenerator.java
```diff
@@ -259,7 +260,7 @@ private static MethodSpec createConstructor(Collection<EnrichedField> fields, Co
             // is private and necessarily called from the builder, which does its own defensive copying.
             if (field.conjureDef().getType().accept(TypeVisitor.IS_LIST)) {
                 // TODO(melliot): contribute a fix to JavaPoet that parses $T correctly for a JavaPoet FieldSpec
-                body.addStatement("this.$1N = $2T.unmodifiableList($1N)", spec, Collections.class);
+                body.addStatement("this.$1N = $2T.unmodifiableList($1N)", spec, ConjureCollections.class);
```
bmarcaur: Pulled this out in to a separate PR so this one can continue to focus on the optimizations: https://github.com/palantir/conjure-java/pull/2409

conjure-java-core/src/test/resources/primitive-collections.yml
```diff
@@ -0,0 +1,9 @@
+types:
+  definitions:
+    default-package: com.palantir.product
+    objects:
+      PrimitiveExample:
```
bmarcaur, line 5: I made this separate of `ListExample` so that I could [control the generator settings separately](https://github.com/palantir/conjure-java/pull/2386/files#diff-cc2d5fa6d815eb2ce4ba4691ac1ceced1dd649598ba8400ab31cc3b1bced1b9bR178) which would allow me to test that my serialization optimization continued to respect various Jackson settings without mucking with existing tests.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -69,4 +70,15 @@ public void clear() {
     public Double set(int index, Double element) {
         return delegate.set(index, element);
     }
+
+    public ConjureDoubleList asUnmodifiable() {
```
carterkozak:
```suggestion
   ConjureDoubleList asUnmodifiable() {
```

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureIntegerList.java
```diff
@@ -69,4 +70,15 @@ public void clear() {
     public Integer set(int index, Integer element) {
         return delegate.set(index, element);
     }
+
+    public ConjureIntegerList asUnmodifiable() {
```
carterkozak:
```suggestion
   ConjureIntegerList asUnmodifiable() {
```

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureSafeLongList.java
```diff
@@ -70,4 +71,15 @@ public void clear() {
     public SafeLong set(int index, SafeLong element) {
         return SafeLong.of(delegate.set(index, element.longValue()));
     }
+
+    public ConjureSafeLongList asUnmodifiable() {
```
carterkozak:
```suggestion
   ConjureSafeLongList asUnmodifiable() {
```

carterkozak (approved):

PR #2389, opened 2024-10-22
[Feature] Non-boxing 'addAll' Methods for Primitive Types

bmarcaur:
## Before this PR
This is an alternative variant of https://github.com/palantir/conjure-java/pull/2312 which attempts to fix the boxing overhead merely introduced by the shape of our API (primitive types on both sides of the abstraction). However, there was some objection to that approach due to some stylistic and API design disagreements. This alternative seeks to remedy some of those concerns by following some "prior art" from libraries we accept / use all the time - Immutables.

Immutables will generate a vararg `addAll` variant for a `List` collection of some primitive boxed type. For example, `List<Double>` will produce something like:

```
public final Builder addField(double... elements)
```

## After this PR
This change first punches a hole through from the underlying eclipse-collections types e.g. `DoubleList` which has routines like `public boolean addAll(double... source)` which allow for direct, non-boxed, additions to the type.

Then, taking Immutables as inspiration, I wired up our codegen to leverage this abstraction. This alternative design addresses the mutability concerns as this API clearly communicates how data will be owned moving forward (i.e. copied). This implementation also addresses the concerns of having a setter that resets the type may cause order of operations bugs in the field while populating the builder.

## Of Note
Yes... I am aware that the `Iterable` variant is a setter in the builder e.g.:

```
@JsonSetter(value = "primitiveItems", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
public Builder primitiveItems(@Nonnull Iterable<Integer> primitiveItems) {
        checkNotBuilt();
        this.primitiveItems = ConjureCollections.newNonNullIntegerList(
                Preconditions.checkNotNull(primitiveItems, "primitiveItems cannot be null"));
        return this;
}
```

This feels like a design oversight (IMO) seeing as how `primitiveItems` is already initialized statically and `newNonNullIntegerList` simply performs a copy using `addAll`. Removing it from the field will be considerable work, so this will need to be inconsistent for the time being. And also previous design deficiencies cannot justify future designs.

## Tangent
Sorry if this is getting off topic, but that `Iterable` variant seems destined to be used incorrectly. Imagine the following code:

```java
// Has a field List<String> items
SomeConjureObject.Builder builder = SomeConjureObject.builder();
String firstToAdd = "first";
List<String> someOtherStrings = List.of(...);

// Clearly an adder
builder.items(firstToAdd);
// Adder? Setter? Setter! No longer contains firstToAdd
builder.items(someOtherStrings);

// Or how about this
// Imagine your code starts off like this
String firstToAdd = "first";
String secondToAdd = someMethod();
builder.items(firstToAdd);
builder.items(secondToAdd);

// But later the API of someMethod changes and now returns a List<String>
// the type checker is happy but now (below) is a setter and no longer
// contains firstToAdd
builder.items(secondToAdd);
```

Either way, necessary or not, I do not love the invalid state that is further representable on this API by adding an additional similarly named setter.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
+++ b/conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
@@ -234,6 +234,13 @@ public final class ListExample {
             return this;
         }
 
+        public Builder addAllPrimitiveItems(@Nonnull int... primitiveItems) {
+            checkNotBuilt();
+            ConjureCollections.addAllToIntegerList(
+                    this.primitiveItems, Preconditions.checkNotNull(primitiveItems, "primitiveItems cannot be null"));
+            return this;
+        }
+
         public Builder addAllPrimitiveItems(@Nonnull Iterable<Integer> primitiveItems) {
             checkNotBuilt();
             ConjureCollections.addAllAndCheckNonNull(
@@ -253,6 +260,13 @@ public final class ListExample {
             checkNotBuilt();
             this.doubleItems = ConjureCollections.newNonNullDoubleList(
                     Preconditions.checkNotNull(doubleItems, "doubleItems cannot be null"));
+            return this;
+        }
+
+        public Builder addAllDoubleItems(@Nonnull double... doubleItems) {
+            checkNotBuilt();
+            ConjureCollections.addAllToDoubleList(
+                    this.doubleItems, Preconditions.checkNotNull(doubleItems, "doubleItems cannot be null"));
             return this;
         }
 
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderAuxiliarySettersUtils.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderAuxiliarySettersUtils.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.java.types;
 
 import com.palantir.conjure.java.ConjureAnnotations;
+import com.palantir.conjure.java.lib.SafeLong;
 import com.palantir.conjure.java.types.BeanGenerator.EnrichedField;
 import com.palantir.conjure.java.util.Javadoc;
 import com.palantir.conjure.java.util.Primitives;
@@ -28,6 +29,7 @@ import com.palantir.conjure.spec.OptionalType;
 import com.palantir.conjure.spec.PrimitiveType;
 import com.palantir.conjure.spec.Type;
 import com.palantir.conjure.visitor.TypeVisitor;
+import com.palantir.javapoet.ArrayTypeName;
 import com.palantir.javapoet.ClassName;
 import com.palantir.javapoet.CodeBlock;
 import com.palantir.javapoet.FieldSpec;
@@ -39,11 +41,43 @@ import java.util.Optional;
 import javax.lang.model.element.Modifier;
 import org.apache.commons.lang3.StringUtils;
 
-public final class BeanBuilderAuxiliarySettersUtils {
+final class BeanBuilderAuxiliarySettersUtils {
 
     private BeanBuilderAuxiliarySettersUtils() {}
 
-    public static MethodSpec.Builder createCollectionSetterBuilder(
+    static MethodSpec.Builder createPrimitiveCollectionSetterBuilder(
+            EnrichedField enriched, TypeMapper typeMapper, ClassName returnClass, SafetyEvaluator safetyEvaluator) {
+        FieldSpec field = enriched.poetSpec();
+        FieldDefinition definition = enriched.conjureDef();
+        TypeName innerTypeName = extractInnerTypeFromList(definition, typeMapper, safetyEvaluator);
+
+        return MethodSpec.methodBuilder("addAll" + StringUtils.capitalize(field.name()))
+                .addJavadoc(Javadoc.render(definition.getDocs(), definition.getDeprecated())
+                        .map(rendered -> CodeBlock.of("$L", rendered))
+                        .orElseGet(() -> CodeBlock.builder().build()))
+                .addAnnotations(ConjureAnnotations.deprecation(definition.getDeprecated()))
+                .addModifiers(Modifier.PUBLIC)
+                // Forces the array argument to instead be variadic
+                .varargs()
+                .addParameter(Parameters.nonnullParameter(ArrayTypeName.of(innerTypeName), field.name()))
+                .returns(returnClass);
+    }
+
+    private static TypeName extractInnerTypeFromList(
+            FieldDefinition conjureDef, TypeMapper typeMapper, SafetyEvaluator safetyEvaluator) {
+        Type innerType = conjureDef.getType().accept(TypeVisitor.LIST).getItemType();
+        TypeName boxedTypeName = typeMapper.getClassName(innerType);
+
+        // SafeLong is just a special case of long
+        if (boxedTypeName.equals(ClassName.get(SafeLong.class))) {
+            return ConjureAnnotations.withSafety(TypeName.LONG, safetyEvaluator.getUsageTimeSafety(conjureDef));
+        } else {
+            return ConjureAnnotations.withSafety(
+                    Primitives.unbox(boxedTypeName), safetyEvaluator.getUsageTimeSafety(conjureDef));
+        }
+    }
+
+    static MethodSpec.Builder createCollectionSetterBuilder(
             String prefix,
             EnrichedField enriched,
             TypeMapper typeMapper,
@@ -66,7 +100,7 @@ public final class BeanBuilderAuxiliarySettersUtils {
                         field.name()));
     }
 
-    public static MethodSpec.Builder createOptionalSetterBuilder(
+    static MethodSpec.Builder createOptionalSetterBuilder(
             EnrichedField enriched, TypeMapper typeMapper, ClassName returnClass, SafetyEvaluator safetyEvaluator) {
         FieldSpec field = enriched.poetSpec();
         OptionalType type = enriched.conjureDef().getType().accept(TypeVisitor.OPTIONAL);
@@ -78,7 +112,7 @@ public final class BeanBuilderAuxiliarySettersUtils {
                         field.name()));
     }
 
-    public static MethodSpec.Builder createItemSetterBuilder(
+    static MethodSpec.Builder createItemSetterBuilder(
             EnrichedField enriched,
             Type itemType,
             TypeMapper typeMapper,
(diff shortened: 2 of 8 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
```diff
@@ -218,6 +218,13 @@ public Builder primitiveItems(@Nonnull Iterable<Integer> primitiveItems) {
             return this;
         }
 
+        public Builder addAllPrimitiveItems(@Nonnull int... primitiveItems) {
```
bmarcaur, line 237: Technically... if my words in the PR description are to be believed and I am mirroring Immutables, then this method should be generated with a name `primitiveItems`. Very easy change to make if people find that clearer.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureSafeLongList.java
```diff
@@ -56,6 +56,14 @@ public boolean addAll(int index, Collection<? extends SafeLong> collection) {
         return delegate.addAllAtIndex(index, target);
     }
 
+    public void addAll(long... source) {
+        for (long value : source) {
+            // Doesn't use SafeLong creation because this causes unnecessary boxing
```
bmarcaur: This comment is a little too specific and doesn't explain how the underlying type holds longs and those must adhere to the abstraction.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -158,11 +165,14 @@ public static List<Double> newNonNullDoubleList(Iterable<Double> iterable) {
...
     }
 
-    /**
-     * The following Conjure boxed list wrappers for the eclipse-collections [type]ArrayList are temporary (except
-     * ConjureSafeLongList). In eclipse-collections 12, a BoxedMutable[type]List will be released. Once available,
-     * Conjure[type]List should be replaced with that.
-     */
+    // This method modifies a list that can't handle nulls. Do not use this unless the nonNullCollections flag is set
+    public static void addAllToDoubleList(Collection<Double> addTo, double[] elementsToAdd) {
+        if (addTo instanceof ConjureDoubleList) {
```
bmarcaur, line 173: An inelegant solution to preserve the feedback on https://github.com/palantir/conjure-java/pull/2307#discussion_r1598822523, but still capture the optimization we added. Also has [some prior art in `addAll`.](https://github.com/palantir/conjure-java/blob/ae51640c5a77e5dc663750fb0e943aefc2a0dcd4/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java#L45)

conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
```diff
@@ -40,7 +40,7 @@ public long longValue() {
         return longValue;
     }
 
-    private static long check(long value) {
+    public static long check(long value) {
```
bmarcaur: I increased the visibility on this, because the logic is simple enough that to satisfy it is tantamount to copying the implementation detail (which felt bad). But also the logic is so simple that we aren't overly exposing some internals. Besides having more places rely on this is likely to prevent skew between implementations (IMO).

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -1039,22 +1076,28 @@ public boolean useNonNullFactory() {
...
-        LIST("List"),
-        DOUBLE_LIST("DoubleList"),
-        INTEGER_LIST("IntegerList"),
-        BOOLEAN_LIST("BooleanList"),
-        SAFE_LONG_LIST("SafeLongList"),
-        SET("Set");
+        LIST("List", false),
+        DOUBLE_LIST("DoubleList", true),
+        INTEGER_LIST("IntegerList", true),
+        BOOLEAN_LIST("BooleanList", true),
```
bmarcaur: But also maybe not? https://github.com/palantir/conjure-java/pull/2390

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderAuxiliarySettersUtils.java
```diff
@@ -66,7 +100,7 @@ public static MethodSpec.Builder createCollectionSetterBuilder(
                         field.name));
     }
 
-    public static MethodSpec.Builder createOptionalSetterBuilder(
```
bmarcaur, line 69: I can extract this into a separate PR, but I noticed that we were overly exposing these methods which in turn was overly exposing the `EnrichedField` type.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -767,10 +787,19 @@ private List<MethodSpec> createAuxiliarySetters(EnrichedField enriched, boolean
         Type type = enriched.conjureDef().getType();
         Optional<LogSafety> safety = safetyEvaluator.getUsageTimeSafety(enriched.conjureDef());
 
+        ImmutableList.Builder<MethodSpec> builder = ImmutableList.builder();
```
bmarcaur, line 790: Made this a builder so I could optionally add another method and not have to duplicate logic to return lists directly.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -1041,25 +1072,31 @@ public boolean useNonNullFactory() {
...
-        SET("Set");
+        BOOLEAN_LIST("List", false),
+        SAFE_LONG_LIST("SafeLongList", true),
+        SET("Set", false);
 
         private final String collectionName;
+        private final Boolean primitiveCollection;
 
-        ConjureCollectionType(String collectionName) {
+        ConjureCollectionType(String collectionName, Boolean primitiveCollection) {
```
carterkozak: please use the unboxed `boolean` type here, or another more expressive enum

conjure-java-core/src/integrationInput/java/com/palantir/product/SafeLongExample.java
```diff
@@ -135,6 +135,13 @@ public Builder safeLongList(@Nonnull Iterable<SafeLong> safeLongList) {
             return this;
         }
 
+        public Builder addAllSafeLongList(@Nonnull long... safeLongList) {
+            checkNotBuilt();
+            ConjureCollections.addAllToSafeLongList(
+                    this.safeLongList, Preconditions.checkNotNull(safeLongList, "safeLongList cannot be null"));
+            return this;
+        }
```
carterkozak: I thought we weren't adding public API taking `long` as `SafeLong` -- given that this doesn't quite match the other cases where we have an exact primitive type to operate upon, I think we should consider this case separately.

I'm all for making the non-public-api changes to optimize this case, but would prefer not to add overloads to generated public builders for it.

bmarcaur: I opted to disable the optimization for `SafeLong` for the moment as it was easier than implementing a new enum to replace the boolean which felt inappropriate for the scope of this current PR.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -55,6 +55,10 @@ public boolean addAll(int index, Collection<? extends Double> collection) {
         return delegate.addAllAtIndex(index, target);
     }
 
+    public void addAll(double... source) {
```
carterkozak: nit: this isn't an override, so the public modifier doesn't buy us anything -- might as well be package private

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureIntegerList.java
```diff
@@ -55,6 +55,10 @@ public boolean addAll(int index, Collection<? extends Integer> collection) {
         return delegate.addAllAtIndex(index, target);
     }
 
+    public void addAll(int... source) {
```
carterkozak: here as well

carterkozak: in practice perhaps these should take `int[]`/etc for clarity? If we're not careful with varargs, we can end up allocating (small) arrays unnecessarily

conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
carterkozak: Can the `@JsonSetter(value = "primitiveItems", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)` annotation be moved to these methods, allowing jackson to materialize the unboxed types instead of an arraylist of boxed primitives?

bmarcaur: Can that be done in a FLUP PR? One that is more focused on Deser optimizations?

bmarcaur: Also would it be worth it to instead move this to be `int[]` or to add an overload for such a method? I am wondering if Jackson will properly handle this signature.

carterkozak: > Can that be done in a FLUP PR? One that is more focused on Deser optimizations?

That sounds reasonable to me

> Also would it be worth it to instead move this to be int[] or to add an overload for such a method? I am wondering if Jackson will properly handle this signature.

We should go with either one or the other, imo not both.
Worth testing to validate jackson handles varargs as expected, but I'd imagine it handles varargs and arrays roughly equivalently -- that would be good to know before we commit to varargs though

bmarcaur: Yeah I can test that out

bmarcaur: Ok I wrote a small test and inspected the call stack using `int...` and `int[]` and Jackson invokes both methods equivalently. So I suspect that there isn't a runtime consideration and I think the varargs version is more usable (IMO). So I think leave this as is. Sound good?

As `int...`:
Screenshot 2025-01-10 at 10 15 28 AM

As `int[]`:
Screenshot 2025-01-10 at 10 16 00 AM

carterkozak: Perfect, thanks for digging into this!

carterkozak (approved):

conjure-java-core/src/integrationInput/java/com/palantir/product/ListExample.java
bmarcaur: > Can the `@JsonSetter(value = "primitiveItems", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)` annotation be moved to these methods, allowing jackson to materialize the unboxed types instead of an arraylist of boxed primitives?

Here is the FLUP: https://github.com/palantir/conjure-java/pull/2431

PR #2390, opened 2024-10-22
Remove Primitive Optimization for Boolean Collections

bmarcaur:
## Before this PR
`list<boolean>` was optimized in [this move to eclipse-collections](https://github.com/palantir/conjure-java/pull/2307), but its [usage is almost non-existent](https://pl.ntr/2o-) (all instances appear to be some copy-paste from Excavator?). IMO, it's usage, or really this optimization, signals to devs that using a wire format of `list<boolean>` is a good idea, when a bit mask would be a much better abstraction for such a structure when it comes to our serialization.

On the other hand, it feels like an odd omission given the shape of the feature. Net net, I think removing it lowers our ownership overhead of this. `boolean` is a weird primitive type and when tackling changes for the 4 primitive types `boolean` is always a little unique. For example:
- When [adding serialization optimizations](https://github.com/palantir/conjure-java/pull/2386) it did not make sense to change the serializer because a bit is turned into the string literal `false` or `true`.
- Additionally, when [trying to optimize the `addAll` paths for primitive arrays](https://github.com/palantir/conjure-java/pull/2389) the `boolean` variant of the methods needed some different routines due to `Arrays.stream` not having a `boolean` variant.

## After this PR
Treated like any old boxed type, in the long term we should consider a bit mask type for an efficient representation of the usage pattern.

## Possible downsides?
This _may_ regress the peak memory usage on the small use cases, but also let's consider leveraging [more expressive external types](https://github.com/palantir/conjure/pull/1661) to experiment.


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -182,24 +181,6 @@ public static List<Integer> newNonNullIntegerList(Iterable<Integer> iterable) {
...
-        List<Boolean> booleanList;
-        if (iterable instanceof Collection) {
-            booleanList = new ConjureBooleanList(new BooleanArrayList(((Collection<Boolean>) iterable).size()));
-        } else {
-            booleanList = new ConjureBooleanList(new BooleanArrayList());
-        }
-        addAll(booleanList, iterable);
-
-        return booleanList;
-    }
```
carterkozak: Removing these public apis is pretty dangerous, it will cause existing published code that has already been generated to fail at runtime when `conjure-lib` is upgraded.
They could instead be stubbed out as aliases to non-specialized `newNonNullList()`/etc.

bmarcaur: Hmm my thinking on this is that `conjure-lib` is bumped in lockstep with the code that generates the conjure classes. But thinking about that more deeply, even though they are housed in the same repo they are delivered by different mechanisms, `versions.props` vs a gradle plugin? So you are envisioning a scenario where `conjure-lib` is bumped by GCV, yet the code gen portions remain on an older gradle version, which has its own classpath resolution, and thus they can drift and this is dangerous?

carterkozak: Not quite. Let’s say you publish an api, the types use this method. Your consumer which uses a client to your service upgrades all their conjure deps. 💥

bmarcaur: I addressed your concerns. I added a `@DoNotCall` I _think_ this be ignored for generated code which is the only legitimate use case. Correct me if I am wrong.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -182,22 +182,28 @@ public static List<Integer> newNonNullIntegerList(Iterable<Integer> iterable) {
...
 
-    // This method returns a list that can't handle nulls. Do not use this unless the nonNullCollections flag is set
+    /**
+     * This should only ever be called by a previously generated conjure internal implementation.
+     *
+     * @return a nonnull boxed list
+     * @deprecated This method is here for GCV back-compat purposes
+     */
+    @Deprecated
+    @DoNotCall
```
carterkozak: let's not chance it with these annotations -- some projects override our defaults and get warnings from generated code. It's not great, but I don't want to block conjure-java rollout unnecessarily. We can recommend against these methods in docs, then deprecate/dnc in a month or two.

carterkozak (approved):
lgtm, thanks!

PR #2401, opened 2024-11-06
 Support endpoint errors

mpritham:
## Before this PR
Endpoint errors were [added to the Conjure specification in this PR](https://github.com/palantir/conjure/pull/1670). conjure-java does not use user-defined endpoint errors.

### Generated `ServerErrors` classes
Developers will be able to create checked exceptions for each error that is defined as an error an endpoint can throw. Specifically, for each endpoint-defined error, `conjure-java` will generate a class for the error's namespace with name `{namespace}ServerErrors`. This class will hold the definition of an exception, which is a sub-class of `CheckedServiceException` [(introduced here)](https://github.com/palantir/conjure-java-runtime-api/pull/1262), as well as methods to construct the exception.

### Changes to `Errors` classes
When an error is used as an endpoint error, factory methods to create `ServiceException`s are no longer provided in the `Errors` utility classes. 🌶️ This means that once an error becomes associated with an endpoint, it can no longer be used to throw un-checked runtime exceptions. This is intentional and acts as a forcing function for adoption: associating errors with one endpoint will require users to add endpoint errors for all of the endpoints throwing the error.

### Updates to Undertow Service Interfaces and Handlers
For each endpoint that defines endpoint-errors, the corresponding endpoint in the Undertow service interface, and the endpoint handler method, will include a `throws` clause including the list of checked exceptions corresponding to the endpoint errors, that were generated in the `ServerErrors` class (see above).   

### Jersey
Endpoint errors will not be supported on Jersey servers. An exception will be thrown if a user attempts to generate code for Jersey services with endpoint errors defined.

==COMMIT_MSG==
Support endpoint errors
==COMMIT_MSG==


versions.props
```diff
@@ -1,4 +1,4 @@
-ch.qos.logback:* = 1.2.11
+ch.qos.logback:* = 1.3.0
```
mpritham: [This PR bumped slf4j-api in conjure](https://github.com/palantir/conjure/pull/1492/files). Versions of logback < 1.3.x are no longer compatible with slf4j-api versions > 2.0.x. [Docs here](https://www.slf4j.org/codes.html#ignoredBindings)

conjure-java-core/src/integrationInput/java/com/palantir/product/UndertowErrorService.java
```diff
@@ -0,0 +1,31 @@
...
+     * @apiNote {@code GET /base/imported}
+     * @throws EndpointError
+     */
+    String testImportedError(AuthHeader authHeader) throws ServerEndpointSpecificErrors.EndpointError;
+
+    /**
+     * @apiNote {@code GET /base/multiple}
+     * @throws InvalidArgument
+     * @throws NotFound Something was not found.
+     * @throws DifferentNamespace
```
mpritham: I should add the package and namespace here if there are name conflicts. It'll be a bit cluttered but it's needed to disambiguate.

conjure-java-core/src/test/resources/errors-for-endpoints.yml
```diff
@@ -0,0 +1,22 @@
+types:
```
mpritham, line 1: I didn't want to re-use the errors in `example-errors.yml`. If I used any of those errors in an endpoint, the `ErrorGenerator` would not create the static `ServiceException` factory methods for them. This would conflict with the generated error classes. I don't want to change the current `ErrorGenerator` tests.

conjure-java-core/src/integrationInput/java/com/palantir/product/ServerEndpointSpecificTwoErrors.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.palantir.conjure.java.undertow.lib.CheckedServiceException;
+import javax.annotation.Nullable;
+import javax.annotation.processing.Generated;
+
+@Generated("com.palantir.conjure.java.types.CheckedErrorGenerator")
+public final class ServerEndpointSpecificTwoErrors {
+    private ServerEndpointSpecificTwoErrors() {}
+
+    public static DifferentNamespace differentNamespace() {
```
mpritham: I forget what we landed on when we discussed this live, but I'm opting to not add the `throwIf` variant. It sounds like that is some syntactic sugar that devs can define in their projects if they need.

mpritham: hm, I actually see it's used a bunch. Including the variant will make the migration easier.

conjure-java-core/src/integrationInput/java/com/palantir/product/UndertowErrorService.java
carterkozak: I think we want a javapoet `$T` reference using the `TypeName` of the error, that way javadoc generation links to the exception

versions.props
mpritham: The latest 2.x version of Dropwizard (used in our tests) depends on [a version of logback < 1.3.x](https://github.com/dropwizard/dropwizard/blame/v2.1.12/dropwizard-dependencies/pom.xml#L58). This is causing the AutoDeserializeTest cases to fail with:

```
Caused by: java.lang.ClassNotFoundException: ch.qos.logback.classic.jmx.JMXConfigurator
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
```

[The JMX Configurator was removed from logback.](https://logback.qos.ch/news.html#1.3.0:~:text=The%20ch.qos.logback.classic.jmx%20package%20was%20removed%20for%20security%20reasons%20and%20for%20lack%20of%20use.)

I can either:
1. bump our dropwizard dep to the latest version and migrate our tests to use it
1. pin slf4j-api to a version < 2.x
1. see if we can change our tests to not depend on dropwizard.

Going to check how much lift option 1 is

mpritham: Going to pin slf4j-api to a pre 2.x version for these tests.

mpritham:
This is good for a re-review. Shouldn't be merged: it has the CheckedServiceException class which we'd like to be in the conjure-java-runtime-api repo, it uses an RC version of conjure. 

Going to branch off of this to work on the error serialization piece.

conjure-java-core/src/integrationInput/java/com/palantir/another/ServerConjureErrors.java
```diff
@@ -0,0 +1,28 @@
...
+    public static final class DifferentPackage extends CheckedServiceException {
+        private DifferentPackage() {
+            super(com.palantir.another.ConjureErrors.DIFFERENT_PACKAGE);
+        }
+
+        private DifferentPackage(@Nullable Throwable cause) {
+            super(com.palantir.another.ConjureErrors.DIFFERENT_PACKAGE, cause);
+        }
+    }
+}
```
carterkozak: I suspect this file can be deleted, it appears to have been generated with an older version of the codegen (no throwIf util functions, and still has two ctors for the exception)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -368,6 +373,14 @@ private TypeSpec generateEndpointHandler(
         return endpointBuilder.build();
     }
 
+    private static ClassName createEndpointErrorException(EndpointError endpointError, Options options) {
```
carterkozak: Can we share this between `CheckedErrorGenerator.errorExceptionsClassName` and this class? Perhaps in ErrorGenerationUtils?

conjure-java-core/src/main/java/com/palantir/conjure/java/util/ErrorGenerationUtils.java
```diff
@@ -0,0 +1,255 @@
...
+                    .name(errorName.getName())
+                    .package_(errorName.getPackage())
+                    .namespace(errorDefinition.getNamespace())
+                    .build();
+            return errors.contains(errorTypeName);
+        }
+    }
+
+    public record ErrorDefinitionsByPackageAndNamespace(
+            Map<String, Map<ErrorNamespace, List<ErrorDefinition>>> packageToNamespacedErrorDefs) {
```
carterkozak: I think this is still a bit of a legibility issue -- we've wrapped the nested generics, but haven't eliminated them.

Would it be reasonable to replace this with a `List<NamespacedErrors>` where `NamespacedErrors` is roughly shaped:
```java
record NamespacedErrors(String javaPackage, ErrorNamespace namespace, List<ErrorDefinition> errors) {}
```

Building the model would require some grouping with similar maps, but the data we're left with to hand the generator is a bit easier to understand. Also, when we build this data model, we can start with something like `record ErrorGroup(String javaPackage, ErrorNamespace namespace)` as the key,  in order to avoid multiple nested maps. We don't need to group the same package together, it's only errors with the same package and namespace in common.

conjure-java-core/src/integrationInput/java/com/palantir/another/ServerConjureErrors.java
mpritham: Sorry, you're right! Will remove.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
mpritham: Yep. Updated. Thought of putting this in `CheckedErrorGenerator` since that's the file which creates the `ServerXErrors` files, but the utils seems more appropriate.

conjure-java-core/src/main/java/com/palantir/conjure/java/util/ErrorGenerationUtils.java
mpritham: Yep, that's cleaner. Updated.

conjure-java-core/src/main/java/com/palantir/conjure/java/util/ErrorGenerationUtils.java
```diff
@@ -76,27 +78,25 @@ public boolean contains(ErrorDefinition errorDefinition) {
         }
     }
 
-    public record ErrorDefinitionsByPackageAndNamespace(
-            Map<String, Map<ErrorNamespace, List<ErrorDefinition>>> packageToNamespacedErrorDefs) {
+    public record ErrorDefinitionsByPackageAndNamespace(List<NamespacedErrors> namespacedErrors) {
```
carterkozak: I don't think record buys us much, using `List<NamespacedErrors> namespacedErrors` directly is pretty clear imo

mpritham: Yeah, it no longer makes sense to try an encapsulate that logic. Moved to using a method that just returns `List<NamespacedErrors>`.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceInterfaceGenerator.java
```diff
@@ -96,17 +99,69 @@ private MethodSpec generateServiceInterfaceMethod(
...
+        Optional<String> docs = endpointDefinition.getDocs().map(Javadoc::render);
+        Optional<String> requestLine = Optional.of(
+                Javadoc.getRequestLine(endpointDefinition.getHttpMethod(), endpointDefinition.getHttpPath()));
+        Optional<String> params = Optional.ofNullable(Strings.emptyToNull(endpointDefinition.getArgs().stream()
+                .flatMap(argument -> Javadoc.getParameterJavadoc(argument, endpointDefinition).stream())
+                .collect(Collectors.joining("\n"))));
+        StringBuilder sb = new StringBuilder();
+        docs.ifPresent(sb::append);
+        requestLine.ifPresent(sb::append);
+        params.ifPresent(sb::append);
```
mpritham: This bit is duplicate code. We could create 2 methods: 1) add `docs`, `requestLine`, `params` 2) add `depr`, `incDoc`, then sandwich the the endpoint error docs in between the two.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceInterfaceGenerator.java
```diff
@@ -96,17 +99,69 @@ private MethodSpec generateServiceInterfaceMethod(
...
+            Optional<String> maybePackagePrefix) {
+        Optional<String> depr = endpointDefinition.getDeprecated().map(Javadoc::getDeprecatedJavadoc);
+        Optional<String> incDoc = Javadoc.getIncubatingJavadoc(endpointDefinition.getTags());
+        Optional<String> docs = endpointDefinition.getDocs().map(Javadoc::render);
+        Optional<String> requestLine = Optional.of(
+                Javadoc.getRequestLine(endpointDefinition.getHttpMethod(), endpointDefinition.getHttpPath()));
+        Optional<String> params = Optional.ofNullable(Strings.emptyToNull(endpointDefinition.getArgs().stream()
+                .flatMap(argument -> Javadoc.getParameterJavadoc(argument, endpointDefinition).stream())
+                .collect(Collectors.joining("\n"))));
+        StringBuilder sb = new StringBuilder();
```
carterkozak: Instead of collecting to this stringbuilder, would it make sense to individually apply each component to `methodBuilder.addJavadoc` if non-empty? It's possible that might do odd things to spacing, but worth trying, I think it could simplify this quite a bit

conjure-java-core/src/main/java/com/palantir/conjure/java/util/Javadoc.java
```diff
@@ -82,7 +82,7 @@ public static Optional<String> getParameterJavadoc(ArgumentDefinition argument,
                         + JavaNameSanitizer.sanitizeParameterName(
                                 argument.getArgName().get(), endpoint)
                         + " "
-                        + Javadoc.render(argument.getDocs().get()));
+                        + Javadoc.render(docs));
```
carterkozak: ha, good catch.

(no change requested, up to you if you'd like to break it out) For this sort of drive-by on a larger PR, sometimes I prefer to create a new small PR into develop for the cleanup. That way it's super easy to review, and keeps the large feature PRs cleaner.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceInterfaceGenerator.java
```diff
@@ -96,17 +99,69 @@ private MethodSpec generateServiceInterfaceMethod(
...
         } else {
             endpointDef.getReturns().ifPresent(type -> methodBuilder.returns(returnTypeMapper.getClassName(type)));
         }
 
+        methodBuilder.addExceptions(endpointDef.getErrors().stream()
+                .map(endpointError -> {
+                    ErrorTypeName errorTypeName = endpointError.getError();
+                    return ClassName.get(
+                            Packages.getPrefixedPackage(errorTypeName.getPackage(), options.packagePrefix()),
+                            "Server" + errorTypeName.getNamespace() + "Errors",
```
carterkozak: It might be helpful to take all the occurrences of `"Server" + namespace + "Errors"` and consolidate them into a utility function.

I think we're at the point in the process where I question my instincts on naming (which I'm the first to admit I'm truly terrible at!), and wonder if we should opt for `namespace + "ServerErrors"` so that namespace provides tab completion. What do you think? (tagging @bjlaub and @bmarcaur to opine on this as well).

The other piece to consider is collision between the existing and server-side error utilities: If you have a namespace called `ServerFoo` and another called `Foo`, you'd end up with two classes called `ServerFoo` potentially with the same fully qualified package name. The fix is pretty simple, though, a java wire break can be avoided by changing the `package` of one of the two, but it's a bit of friction that would be nice to avoid if possible.

mpritham: We sync'd earlier and came to the conclusion that we should use `Server<namespace>Errors`. Though now, I I like the discoverability that `<namespace>ServerErrors` provides. The only advantage of prefixing the current errors class (i.e. using `Server<namespace>Errors`) was to make migrations slightly less painful.

I'll update.

conjure-java-core/src/integrationInput/java/com/palantir/another/ServerEndpointSpecificErrors.java
```diff
@@ -0,0 +1,37 @@
+package com.palantir.another;
+
+import com.palantir.conjure.java.api.errors.CheckedServiceException;
+import javax.annotation.Nullable;
+import javax.annotation.processing.Generated;
+import org.jetbrains.annotations.Contract;
+
+@Generated("com.palantir.conjure.java.types.CheckedErrorGenerator")
+public final class ServerEndpointSpecificErrors {
```
carterkozak: I think we need to delete the orphaned `Server`+namspace+`Errors` classes

conjure-java-core/src/main/java/com/palantir/conjure/java/services/ServiceGenerators.java
```diff
@@ -17,45 +17,63 @@
...
+import com.palantir.javapoet.MethodSpec;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
 public final class ServiceGenerators {
+    public record EndpointJavaDocGenerationOptions(boolean includeRequestLine, boolean includeEndpointErrors) {}
 
-    public static Optional<String> getJavaDoc(EndpointDefinition endpointDef) {
-        return getJavaDocInternal(endpointDef, false);
+    public static void addJavaDocForEndpointDefinition(
```
mpritham, line 44: Introduced this method which adds the javadoc to the provided `methodBuilder`. I opted to add an "options" record because there are two dimensions of generation options we need to support: 1) if we should include the request line (e.g. `@apiNote {@code GET /foo/bar}`), 2) if we should include endpoint errors.

| Usage | Request Line? | Endpoint Errors? | 
| --- | --- | --- |
| Jersey service generation | No | No |
| Undertow service generation | Yes | Yes |
| Dialogue interface generation | Yes | No* |

*Not yet at least. We can either decide to return a result type in Dialogue, or throw a checked exception. If we do the latter, we need to add `@throws` pieces to the JavaDoc similar to what we do when generating Undertow service interfaces. If we go with the former, we'll still need to document the result union type but the JavaDoc would be added where we define the results type.

mpritham: For Jersey service generation, we shouldn't even get to the logic that adds the JavaDoc. Endpoint errors are not supported for Jersey, and we will throw an `IllegalArgumentException` if a user attempts to code-gen Jersey services with endpoint errors defined.

carterkozak: Agreed, I think we always want to add the error javadoc if errors are present -- we validate that they aren't when the legacy Jersey generator is used

conjure-java-core/src/main/java/com/palantir/conjure/java/services/ServiceGenerators.java
```diff
@@ -17,45 +17,63 @@
...
+import com.palantir.conjure.java.util.Packages;
 import com.palantir.conjure.spec.EndpointDefinition;
+import com.palantir.javapoet.ClassName;
+import com.palantir.javapoet.CodeBlock;
+import com.palantir.javapoet.MethodSpec;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
 public final class ServiceGenerators {
+    public record EndpointJavaDocGenerationOptions(boolean includeRequestLine, boolean includeEndpointErrors) {}
```
carterkozak: We could use an enum instead of booleans for clarity e.g.

```java
enum RequestLineJavadoc {
    INCLUDE,
    EXCLUDE
}
```

methods/ctors which take multiple booleans are easy to get wrong, and difficult to read. IDE features help, but the github UI isn't so kind

conjure-java-core/src/integrationInput/java/com/palantir/another/ServerEndpointSpecificErrors.java
mpritham: Yes we do, thanks. I wonder if there's a build time check we could add to ensure we don't accidentally check these in.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/ServiceGenerators.java
mpritham: Yep that's cleaner. Updated.

conjure-java-core/src/main/java/com/palantir/conjure/java/util/Javadoc.java
mpritham: Made #2406 instead of adding the drive-by here.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceInterfaceGenerator.java
mpritham: Updated, that made things simpler.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/ServiceGenerators.java
carterkozak: Ah right, sorry, missed the piece about dialogue, definitely don't want to leak the server-side piece there!

versions.props
```diff
@@ -6,10 +6,10 @@ com.google.code.findbugs:jsr305 = 3.0.2
 com.google.guava:guava = 33.1.0-jre
 com.google.testing.compile:compile-testing = 0.21.0
 com.palantir.common:streams = 2.3.0
-com.palantir.conjure.java.api:* = 2.54.0
+com.palantir.conjure.java.api:* = 2.56.0-rc1
 com.palantir.conjure.java.runtime:* = 8.7.0
 com.palantir.conjure.verification:* = 0.19.0
-com.palantir.conjure:* = 4.36.0
+com.palantir.conjure:* = 4.50.0-rc2
```
carterkozak: we should be able to update both of these now

carterkozak:
Looks good, just needs a changelog!

carterkozak (approved):

PR #2402, opened 2024-11-08
Ensure endpoint associated error serialization follows the Conjure spec

mpritham:
## Before this PR
Builds on #2401. Serialize the arguments in `CheckedServiceException`s as `ConjureError`s

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java
```diff
@@ -0,0 +1,50 @@
...
+        @JsonProperty("errorCode") String errorCode,
+        @JsonProperty("errorName") String errorName,
+        @JsonProperty("errorInstanceId") String errorInstanceId,
+        @JsonProperty("parameters") Map<String, Object> parameters) {
+
+    ConjureError {
+        Preconditions.checkNotNull(errorCode, "errorCode cannot be null");
+        Preconditions.checkNotNull(errorName, "errorName cannot be null");
+        Preconditions.checkNotNull(parameters, "parameters cannot be null");
+        // TODO(pm): why was there a default empty string value for errorInstanceId in SerializableError?
```
carterkozak: My guess is backcompat. Not needed here!

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java
```diff
@@ -0,0 +1,50 @@
...
+import com.palantir.conjure.java.undertow.lib.CheckedServiceException;
+import com.palantir.logsafe.Preconditions;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+record ConjureError(
+        @JsonProperty("errorCode") String errorCode,
+        @JsonProperty("errorName") String errorName,
+        @JsonProperty("errorInstanceId") String errorInstanceId,
+        @JsonProperty("parameters") Map<String, Object> parameters) {
```
carterkozak, line 58: Worth adding some test coverage on the serialization here, but I suspect we may want to add `@JsonInclude(content = Include.NON_ABSENT)` to the parameters map, so we can avoid sending explicit `null` values for empty optional parameters. Upside of that behavior is that it's easier for folks to deprecate and remove fields from APIs, which shouldn't be relied upon, but folks try to do it anyhow, so we might as well help them out!

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -186,17 +199,33 @@ private static void error(HttpServerExchange exchange, Error error) {
...
+            if (maybeBody.isPresent()) {
+                serializer.serialize(maybeBody.get(), exchange);
+            }
+        });
+    }
+
+    @FunctionalInterface
+    private interface SerializerRunnable {
+        void run() throws IOException;
+    }
```
carterkozak: not 100% on it at this point, but I'm curious what happens to the diff if we add `static ConjureError fromServiceException(ServiceException exception) {` to `ConjureError` and use the new type for all cases in this file? The new factory would need to implement the old behavior which calls `Objects.toString` on all inputs, to avoid making wire changes.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java
```diff
@@ -0,0 +1,50 @@
...
+
+package com.palantir.conjure.java.undertow.runtime;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.palantir.conjure.java.undertow.lib.CheckedServiceException;
+import com.palantir.logsafe.Preconditions;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+record ConjureError(
```
carterkozak, line 54: Let's add some javadoc to this class describing why we're using it instead of SerializableError

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
```diff
@@ -80,6 +85,59 @@ public void handlesServiceException() throws IOException {
...
+        assertThat(connection.getResponseCode()).isEqualTo(ErrorType.CONFLICT.httpErrorCode());
+        String expectedSerializedArg =
+                "\"Argument[field1=foo, maybeField2=Optional[42], collectionField=[bar, baz], nullString=null]\"";
+        assertThat(getErrorBody(connection))
+                .contains("{\"errorCode\":\"CONFLICT\"")
+                .contains("\"parameters\":{\"arg\":" + expectedSerializedArg + "}}");
+    }
+
+    @Test
+    public void handlesCheckedServiceException() throws IOException {
```
mpritham, line 113: Adding more tests with interesting parameter types: `Optional`, `Collections` and combinations of the two.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java
```diff
@@ -0,0 +1,100 @@
...
+        Preconditions.checkNotNull(errorInstanceId, "errorInstanceId cannot be null");
+        Preconditions.checkNotNull(parameters, "parameters cannot be null");
+    }
+
+    static ConjureError fromCheckedServiceException(CheckedServiceException exception) {
+        Map<String, Object> parameters = new HashMap<>();
+        for (Arg<?> arg : exception.getArgs()) {
+            if (arg.getValue() == null) {
+                continue;
+            }
```
mpritham: I was thinking of the case where service developers can add `Optional` fields to their error definitions without causing a break so long as they send `Optional.empty` until clients roll onto a new version of their API.

But we get even better back-compat safety by having the client deserializer ignore unknown properties. I think using a deserializer which ignores unknown properties sounds reasonable. Adding a whole new error that an endpoint can throw is certainly an API break and should fail loudly. While the fields in an error are part of the API surface, I'm not sure we need to have clients fail if they encounter a field in an error they haven't seen.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
mpritham: This works. Removed the use of the slightly hacky `SerializerRunnable`. I also added a `fromRemoteException` version.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java
carterkozak: Doesn't `@JsonInclude(content = Include.NON_ABSENT)` take care of this for us already? The same applies to Optional.empty, OptionalInt.empty, etc.

> But we get even better back-compat safety by having the client deserializer ignore unknown properties. I think using a deserializer which ignores unknown properties sounds reasonable

Our clients use [ObjectMappers. newClientJsonMapper()](https://github.com/palantir/conjure-java-runtime/blob/349e914a8620a5141c78da2f2ea6a48698f71d25/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java#L52-L56) which sets `.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)` for precisely this reason -- this is true of all response types :-)

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
carterkozak: The piece I'm most interested in testing is `SafeArg.of("emptyOptional", Optional.empty()), SafeArg.of("null", null));`

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -186,17 +192,14 @@ private static void error(HttpServerExchange exchange, Error error) {
         writeResponse(exchange, Optional.empty(), ErrorType.INTERNAL.httpErrorCode());
     }
 
-    private static void writeResponse(
-            HttpServerExchange exchange, Optional<SerializableError> maybeBody, int statusCode) {
+    private static void writeResponse(HttpServerExchange exchange, Optional<ConjureError> maybeBody, int statusCode) {
         // Do not attempt to write the failure if data has already been written
         if (!isResponseStarted(exchange)) {
             exchange.setStatusCode(statusCode);
-            if (maybeBody.isPresent()) {
```
carterkozak: was this intentionally deleted? I think the empty case will produce an exception and log it at info as `log.info("Failed to write error response", e);`

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
```diff
@@ -80,6 +87,90 @@ public void handlesServiceException() throws IOException {
...
+                Integer integer,
+                Integer nullValue,
+                OptionalInt optionalInt,
+                OptionalInt emptyOptionalInt,
+                OptionalDouble optionalDouble,
+                OptionalDouble emptyOptionalDouble,
+                Optional<String> optional,
+                Optional<String> emptyOptional,
+                Optional<List<String>> optOfList,
+                List<Optional<String>> listOfOpt) {}
```
carterkozak: I don't think this tells us much, we can trust the conjure ObjectMapper to serialize conjure-generated objects correctly (we have plenty of tests for that!)

it's the top-level fields used as SafeArg and UnsafeArg values themselves that are important to test here.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java
mpritham: > Doesn't @JsonInclude(content = Include.NON_ABSENT) take care of this for us already?

It didn't. I think it has to do with the fact that the value is `Object`. The `NON_ABSENT` flag worked when I used an `Optional` as the value. I wrote a custom contentFilter.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
mpritham: Yeah that makes sense. I updated the tests.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
mpritham: Sorry, this was a typo.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java
```diff
@@ -0,0 +1,117 @@
...
+                    || (obj instanceof OptionalInt && ((OptionalInt) obj).isEmpty())
+                    || (obj instanceof OptionalLong && ((OptionalLong) obj).isEmpty())
+                    || (obj instanceof OptionalDouble && ((OptionalDouble) obj).isEmpty());
+        }
+
+        @Override
+        public int hashCode() {
+            return getClass().hashCode();
+        }
+    }
```
carterkozak: I see, given this, I think this will be easier to follow (and involve less reflection) if we do away with the `@JsonInclude` annotation in favor of filtering args that we place in the map (similar to your original implementation, but with additional instanceof checks for the other optional types).
Sorry for the runaround on this one!

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
```diff
@@ -80,6 +86,74 @@ public void handlesServiceException() throws IOException {
...
+                String field1, Optional<Integer> maybeField2, List<String> collectionField, String nullString) {}
+        exception = new ServiceException(
+                ErrorType.CONFLICT,
+                SafeArg.of("arg", new Argument("foo", Optional.of(42), List.of("bar", "baz"), null)));
+        HttpURLConnection connection = execute();
+
+        assertThat(connection.getResponseCode()).isEqualTo(ErrorType.CONFLICT.httpErrorCode());
+        String expectedSerializedArg =
+                "\"Argument[field1=foo, maybeField2=Optional[42], collectionField=[bar, baz], nullString=null]\"";
+        assertThat(getErrorBody(connection))
```
carterkozak: We could deserialize into the `ConjureError` type here in order to get structured data -- that way tests won't fail when order of args changes, as order is not guaranteed by conjure wire spec

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
```diff
@@ -80,6 +86,74 @@ public void handlesServiceException() throws IOException {
                 .contains("\"parameters\":{\"foo\":\"bar\"}}");
     }
 
+    @Test
+    public void checkServiceExceptionArgumentsSerializedWithToString() throws IOException {
+        record Argument(
+                String field1, Optional<Integer> maybeField2, List<String> collectionField, String nullString) {}
+        exception = new ServiceException(
+                ErrorType.CONFLICT,
+                SafeArg.of("arg", new Argument("foo", Optional.of(42), List.of("bar", "baz"), null)));
```
carterkozak: thought on extracting `Argument` to a variable, that way instead of encoding `String expectedSerializedArg = <string literal>` we can assert against `argument.toString()`? that keeps the code a bit easier to follow, and ensures that if the toString changes slightly, the test won't fail.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
```diff
@@ -80,6 +86,74 @@ public void handlesServiceException() throws IOException {
...
+        HttpURLConnection connection = execute();
+
+        assertThat(connection.getResponseCode()).isEqualTo(ErrorType.CONFLICT.httpErrorCode());
+        String expectedSerializedArg =
+                "\"Argument[field1=foo, maybeField2=Optional[42], collectionField=[bar, baz], nullString=null]\"";
+        assertThat(getErrorBody(connection))
+                .contains("{\"errorCode\":\"CONFLICT\"")
```
carterkozak: We should verify the errorName in addition to the errorCode (name is arguably much more important!)
parsing to either `ConjureError` or a locally-defined variant for testing will make that more robust than string matching json contents

settings.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
         gradlePluginPortal() { metadataSources { mavenPom(); ignoreGradleMetadataRedirection() } }
     }
     dependencies {
-        classpath 'com.palantir.gradle.jdks:gradle-jdks-settings:0.55.0'
+        classpath 'com.palantir.gradle.jdks:gradle-jdks-settings:0.56.0'
```
carterkozak: I think this is left over from a bad merge/rebase

carterkozak (dismissed):
lgtm, thanks!

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java
mpritham: No worries! This was pretty non-obvious to me.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
mpritham: Updated to also check the error name.

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
```diff
@@ -80,6 +92,89 @@ public void handlesServiceException() throws IOException {
...
+
+        exception = new OptionalException();
+        HttpURLConnection connection = execute();
+
+        // Then:
+        assertThat(connection.getResponseCode()).isEqualTo(ErrorType.CONFLICT.httpErrorCode());
+        ConjureError error = CLIENT_JSON_MAPPER.readValue(getErrorBody(connection), ConjureError.class);
+        assertThat(error.errorName()).isEqualTo("Default:Conflict");
+        assertThat(error.errorCode()).isEqualTo("CONFLICT");
+        assertThat(error.parameters()).containsOnlyKeys("optionalInt", "arg", "optionalDouble", "optional");
```
mpritham, line 164: notably the keys `nullValue`, `emptyOptionalInt`, `emptyOptionalDouble`, and `emptyOptional` are not in the map as expected.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
+
+/**
+ * A representation of a Conjure defined error that will be serialized and sent over the wire from a server to a client.
+ * <p>Previously {@link SerializableError} was used for this purpose. The main difference is that
+ * {@link SerializableError#parameters()} is a {@code Map<String, String>} while {@link ConjureError#parameters} is a
+ * {@code Map<String, Object>}. When constructing a {@link SerializableError} from a {@link ServiceException}, the error
+ * parameters were converted to strings using {@link Objects#toString} which made it difficult for the client to recover
```
mpritham:
```suggestion
 * parameters were converted to strings using {@link Objects#toString} which made it difficult for clients to recover
```

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
mpritham: That was cleaner!

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
mpritham: Updated to deserialize into a `ConjureError` with `ObjectMappers.newClientJsonMapper()`.

build.gradle
```diff
@@ -21,13 +21,13 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.palantir.baseline:gradle-baseline-java:6.2.0'
+        classpath 'com.palantir.baseline:gradle-baseline-java:6.3.0'
```
mpritham: bad merge. I merged master into `pm/endpoint-errors` though. Going to try rebasing instead of merging.

PR #2410, opened 2024-11-18
Revert "Migrate squareup javapoet to Palantir javapoet (#2399)"

mpritham:
This reverts commit ac12ad0bdd229065c026ca7d6125e894e1791315.

## Before this PR
#2399 introduced an ABI break. To use the version of `conjure-java` with the Palantir fork of JavaPoet, we require all plugins that generate code using JavaPoet, and use conjure-java, to also migrate. We can automate the upgrade internally, but this will take some time. For now, let's continue using the Square version of JavaPoet until we migrate.

## After this PR

==COMMIT_MSG==
Revert "Migrate squareup javapoet to Palantir javapoet (#2399)"
==COMMIT_MSG==


carterkozak:
The readme is very clear about ABI guarantees:

> Note that all Java APIs in this repository (including conjure-java-core) are considered "internal" in the sense that they may change at any time and without prior warning or deprecation notice. These packages are published for convenience, but users must assume responsibility for adapting their code when dependencies change.

mpritham:
Closing. Library consumers will need to update their usages.

PR #2414, opened 2024-11-20
UnclosableOutputStream implements BufferWritableOutputStream

tom-s-powell:
Some clients make use of the knowledge a `BinaryResponseBody` was being given an `UndertowOutputStream` as there could be more efficient writes using `ByteBuffer`. These were not tightly coupled to this fact but I noticed this recently changed with https://github.com/palantir/conjure-java/pull/2352 so the more efficient code paths are no longer hit.

==COMMIT_MSG==
UnclosableOutputStream implements BufferWritableOutputStream.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2414.v2.yml
+++ b/changelog/@unreleased/pr-2414.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: UnclosableOutputStream implements BufferWritableOutputStream.
+  links:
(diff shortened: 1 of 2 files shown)


carterkozak:
Thoughts on a slight variation along these lines instead? https://github.com/palantir/conjure-java/pull/2415

I don't think we want to produce this interface if the underlying stream is unexpectedly changed not to implement BufferWritableOutputStream.

carterkozak:
Closing because #2415 has shipped

PR #2415, opened 2024-11-20
UnclosableOutputStream passes through BufferWritableOutputStream

carterkozak:
==COMMIT_MSG==
UnclosableOutputStream passes through BufferWritableOutputStream
==COMMIT_MSG==

This is an alternative implementation for #2414 such that the secondary interface is only produced when the underlying stream supports it.


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/UnclosableOutputStreams.java
```diff
@@ -0,0 +1,122 @@
...
+final class UnclosableOutputStreams {
+
+    /**
+     * Wraps the provided {@link OutputStream} such that {@link OutputStream#close()} is a no-op.
+     * This implementation ensures that streams like {@code UndertowOutputStream} are wrapped in a
+     * way that preserves the {@link BufferWritableOutputStream} interface.
+     */
+    static OutputStream wrap(OutputStream delegate) {
+        if (delegate instanceof BufferWritableOutputStream) {
+            return new UnclosableBufferWritableOutputStream(delegate, (BufferWritableOutputStream) delegate);
```
mpritham, line 37: no change suggested. Just slightly annoying that we have to do this because `OutputStream` is an abstract class and `BufferWritableOutputStream` is an interface. I initially expected `BufferWritableOutputStream` to be a more specific `OutputStream`, and we could make use of polymorphism to initialize `UnclosableBufferWritableOutputStream`. This is fine though. 🤷

mpritham (approved):
lgtm, thanks!

PR #2417, opened 2024-11-22
[Prework] Landing all 'conjure-lib' API changes Ahead of CodeGen Adoption

bmarcaur:
```
Includes:
  New conjure collections API for:
    - unmodifiableList
    - addAll Optimized add and serialize methods on native containers
```

Landing these ahead of their adoption in PRs like:
  - https://github.com/palantir/conjure-java/pull/2389
  - https://github.com/palantir/conjure-java/pull/2386

So that there are no concerns with upgrade ordering and accidental transitive drift.


conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
```diff
@@ -40,7 +40,7 @@ public long longValue() {
         return longValue;
     }
 
-    private static long check(long value) {
+    public static long check(long value) {
```
bmarcaur: Copied from https://github.com/palantir/conjure-java/pull/2389/files#r1811730849:
> I increased the visibility on this, because the logic is simple enough that to satisfy it is tantamount to copying the implementation detail (which felt bad). But also the logic is so simple that we aren't overly exposing some internals. Besides having more places rely on this is likely to prevent skew between implementations (IMO).

carterkozak: I'm not sure I agree that we should expose apis for `SafeLong` using the `long` primitive type

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -37,6 +39,35 @@ private ConjureCollections() {
...
+        if (list instanceof ConjureIntegerList) {
+            return (List<T>) ((ConjureIntegerList) list).asUnmodifiable();
+        } else if (list instanceof ConjureDoubleList) {
+            return (List<T>) ((ConjureDoubleList) list).asUnmodifiable();
+        } else if (list instanceof ConjureSafeLongList) {
+            return (List<T>) ((ConjureSafeLongList) list).asUnmodifiable();
+        } else {
+            // Otherwise use the JDK types
+            return Collections.unmodifiableList(list);
+        }
```
carterkozak: we don't need to bring along any of the implementations, they're all internal details -- `return Collections.unmodifiableList(list);` is sufficient

conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
bmarcaur: How would you suggest a non-boxing version of the check that happens during creation that also doesn't simply duplicate this logic?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
bmarcaur: Is this the only case you care about or is this just a comment on the first instance? For example, I added a number of methods to the `ConjureDoubleList` shaped types for the same reasons that we want the APIs to exist (also they are even more internal), but there isn't really a "pass-through" implementation equivalent.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
carterkozak: I don't think there's anything wrong with duplicating logic -- SafeLong provides two constants similarly to `Integer.MAX_VALUE` and `Integer.MIN_VALUE`.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
carterkozak: `ConjureDoubleList` is only created and checked from within `conjure-lib` code. Because it's kept insulated, we can make any changes we need without letting them soak first. In the case where code runs with a faulty classloader and inadvertently forces the conjure-lib version downward, they may not receive optimizations yet, but nothing would fail because the public API surface still exists.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
bmarcaur: You against making the primitive constants public? Cause right now I have to compare it to a `SafeLong`. Not impossible just more awkward.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
bmarcaur: I understand that this isn't a concern within a jar, but adding API's like https://github.com/palantir/conjure-java/pull/2417/files/bc1e6311d872a0e808367c373369fd0a3d9aff2f#diff-c7c23a70bf9b00504cc05e26d2bb08a161f4e036eadd36cd0e13efbbddb4e02cR198 is necessary for the reasons you laid out. Now I could have that just call https://github.com/palantir/conjure-java/pull/2417/files/bc1e6311d872a0e808367c373369fd0a3d9aff2f#diff-c7c23a70bf9b00504cc05e26d2bb08a161f4e036eadd36cd0e13efbbddb4e02cR202 but I am trying to tease out your goal. Is it to minimize the surface area of this change?

carterkozak: Yes, api surface is the piece that's most critical to get right, where buggy or inefficient logic can be fixed later with minimal regret. It's helpful to carefully and precisely adopt changes that expand the API without mixing the concerns with new logic in conjure any time we can get away with this separation.
Fundamentally these two types of changes are reviewed through different lenses, so my preference is to merge them separately.
Selfishly, it makes the code review process a lot easier, but also allows us to more safely revert entire PRs without breaking API if the need arises.

bmarcaur: Understood, will slim as much as possible

conjure-lib/src/main/java/com/palantir/conjure/java/lib/SafeLong.java
carterkozak: Hmm, I'd prefer not to make those public because we'd lose some symmetry with the standard boxed primitive types from java.

carterkozak: At a high level, I don't want to make changes to SafeLong lightly. My preference is not to expand in the process of implementing this feature -- if we find that there are many cases where enhancements are helpful, we should by all means make the improvements separately, then apply them to places like this which may benefit.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -181,9 +192,15 @@ public static List<Integer> newNonNullIntegerList(Iterable<Integer> iterable) {
         return integerList;
     }
 
+    // This method modifies a list that can't handle nulls. Do not use this unless the nonNullCollections flag is set
+    public static void addAllToIntegerList(Collection<Integer> addTo, int[] elementsToAdd) {
+        addAll(addTo, () -> Arrays.stream(elementsToAdd).iterator());
```
carterkozak: we could add an implementation dependency on guava for something like `addAll(addTo, Ints.asList(elementsToAdd));` -- streams can add quite a lot of overhead on their own. In practice this shouldn't really matter because we'll provide a special case implementation before this is used, but we'll need to maintain the fallback codepath either way.

bmarcaur: It doesn't need the optimization in that function so I will just inline the simple addition.

carterkozak (approved):

PR #2421, opened 2024-12-03
remove bounds check on input 'long' values into 'alias: double'

carterkozak:
==COMMIT_MSG==
remove bounds check on input 'long' values into 'alias: double'
==COMMIT_MSG==

Aliases are meant to behave identically to the aliased type, however in this case we were imposing a restriction that was _not_ present in the non-aliased path. As described in the original PR, this mans the bounding is different based on the _shape_ of the input data: https://github.com/palantir/conjure-java/pull/1006#discussion_r453888358 e.g. the restriction would apply to `9007199254740993.0` but not `9007199254740993`.

The PR which implemented this bounds check primarily aimed to avoid deserialization failures for large input values, matching behavior of raw `double`. Removing the safelong intermediate check should produce more consistent behavior.

### DIFF
--- a/changelog/@unreleased/pr-2421.v2.yml
+++ b/changelog/@unreleased/pr-2421.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: 'remove bounds check on input ''long'' values into ''alias: double'''
+  links:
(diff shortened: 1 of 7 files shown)


mpritham (approved):
Thanks!

PR #2423, opened 2024-12-11
Add flag to exclude generation of async Dialogue interfaces

mpritham:
## Before this PR
With Java 24's Loom feature, we will no longer need to use async interfaces. 

## After this PR
This PR introduces a flag `excludeDialogueAsyncInterfaces` that, when set, does not generate the async Dialogue interfaces. This will help with roll-out of the future Conjure version where async interfaces are no longer generated at all. If devs use async interfaces, and the flag is set, there should be a compile-break, so there is no risk of runtime errors with this change.

==COMMIT_MSG==
Add flag to exclude generation of async Dialogue interfaces
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2423.v2.yml
+++ b/changelog/@unreleased/pr-2423.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add flag to exclude generation of async Dialogue interfaces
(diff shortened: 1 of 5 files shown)


carterkozak (approved):
that was fast, thanks!

PR #2431, opened 2025-01-10
Optimized Jackson Deserialization for Optimized List of Primitives

bmarcaur:
## Before this PR
Comment that inspired this work: https://github.com/palantir/conjure-java/pull/2389#discussion_r1909090561

Previously, Jackson deserialization would pass through a generated setter leveraging an `Iterable<Type>` which inherently boxes the type to ultimately unbox while storing it in the optimized storage.

## After this PR
This changes simply moves the `JacksonSetter` to the non box'd setter for these optimized situations.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
@@ -601,11 +601,12 @@ public final class BeanBuilderGenerator {
             Map<com.palantir.conjure.spec.TypeName, TypeDefinition> typesMap,
             boolean override,
             boolean strict) {
-        Collection<MethodSpec> setters = Lists.newArrayListWithExpectedSize(fields.size());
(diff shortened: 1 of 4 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -601,11 +601,12 @@ private Iterable<MethodSpec> createSetters(
             Map<com.palantir.conjure.spec.TypeName, TypeDefinition> typesMap,
             boolean override,
             boolean strict) {
-        Collection<MethodSpec> setters = Lists.newArrayListWithExpectedSize(fields.size());
+        // Each field tends to fan out into many setters
```
carterkozak, line 604: This is only true of container types (optional/set/list), not standard fields. Difficult to tell whether a 5x multiplier is ideal or not, but probably not worth spending much time optimizing

carterkozak (approved):
Nice!

PR #2435, opened 2025-01-15
Generate error result types in Dialogue interfaces for endpoints with associated errors

mpritham:
## Before this PR
In #2401, we added the ability to associate errors with endpoints so servers could throw checked exceptions (extensions of [`CheckedServiceException`](https://github.com/palantir/conjure-java-runtime-api/blob/4cdb62d4c0a1eb680d67a8a1adbca7df50673ffd/errors/src/main/java/com/palantir/conjure/java/api/errors/CheckedServiceException.java#L27)). Clients still catch `RemoteException`s.

## After this PR
Dialogue client interface methods for endpoints which have associated errors will return result types - implemented as sealed interfaces permitting a successful result member, and a member for each error. This enforces that clients handle all errors an endpoint can throw at compile time.

[Examples of generated code are given here](https://github.com/palantir/conjure-java/blob/63b497937397a7412301f958808a3e8dcc94b8ba/conjure-java-core/src/integrationInput/java/dialogueendpointresulttypes/com/palantir/product/ErrorServiceBlocking.java)

This feature should be used with Java 21 to make use of the [exhaustive pattern matching switch](https://openjdk.org/jeps/441#:~:text=Exhaustiveness%20of%20switch%20expressions%20and%20statements,-Type%20coverage) feature.

==COMMIT_MSG==
Generate error result types in Dialogue interfaces for endpoints with associated errors
==COMMIT_MSG==

## Possible downsides?

If not rolled out judiciously, this feature may result in compile breaks. Concretely, if a service enables this feature and publishes JARs containing the generated Dialogue clients, consumers of the JARs will be forced to resolve the compile breaks they encounter when using the Dialogue clients.


conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -100,6 +101,8 @@ public final class UndertowServiceEteTest extends TestBase {
     private static Undertow server;
 
     private final EteServiceBlocking client;
+    // TODO(pm): this should be in a different test.
```
mpritham: Noting for myself to add more tests here, then move them to their own file.

conjure-java-core/src/integrationInput/java/com/palantir/product/EndpointSpecificServerErrors.java
```diff
@@ -40,7 +41,7 @@ public static void throwIfEndpointError(boolean shouldThrow, @Safe String typeNa
     public static final class EndpointError extends CheckedServiceException {
         private EndpointError(@Safe String typeName, @Unsafe Object typeDef, @Nullable Throwable cause) {
             super(
-                    EndpointSpecificErrors.ENDPOINT_ERROR,
+                    EndpointSpecificErrors.DIFFERENT_PACKAGE,
```
mpritham, line 44: This shouldn't have changed.

conjure-java-core/src/integrationInput/java/com/palantir/product/ErrorServiceAsync.java
```diff
@@ -0,0 +1,386 @@
...
+            }
+        }
+    }
+
+    sealed interface TestBinaryResponse permits TestBinaryResponse.Success, TestBinaryResponse.InvalidArgument {
+        record Success(@MustBeClosed InputStream value) implements TestBinaryResponse {
+            public Success {
+                Preconditions.checkArgumentNotNull(value, "value cannot be null");
+            }
+        }
```
carterkozak: When the success type is an `InputStream` or `Optional<InputStream>`, I think the base type should implement `Closeable` so that callers can use a try-with-resources block on the call without risking resource leaks.

We can bound the codegen by generating a no-op close method on the base type, then we don't need to repeat it in every error type, and can implement the real closure logic in the `Success` variant.
```java
@Override
default void close() throws IOException {}
```

conjure-java-core/src/integrationInput/java/com/palantir/product/ErrorServiceAsync.java
```diff
@@ -0,0 +1,386 @@
...
+
+        final class InvalidArgument extends ConjureErrors.BaseEndpointError<TestErrors.InvalidArgumentParameters>
+                implements TestBinaryResponse {
+            @JsonCreator
```
carterkozak: Let's make the annotation as specific as possible, which reduces the risk of jackson changes either breaking deserialization, or inadvertently causing us to successfully deserialize shapes we don't expect:
```suggestion
            @JsonCreator(Mode.PROPERTIES)
```

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureErrors.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.palantir.logsafe.Safe;
+
+public final class ConjureErrors {
+    private ConjureErrors() {}
+
+    // This is used in generated code when generating errors which don't have any parameters.
+    public static final EmptyErrorParameters EMPTY_ERROR_PARAMETERS_INSTANCE = new EmptyErrorParameters();
+
+    public record EmptyErrorParameters() {}
```
carterkozak: I think I'd prefer to generate empty records for each error that doesn't define parameters, that way they're consistent.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureErrors.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public abstract static class BaseEndpointError<T> {
```
carterkozak: This should live in the `dialogue-target` module within dialogue, ideally as a top-level class to keep the code a bit more concise

conjure-java-core/src/integrationInput/java/com/palantir/product/ErrorServiceBlocking.java
```diff
@@ -0,0 +1,386 @@
...
+        record Success(String value) implements TestBasicErrorResponse {
+            public Success {
+                Preconditions.checkArgumentNotNull(value, "value cannot be null");
+            }
+        }
+
+        final class InvalidArgument extends ConjureErrors.BaseEndpointError<TestErrors.InvalidArgumentParameters>
+                implements TestBasicErrorResponse {
+            @JsonCreator
+            InvalidArgument(
```
carterkozak: let's generate log-safety annotations for these based on the safety of the *Parameters type

conjure-java-core/src/integrationInput/java/com/palantir/product/ErrorServiceBlocking.java
```diff
@@ -0,0 +1,386 @@
...
+        }
+    }
+
+    sealed interface TestEmptyBodyResponse
+            permits TestEmptyBodyResponse.Success, TestEmptyBodyResponse.InvalidArgument {
+        record Success() implements TestEmptyBodyResponse {
+            @JsonCreator
+            public static Success create() {
+                return new Success();
+            }
```
carterkozak: The other success variants don't appear to be annotated -- is this method necessary?

carterkozak (commented):
Let's add some end-to-end tests using a conjure-undertow service producing failures, and a dialogue client handling them to make sure the pieces fit together in practice!

conjure-java-core/src/integrationInput/java/com/palantir/product/ErrorServiceBlocking.java
mpritham: 🤦 of course, thanks.

conjure-java-core/src/integrationInput/java/com/palantir/product/ErrorServiceBlocking.java
mpritham: We only need it when deserializing empty body responses so that the reflection code in Dialogue's `JacksonEmptyContainerLoader` [(src)](https://github.com/palantir/dialogue/pull/2443/files#diff-8d97a05b5c1cad8e5e6afe3d1d1b8e6a3fbc5f106fd7417f6282244184422142) can invoke this method to construct an empty instance.

carterkozak: Could we teach it how to handle an empty record instead?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureErrors.java
mpritham: I initially had that. I traded consistency for slightly lower amount of generated code and to omit the `parameters` field in the type that Jackson extracts from the request body. I _think_ this saves a bit of time during request processing because Jackson isn't spending time attempting to extract empty params. I don't have a great sense of how much time this actually adds.

Happy to switch back - also removes the need for the `ErrorGenerationUtils.ErrorNameToParameterExistenceMapping` thing.

mpritham: hm, I can generate this record for parameters and still omit the "parameters" field from the type used by Jackson deserialization.

conjure-java-core/src/integrationInput/java/com/palantir/product/ErrorServiceBlocking.java
mpritham: yeah that works too. Avoids us generating these ctors in the records.

conjure-java-core/src/integrationInput/java/com/palantir/product/ErrorServiceBlocking.java
```diff
@@ -0,0 +1,433 @@
...
+    final class Factory implements DialogueServiceFactory<ErrorServiceBlocking> {
+        @Override
+        public ErrorServiceBlocking create(EndpointChannelFactory endpointChannelFactory, ConjureRuntime runtime) {
+            return ErrorServiceBlocking.of(endpointChannelFactory, runtime);
+        }
+    }
+
+    sealed interface TestBasicErrorResponse
+            permits TestBasicErrorResponse.Success, TestBasicErrorResponse.InvalidArgument {
+        record Success(@JsonValue String value) implements TestBasicErrorResponse {
```
mpritham: Another change since last review: I added the @JsonValue. This is needed to deserialize the records (similar to `@JsonCreator(mode = Mode.DELEGATING)`): [this is a related issue.](https://github.com/FasterXML/jackson-databind/issues/3180#issuecomment-1306160265)

conjure-java-core/src/integrationInput/java/com/palantir/product/ErrorServiceBlocking.java
mpritham: Updated the [dialogue PR here](https://github.com/palantir/dialogue/pull/2443/files#diff-8d97a05b5c1cad8e5e6afe3d1d1b8e6a3fbc5f106fd7417f6282244184422142R169) to invoke the 0-parameter record constructor.

carterkozak (dismissed):
approving for a release candidate!

carterkozak:
we'll need to rebase this on top of the test refactors

conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
```diff
@@ -187,6 +187,16 @@ default boolean preferObjectBuilders() {
         return false;
     }
 
+    /**
+     * If enabled, endpoints that have associated errors will return a result type: a sealed interface permitting
+     * subclasses for the endpoint's return value, and each endpoint error. Each endpoint error is a subclass of
+     * {@link com.palantir.dialogue.EndpointError}.
+     */
```
bjlaub, line 199: i think we should add some language (perhaps here and in the readme) stressing that:
- this is an experimental feature at the moment
- enabling this flag will alter generated dialogue clients in a way that will cause compilation breaks for consumers
- enabling this flag in combination with conjure-java-local on the consumer side in library code may result in ABI breaks for consumers of that library that _also_ consume generated conjure types from the same API definition.

That last bit is more a statement about how conjure-java-local might be expected to affect consumers, so perhaps its irrelevant here, but worth noting somewhere.

Discussed offline: we might want to put together a separate document/wiki of some kind describing the expected way of rolling out enablement of this flag in concert with new error type definitions on an API. Basically a runbook that also describes potential pitfalls so there are no surprises.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueServiceGenerator.java
```diff
@@ -96,18 +100,21 @@ public Stream<JavaFile> generate(ConjureDefinition conjureDefinition) {
...
     private static Stream<JavaFile> generateFilesForService(
             boolean excludeDialogueAsyncInterfaces,
+            boolean generateDialogueEndpointErrorResultTypes,
             ServiceDefinition serviceDef,
             DialogueEndpointsGenerator endpointsGenerator,
             DialogueInterfaceGenerator interfaceGenerator,
             StaticFactoryMethodGenerator blockingGenerator,
             StaticFactoryMethodGenerator asyncGenerator) {
-        List<JavaFile> files = new ArrayList<>(/* initialCapacity= */ 3);
+        List<JavaFile> files = new ArrayList<>(/* initialCapacity= */ 4);
```
bjlaub: why did this need to change?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/ReturnTypeMapper.java
```diff
@@ -44,4 +46,16 @@ public TypeName baseType(Optional<Type> type) {
     public TypeName async(Optional<Type> type) {
         return ParameterizedTypeName.get(LISTENABLE_FUTURE, Primitives.box(baseType(type)));
     }
+
+    public boolean isBinaryOrOptionalBinary(TypeName returnType) {
```
bjlaub: do these need to be public?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -159,41 +174,74 @@ private Optional<FieldSpec> serializer(EndpointName endpointName, Type type) {
                 .build());
     }
 
-    private Optional<FieldSpec> deserializer(EndpointName endpointName, Optional<Type> type) {
+    private Optional<FieldSpec> deserializer(
+            TypeName responseType, EndpointDefinition endpointDef, EndpointName endpointName, Optional<Type> type) {
```
bjlaub: nit: no need for `endpointName` and `type` if we are passing the full `EndpointDefinition` here.

```suggestion
    private Optional<FieldSpec> deserializer(TypeName responseType, EndpointDefinition endpointDef) {
            EndpointName endpointName = endpointDef.getEndpointName();
            Optional<Type> type = endpointDef.getReturns();
```

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueServiceGenerator.java
mpritham: it doesn't need to change. this was a remnant of a previous iteration where I created a separate file for the result types.

will revert, thanks!

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/ReturnTypeMapper.java
mpritham: They don't. Neither do the other methods in this class. Can make a separate PR for that.

mpritham: Opened #2472

bjlaub (dismissed):
lgtm. probably worth a changelog

bjlaub (approved):

PR #2440, opened 2025-01-22
[Improvement] Adopt RevAPI

bmarcaur:
## Before this PR
I was able to accidentally remove part of the public ABI as part of https://github.com/palantir/conjure-java/pull/2389. I fixed it in https://github.com/palantir/conjure-java/pull/2434.

## After this PR
Accidental breaks like this would produce an error requiring closer scrutiny

## Possible downsides?
Breaks != unacceptable change, but may incorrectly communicate that intention. Also might be overly applied.

### DIFF
--- a/conjure-java-core/build.gradle
+++ b/conjure-java-core/build.gradle
@@ -15,6 +15,7 @@
  */
 
 apply plugin: 'com.palantir.external-publish-jar'
+apply plugin: 'org.revapi.revapi-gradle-plugin'
 
 sourceSets {
     integrationInput
--- a/conjure-java-core/build.gradle
+++ b/conjure-java-core/build.gradle
@@ -15,7 +15,6 @@
  */
 
 apply plugin: 'com.palantir.external-publish-jar'
-apply plugin: 'org.revapi.revapi-gradle-plugin'
 
 sourceSets {
     integrationInput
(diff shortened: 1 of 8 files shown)


conjure-java-core/build.gradle
```diff
@@ -15,6 +15,7 @@
  */
 
 apply plugin: 'com.palantir.external-publish-jar'
+apply plugin: 'org.revapi.revapi-gradle-plugin'
```
carterkozak: This project has no api guarantees, and is documented as such in the readme

conjure-java/build.gradle
```diff
@@ -15,6 +15,7 @@
  */
 
 apply plugin: 'com.palantir.external-publish-application-dist'
+apply plugin: 'org.revapi.revapi-gradle-plugin'
```
carterkozak: No api guarantees beyond rfc-002

conjure-undertow-processor/build.gradle
```diff
@@ -1,4 +1,5 @@
 apply plugin: 'com.palantir.external-publish-jar'
+apply plugin: 'org.revapi.revapi-gradle-plugin'
```
carterkozak: This is an annotation processor, the annotations themselves must be compatible, but not the processor implementation

carterkozak (approved):
thanks!

PR #2445, opened 2025-01-29
Remove our workaround for an old Javapoet class init deadlock

carterkozak:
https://github.com/square/javapoet/issues/637 impacted the original Square builds of Javapoet, however our fork fixed the bug in https://github.com/palantir/javapoet/pull/8

Relevant tech blog post from a few years ago here: https://blog.palantir.com/using-static-analysis-to-prevent-java-class-initialization-deadlocks-c2f31ca967d6

==COMMIT_MSG==
Remove our workaround for an old Javapoet class init deadlock
==COMMIT_MSG==


mpritham (approved):
Thanks!

Issue #1095, opened 2020-09-25
nonNullCollections flag isn't being respected for alias types


a-k-g:
## What happened?

We have the following conjure type:

      AndSearchFilter:
        docs: |
          A SearchFilter used to combine multiple SearchFilters.
          A row satisfies this SearchFilter if and only if it satisfies all combined SearchFilters.
        alias: set<SearchFilter>

And we have enabled the `nonNullCollections` flag to enforce that the set cannot contain `null` entries. However, this is not being enforced.

## What did you want to happen?

Constructing the `AndSearchFilter` with a set containing `null` should fail.

mpritham:
When the `nonNullCollections` flag is set, we could update the constructor to create a copy of the list passed in (for lists, via a call to [this method in ConjureCollections](https://github.com/palantir/conjure-java/blob/06973e464629b007f610621efd78ff89031434f4/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java#L144-L146)). 

The generation of this code should be under a new flag (maybe `--copyCollectionsOnConstruction` or something): it's currently possible that the aliased collection is mutated in _some_ code that uses Conjure. Specifically, something like this could exist:
```java
ListAlias createListAlias() {
  List<String> arr = new ArrayList<>();
  ListAlias listAlias = ListAlias.of(arr);
  callThatPopulatesList(arr);
  return listAlias;
}
```
We don't want any new changes we introduce to `conjure-java` to automatically lead to failures for clients that have the above (discouraged) pattern.

Conjure objects are meant to be immutable. As part of this change, when the `copyCollectionsOnConstruction` flag is set and the `nonNullCollections` flag is not set, we could create copies of collections (that allow null values) in the constructors for aliased collection types.

carterkozak:
bikeshed: The feature flag should be a bit more specific since we do already copy collections in most cases (conjure bean types, just not aliases!).  Perhaps something along the lines of `--defensiveCollectionAliases`.

> There are only two hard things in Computer Science: cache invalidation and naming things.

PR #2448, opened 2025-02-04
Support collection copying in aliases using `defensiveCollectionAliases`

kunalrkak:
## Before this PR
#1095

## After this PR

==COMMIT_MSG==
Makes immutable copies of collections for union and alias types if the defensiveCollections flag is enabled.
==COMMIT_MSG==


conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -244,6 +255,79 @@ private static boolean isAliasOfDouble(AliasDefinition typeDef) {
...
+        @Override
+        public Optional<CodeBlock> visitOptional(OptionalType _value) {
+            return Optional.empty();
+        }
+
+        @Override
+        public Optional<CodeBlock> visitList(ListType value) {
+            if (options.nonNullCollections()) {
+                return Optional.of(CodeBlock.of(
+                        "return new $T($T.newNonNullList($L))",
```
carterkozak: Can we specialize this taking advantage of `newNonNullIntegerList`/`newNonNullDoubleList`/`newNonNullSafeLongList` where possible? That can reduce the overhead of data on heap by avoiding boxed types.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -228,6 +231,14 @@ public static JavaFile generateAliasType(
                 .build();
     }
 
+    private static CodeBlock getStaticConstructor(Type type, ClassName thisClass, Options options) {
+        if (options.defensiveCollectionAliases() && type.accept(MoreVisitors.IS_COLLECTION)) {
+            return type.accept(new DefensiveCollectionConstructorVisitor(thisClass, options))
+                    .orElseThrow();
+        }
+        return CodeBlock.of("return new $T(value)", thisClass);
```
carterkozak: Thoughts on moving the options check into `DefensiveCollectionConstructorVisitor` where we implicitly handle the `MoreVisitors.IS_COLLECTION` check via the visitor? We'd probably want to rename `DefensiveCollectionConstructorVisitor` to `AliasFactoryVisotor implements Visitor<CodeBlock>`

kunalrkak:
@carterkozak refactored this as per your comments, should be ready for another look!

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -1074,61 +1076,6 @@ public CollectionType visitSet(SetType _value) {
         });
     }
 
-    private static final class CollectionType {
```
kunalrkak, line 1077: Pulled this out to a non-internal class so I could reuse it in `AliasGenerator`

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -244,6 +250,145 @@ private static boolean isAliasOfDouble(AliasDefinition typeDef) {
...
+        }
+
+        @Override
+        public CodeBlock visitMap(MapType value) {
+            if (!options.defensiveCollectionAliases()) {
+                return defaultFactoryStatement();
+            }
+
+            return CodeBlock.of(
+                    "return new $T(new $T<>($L))", thisClass, LinkedHashMap.class, nonNullValueExpression());
```
kunalrkak: Along with the annotation added to the parameter, kept the List copy here for object immutability -- let me know if there's a different type preferred though, since we don't have anything baked into `ConjureCollections` by default for Maps

carterkozak (commented):
Could you add a test for deserialization of an `alias<map<string, optional<string>>>` from `{"key": null}`?

kunalrkak:
thanks for flagging this map/optional issue, right now this doesn’t deserialize as expected (fails) — I think I’m missing logic like [this](https://github.com/palantir/conjure-java/blob/83990b0fad03e2f6be5d5161eecc8f776976cf98/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java#L664-L668) when creating the annotation to support empty optionals in map values, going to try to pull this out to somewhere shared

carterkozak (commented):
As we add more feature flags, our existing tests become less effective. I'm not sure that I'd want to test all permutations of flags, but I do think we need to put some thought into the best way to maintain tests across configurations.

The set of flags used in our auto-verifier tests is quite stale, and may not be representative of most fielded products at this point (which themselves span a wide range of sets of feature flags enabled).

The tests for defensive aliases look great, but it's unfortunate that don't get any verification for this change from our existing test coverage, where only types are used without this feature. I'm not sure exactly how I'd like to make this better quite yet, but I'm open to suggestions if you have ideas :-]

conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
```diff
@@ -187,6 +187,11 @@ default boolean preferObjectBuilders() {
         return false;
     }
 
+    @Value.Default
```
mpritham, line 208: nit: let's paste the docstring you've added in the readme?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -259,6 +250,55 @@ private static boolean isAliasOfDouble(AliasDefinition typeDef) {
                 && typeDef.getAlias().accept(TypeVisitor.PRIMITIVE).equals(PrimitiveType.DOUBLE);
     }
 
+    private static ParameterSpec getAliasFactoryParameter(
+            Type aliasType, TypeName aliasTypeName, TypeMapper typeMapper, Options options) {
+        ParameterSpec parameterSpec = Parameters.nonnullParameter(aliasTypeName, "value");
+
+        if (aliasType.accept(TypeVisitor.IS_SET)) {
+            AnnotationSpec deserializeAnnotation = AnnotationSpec.builder(JsonDeserialize.class)
```
mpritham, line 258: nit: might be worth calling out this change in the commit message? 

I think we've already settled on this but writing it down - Using `LinkedHashSet` by default now is fine: previously we didn't guarantee an ordering and now we do, so there's no wire break. According to the LinkedHashSet docstring, there is a slight performance degradation compared to HashSet from maintaining a doubly-linked list, but the benefits of providing ordering outweigh the perf cost.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -259,6 +250,55 @@ private static boolean isAliasOfDouble(AliasDefinition typeDef) {
...
+                    .addMember("as", "$T.class", LinkedHashSet.class)
+                    .build();
+            return parameterSpec.toBuilder()
+                    .addAnnotation(deserializeAnnotation)
+                    .build();
+        }
+
+        if (options.defensiveCollections() && options.nonNullCollections() && aliasType.accept(TypeVisitor.IS_MAP)) {
```
mpritham, line 266: I wonder if we need the `defensiveCollections` flag to use the `Nulls` annotations? 

Maybe we should do this when the `nonNullCollections` flag is set (irrespective of `defensiveCollections`). Reading the docs for `nonNullCollections`: "generate objects that fail to deserialize collections with null values."

kunalrkak: You're right, but I guess my thought here was that only gating by `nonNullCollections` could break existing clients that have the feature flag enabled, even if it _should_ have been disallowing nulls in this case. So figured it would be best to introduce this change behind the defensiveCollections FF, and would suggest we enable this by default in the templates as well. But happy to drop the `defensiveCollections` check both here and in `UnionGenerator` if you think that it's more of a "fix" to the existing flag!

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -232,6 +232,13 @@ public static final class GenerateCommand implements Runnable {
                         + "that for objects without any fields, this will still generate the static factory method.")
         private boolean preferObjectBuilders;
 
+        @CommandLine.Option(
+                names = "--defensiveCollections",
+                defaultValue = "false",
+                description = " Makes immutable copies of collections for union and alias types, respecting the "
```
mpritham:
```suggestion
                description = "Makes immutable copies of collections for union and alias types, respecting the "
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
mpritham: hm, I suppose this can break existing clients if we have a null member in a list that we've created while the `nonNullCollections` flag is set. I'm not sure we'd actually run into this case, but best to be safe. Rolling this out with the `defensiveCollections` flag would be safer as you suggested.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/Expressions.java
```diff
@@ -44,6 +48,18 @@ public static CodeBlock objectArray(Collection<?> args) {
...
 
+    public static CodeBlock wrapUnmodifiableCollections(Type type, String fieldName) {
+        if (type.accept(TypeVisitor.IS_LIST)) {
+            return CodeBlock.of("$T.unmodifiableList($L)", ConjureCollections.class, fieldName);
+        } else if (type.accept(TypeVisitor.IS_SET)) {
+            return CodeBlock.of("$T.unmodifiableSet($L)", Collections.class, fieldName);
+        } else if (type.accept(TypeVisitor.IS_MAP)) {
+            return CodeBlock.of("$T.unmodifiableMap($L)", Collections.class, fieldName);
+        } else {
+            return CodeBlock.of("$L", fieldName);
```
mpritham: sorry didn't mention earlier: Hitting this case is unexpected because we check `IS_COLLECTION` before calling this. Thoughts on just throwing here?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -806,12 +808,21 @@ private static Iterable<AnnotationSpec> deserializationAnnotationForSets(FieldDe
...
         AnnotationSpec.Builder builder = AnnotationSpec.builder(JsonSetter.class)
                 .addMember("value", "$S", field.getFieldName().get());
         Type dealiased = TypeFunctions.toConjureTypeWithoutAliases(field.getType(), typesMap);
         if (dealiased.accept(DefaultableTypeVisitor.INSTANCE)) {
             builder.addMember("nulls", "$T.AS_EMPTY", Nulls.class);
+            // We only need to restrict nulls by annotations on Maps, the other collections have non-null constructors
+            if (options.defensiveCollections()
+                    && options.nonNullCollections()
+                    && field.getType().accept(TypeVisitor.IS_MAP)) {
+                // TODO(kkak): TypeMapper isn't needed since we've de-aliased
```
mpritham: nit: let's remove

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -806,12 +808,21 @@ private static Iterable<AnnotationSpec> deserializationAnnotationForSets(FieldDe
...
+                if (!TypeFunctions.isOptionalInnerType(dealiased, new TypeMapper(typesMap, options))) {
```
mpritham: [The wire spec says](https://palantir.github.io/conjure/#/docs/spec/wire?id=_561-coercing-json-null-absent-to-conjure-types):
>  If a JSON key is absent or the value is null, two rules apply:
>   1. Conjure optional, list, set, map types must be initialized to their empty variants,
>   2. Attempting to coerce null/absent to any other Conjure type must cause an error, i.e. missing JSON keys must cause an error.

I think we shouldn't be adding this annotation for any of `optional`, `list`, `set` or `map` here?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/Expressions.java
kunalrkak: We won't hit this case with the current usage, but this is in a generic `Expressions` class I feel like returning a default is safer than throwing -- maybe we can make this a private method in the calling class and then throw at the end?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
kunalrkak: Will remove the comment, I wonder if its worth making an `isOptionalInnerType` method without the type mapper since `TypeFunctions.toConjureTypeWithoutAliases` already recursively strips the aliases (at least seems to, will confirm)

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
kunalrkak: I think the wire spec reference you linked is related to this existing snippet: https://github.com/palantir/conjure-java/blob/014aa9a668bf0f7f583b378d864b930320d57647/conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java#L816
i.e. any "defaultable" value type is initialized to their empty variant if attempting to deserialize from `null`.

The annotation we're adding (for maps specifically) is to disallow content nulls, i.e. `{"fieldName": {"key": null}}`, which should not be supported when the nonNullCollections feature flag is enabled. Using the content-nulls annotation seems to be the best way to do this for maps -- for lists and sets we iterate through the elements to validate this:

https://github.com/palantir/conjure-java/blob/614ff6c394cd8b92ddfa920446e9c2622f485690/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java#L163-L170

mpritham: ah I see. This also fixes the divergence between the BeanGenerator and the UnionGenerator: we add this annotation when generating code with the BeanGenerator: https://github.com/palantir/conjure-java/blob/614ff6c394cd8b92ddfa920446e9c2622f485690/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java#L667

conjure-java-core/src/main/java/com/palantir/conjure/java/types/Expressions.java
mpritham: I don't think we ever expect to call this method when `type` is not one of list, set, map? By throwing when we do, we could catch the mistake during test code-gen.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
mpritham: Can we pass in the `typeMapper` that was given to `generateWrapperClasses`? This would avoid the extra object creation here.

If the type has been de-aliased, then the `reference` method will just never be visited which is fine. If it hasn't been de-aliased, I think it's correct that we de-alias it and check the type.

The benefit of creating a new method that doesn't take `TypeMapper` is that we'd catch errors in the de-aliasing logic above. I don't have a strong opinion here. Happy to keep the current method with `TypeMapper`

conjure-java-core/src/main/java/com/palantir/conjure/java/types/Expressions.java
kunalrkak: Yeah that's fair, can throw if not one of the collection types 👍🏼

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
kunalrkak: Ah good catch! Yes we can pass in the existing type mapper, I don't think the new method is necessary then -- we just won't visit the `reference` part of the visitor but that's okay

mpritham (approved):
Thanks!

PR #2449, opened 2025-02-06
Handle inner optional checking recursively for sets

kunalrkak:
## Before this PR

Sets only inspected the first layer of inner types for optionals

## After this PR

==COMMIT_MSG==
Sets recursively handle inspecting inner types.
==COMMIT_MSG==

## Possible downsides?

Bean generation would change for classes of type `set<set<optional<*>>>`, becoming less restrictive (current behavior would be to fail on nulls -> interpret these as empty optionals).

### DIFF
--- a/changelog/@unreleased/pr-2449.v2.yml
+++ b/changelog/@unreleased/pr-2449.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Sets recursively handle inspecting inner types.
+  links:
+  - https://github.com/palantir/conjure-java/pull/2449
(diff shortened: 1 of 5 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/util/TypeFunctions.java
```diff
@@ -186,7 +186,7 @@ public Boolean visitList(ListType value) {
 
             @Override
             public Boolean visitSet(SetType value) {
-                return value.getItemType().accept(TypeVisitor.IS_OPTIONAL);
+                return isOptionalInnerType(value.getItemType(), typeMapper);
```
carterkozak: Can we add an example type where codegen is impacted by this change? I like to set this sort of thing up with an initial commit adding the type, then the commit which makes the logic change in order to show the diff

kunalrkak: Added!

carterkozak (approved):
thanks!

PR #2450, opened 2025-02-10
[idea - do not merge] better snapshot testing with a JUnit extension

mpritham:
## Before this PR
Currently we test the output of various code [Generators](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/main/java/com/palantir/conjure/java/Generator.java) by checking in the expected source files in an [integrationInput](https://github.com/palantir/conjure-java/tree/develop/conjure-java-core/src/integrationInput) directory, and [comparing the output of the generated code to files in that directory](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceGeneratorTests.java#L81). When there is an update in the generated code, we re-create the reference files by setting an environment variable (`recreate`) before running tests.

Problem: we store all of the reference files in one directory. This makes it tedious to test multiple configurations which generate the same files. 
- For example, when testing an update that omits static factories, we checked in [a new conjure definition](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/test/resources/example-types-no-static-factory.yml) with new types, when we could have used [the existing file](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/test/resources/example-types.yml) and set [the relevant flag when testing code gen](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java#L186)

## After this PR
We can remove some of the tedium with our snapshot testing setup by having a JUnit extension (and maybe Gradle plugin - see downsides section) handle the generation and update of the directories and files associated with each test.

Annotating a parameter of a JUnit test (with `@SnapshotTest` in this case) will trigger the JUnit extension which will construct a `SnapshotTest` object with required test context (like test name) used for getting the directory of the reference files.

## Possible downsides?
- No nice tooling to update expected results (yet). Snapshot testing libraries in other languages have tools which provide a nice dev-experience when updating tests. See [jest (ts)](https://jestjs.io/docs/snapshot-testing#interactive-snapshot-mode), [insta (rust)](https://insta.rs/docs/quickstart/#:~:text=(words)%3B%0A%7D-,Reviewing%20Snapshots,-The%20recommended%20flow) or [ppx_expect (ocaml)](https://blog.janestreet.com/the-joy-of-expect-tests/)
  - it's possible to build a gradle plugin to have a similar dev-ex.
  - I found [this library](https://github.com/origin-energy/java-snapshot-testing) in Java, but I only see the `update-snapshot` flag which acts like our `recreate`
- Not a huge drawback. Lots of sources files checked into the repository: we'll have to check in the full set of files generated by each configuration we test.

### DIFF
--- a/conjure-java-core-tests/build.gradle
+++ b/conjure-java-core-tests/build.gradle
@@ -1,0 +1,110 @@
+buildscript {
+    repositories {
+        mavenCentral() { metadataSources { mavenPom(); ignoreGradleMetadataRedirection() } }
+    }
+    dependencies {
+        classpath 'org.yaml:snakeyaml:1.23'
+        classpath 'com.palantir.conjure:conjure-api-objects:4.50.0'
+        classpath 'com.palantir.conjure:conjure-core:4.50.0'
+        classpath 'com.palantir.javapoet:javapoet:0.5.0'
+        classpath 'com.atlassian.commonmark:commonmark:0.12.1'
+        classpath 'org.yaml:snakeyaml:2.3'
+        classpath files("${project(':conjure-java-core').buildDir}/classes/java/main")
+    }
+}
+
+plugins {
+    id 'java'
+}
+
+sourceSets {
+    tests {
+        java {
+            srcDir 'src/tests'
+        }
+    }
+    myExample {
+        java {
+            srcDirs 'src/myExampleTest**'
+        }
+    }
+
+    example {
+        java {
+            srcDirs.each { dir->
+                srcDir dir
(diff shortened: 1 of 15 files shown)


mpritham:
Closing this in favor of a solution that dynamically creates subprojects containing the generated files for each configuration. This will allow us to ensure the generated code is compiled, and that we can write unit tests against it.

PR #2454, opened 2025-02-12
handle DeadlineExpiredException in standard exception mapper

bjlaub:
## Before this PR
[DeadlineExpiredException](https://github.com/palantir/deadlines-java/pull/25) would always map to a ServiceException no matter what.

## After this PR
Add special handling for `DeadlineExpiredException.External` and `DeadlineExpiredException.Internal`, mapping them to HTTP 400 and 500 status codes in the response, respectively.

Note that deadlines-java does not actually throw this exception yet, but having this in place on the standard conjure runtime exception mapper will set us up to correctly enable enforcement in the future.

==COMMIT_MSG==
handle DeadlineExpiredException in standard exception mapper
==COMMIT_MSG==


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -183,6 +186,20 @@ private static void frameworkException(HttpServerExchange exchange, FrameworkExc
...
+        // need to further narrow down to external versus internal exception types
+        if (exception instanceof DeadlineExpiredException.External externalException) {
+            log.info("external deadline exceeded", externalException);
+            writeResponse(exchange, Optional.empty(), 400);
+        } else if (exception instanceof DeadlineExpiredException.Internal internalException) {
+            log.error("internal deadline exceeded", internalException);
+            writeResponse(exchange, Optional.empty(), 500);
+        } else {
+            log.error("unknown deadline exception type", exception);
+            writeResponse(exchange, Optional.empty(), 500);
```
bjlaub: perhaps it would be better to fall back to the `ServiceException` case instead here?

carterkozak: Ya, the full response with errorInstanceId would be helpful

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -183,6 +187,20 @@ private static void frameworkException(HttpServerExchange exchange, FrameworkExc
         writeResponse(exchange, Optional.of(ConjureError.fromServiceException(exception)), statusCode);
     }
 
+    private static void deadlineExpiredException(HttpServerExchange exchange, DeadlineExpiredException exception) {
+        // need to further narrow down to external versus internal exception types
+        if (exception instanceof DeadlineExpiredException.External externalException) {
+            log.info("external deadline exceeded", externalException);
+            writeResponse(exchange, Optional.empty(), 400);
```
carterkozak: How does the calling client know that this was the result of a provided deadline, and not some other 400?

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -183,6 +187,20 @@ private static void frameworkException(HttpServerExchange exchange, FrameworkExc
...
+        } else if (exception instanceof DeadlineExpiredException.Internal internalException) {
+            log.error("internal deadline exceeded", internalException);
+            writeResponse(exchange, Optional.empty(), 500);
```
carterkozak: I think it's probably less important for callers to distinguish between whether an internal deadline has been exceeded vs other types of 500, but if we can convey it, we may be able to make the system a bit easier to observe.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
bjlaub: ah, good point - i guess we don't make a distinction currently :(

we could tack on a header, or could consider returning something like [HTTP 417](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/417) though I'm wary of re-using that for this purpose.

Perhaps we should add a response header encoder for these exception types similar to `QosReasons.encodeToResponse`?

EDIT: and place that in deadlines-java, not here

bjlaub: > EDIT: and place that in deadlines-java, not here

or maybe it's better to house that here; i dunno, i'm not very good at these things. the mapping of exception types to a header hinting to the reason seems unique to this codepath, whereas deadlines-java is primarily concerned with throwing an exception of the right type, at the right time.

carterkozak: Ya, I really the idea of encapsulating this within the deadlines library

bjlaub: draft of that here: https://github.com/palantir/deadlines-java/pull/30

bjlaub: i've updated this PR to incorporate the changes that set a header+status

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -72,6 +75,10 @@ public void handle(HttpServerExchange exchange, Throwable throwable) {
             illegalArgumentException(exchange, throwable);
         } else if (throwable instanceof FrameworkException) {
             frameworkException(exchange, (FrameworkException) throwable);
+        } else if (throwable instanceof DeadlineExpiredException.External) {
+            externalDeadlineExpiredException(exchange, (DeadlineExpiredException.External) throwable);
+        } else if (throwable instanceof DeadlineExpiredException.Internal) {
+            internalDeadlineExpiredException(exchange, (DeadlineExpiredException.Internal) throwable);
```
carterkozak: We might as well handle this in one branch, that way new types of `DeadlineExpiredException` are handled without code change (caveat being that the logging may not be ideal for that case, but behavior should be reasonable)

bjlaub: done

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -183,6 +188,13 @@ private static void frameworkException(HttpServerExchange exchange, FrameworkExc
         writeResponse(exchange, Optional.of(ConjureError.fromServiceException(exception)), statusCode);
     }
 
+    private static void deadlineExpiredException(HttpServerExchange exchange, DeadlineExpiredException exception) {
+        log.info("deadline exceeded", exception);
```
carterkozak: should we specifically log the internal variant at `error` and everything else at `info`?

bjlaub: sure, done

carterkozak (approved):

PR #2456, opened 2025-02-13
Snapshot Testing Replacement

kunalrkak:
## Before this PR

We have lots of disjoint snapshot test coverage, which makes it difficult when adding a new feature flag for example to ensure compatibility with other permutations of options. This PR sets up a new framework for auto-generation of classes via a parameterized test which should help make it easier to cover more desired option permutations in the future.

## After this PR

==COMMIT_MSG==
Initializes a new framework for snapshot testing
==COMMIT_MSG==


conjure-java-core-tests/src/main/java/com/palantir/internal/ParameterizedTestCase.java
```diff
@@ -0,0 +1,265 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.annotation.processing.Generated;
+
+@JsonDeserialize(builder = ParameterizedTestCase.Builder.class)
+@Generated("com.palantir.conjure.java.types.BeanGenerator")
+public final class ParameterizedTestCase {
```
mpritham: Thoughts on having this just be a record? It allows us to control what we want to do in this class. I can think of a case where we want to do some custom validation on construction, e.g. maybe we shouldn't allow setting `preferObjectBuilders` if `Object` is not one of the generators.

mpritham:
Should we attempt to update the tests in `conjure-java-core` to be parameterized using `ParameterizedTestCase`? We may be able to update the existing test logic to emit to some subdirectory of `integrationInput` there.

I think this may break some tests (like [this ete test](https://github.com/palantir/conjure-java/blob/e4f5cb6bd9b12b97ff690f2686ac8f81c4163ec0/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java#L94)). One fix is to generate new files for the ete tests, and move the test class to the appropriate package (named after the test).

edit: An example test that may fail is the [`UndertowServiceEteTest`](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java#L94) when it tests the endpoints in `ete-service.yml`: there is [an external type import for a Conjure generated type](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/test/resources/ete-service.yml#L6) which is not generated in the `UndertowServiceEteTest`.

conjure-java-core-tests/src/main/java/com/palantir/internal/ParameterizedTestCase.java
kunalrkak: Yeah good call - I just generated these out of simplicity but records make sense

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/ParameterizedConjureGenerationTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @ParameterizedTest
+    @MethodSource("getTestCases")
+    void testGeneratedCode(ParameterizedTestCase testCase) throws IOException {
+        ConjureDefinition def = Conjure.parse(testCase.files().stream()
+                .map(filePath -> new File(filePath.get()))
+                .toList());
+        List<Path> files = new GenerationCoordinator(MoreExecutors.directExecutor(), testCase.getGenerators()).emit(def, tempDir);
+
+        assertThatFilesAreTheSame(files);
```
mpritham: I wonder if we could pass in the `docs` field in the `ParameterizedTestCase` here. In `assertThatFilesAreTheSame` we could do `assertThat(file).as(testCase.docs()).hasSameContentAs(expectedFile)`. This description could help us see which test case failed quickly because it would appear in the test failure output.

kunalrkak: Oh that's neat, I'll try that!

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/TestCases.java
```diff
@@ -0,0 +1,56 @@
...
+                    .options(Options.builder()
+                            .useImmutableBytes(true)
+                            .strictObjects(true)
+                            .nonNullCollections(true)
+                            .useStagedBuilders(true)
+                            .excludeEmptyOptionals(true)
+                            .preferObjectBuilders(true)
+                            .build())
+                    .generators(GeneratorType.OBJECT)
+                    .build())
```
kunalrkak: This case makes the diff large -- happy to move this out into a separate PR

conjure-java-core/build.gradle
```diff
@@ -122,3 +122,4 @@ tasks.withType(JavaCompile) {
     // this complains about the integrationInput directory
     options.errorprone.disable 'StrictUnusedVariable'
 }
+compileJava.dependsOn compileIntegrationInputJava
```
carterkozak: Is this line necessary? I don't think compilation of the main source set should depend on the compilation of the integration input sources. I'd expect this line to require integration input compilation before our tests can compile: https://github.com/palantir/conjure-java/blob/8f78f7d9a371413c8b5caca318c1fa7d44948c16/conjure-java-core/build.gradle#L67

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/ParameterizedConjureGenerationTest.java
```diff
@@ -0,0 +1,71 @@
...
+                new GenerationCoordinator(MoreExecutors.directExecutor(), testCase.getGenerators()).emit(def, tempDir);
+
+        assertThatFilesAreTheSame(files, testCase.docs());
+    }
+
+    private void assertThatFilesAreTheSame(List<Path> files, String testCaseDocs) throws IOException {
+        for (Path file : files) {
+            Path relativized = tempDir.toPath().relativize(file);
+            Path expectedFile = Paths.get(REFERENCE_FILES_FOLDER, relativized.toString());
+            if (Boolean.parseBoolean(System.getProperty("recreate", "false"))) {
```
carterkozak: In the recreate case, we may want to delete all the data in tje expected output directory first, this way when we remove generated types, they don't get left around in the integrationInputs sources forever:
`Paths.get(REFERENCE_FILES_FOLDER, testCase.getPackageName().replace('.', File.separatorChar))`

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/TestCases.java
carterkozak: I'm not worried about the generated sources, github UI makes them easy to ignore using our `.gitattributes` config here: https://github.com/palantir/conjure-java/blob/838d57701edc689bfcebe8c70ad58bd50353b472/.gitattributes#L1

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/objects/GeneratorType.java
```diff
@@ -0,0 +1,342 @@
...
+ * <p>For example, {@code GeneratorType.valueOf("corrupted value").get()} will return
+ * {@link Value#UNKNOWN}, but {@link GeneratorType#toString} will return "corrupted value".
+ *
+ * <p>There is no method to access all instantiations of this class, since they cannot be known at compile time.
+ */
+@SuppressWarnings("checkstyle:HiddenField")
+@Generated("com.palantir.conjure.java.types.EnumGenerator")
+@Safe
+@Immutable
+public final class GeneratorType {
```
carterkozak: I don't think we're gaining anything by using checked-in conjure generated types for `FilePath` and `GeneratorType` -- I think we might as well use a standard java enum for `GeneratorType` and perhaps a standard `java.nio.file.Path` for the path, if possible

conjure-java-core/build.gradle
kunalrkak: yeah that makes sense, its already wired up to compiling the test code so will remove this!

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/ParameterizedConjureGenerationTest.java
kunalrkak: Good call, will do -- any thoughts on how to handle removing a test case entirely where "recreate" wouldn't get called?

carterkozak: We could centralize our `TestCase` definitions into a single enum, where we validate that only registered packages are present. I think this case is easier to handle manually than removal/replacement/rename of a single object, since removing a test case means removing an entire package root. I don't think we need to implement anything in this PR, but perhaps follow-ups later could be helpful

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/TestCases.java
kunalrkak: Ah I'm assuming this is what's automatically collapsing all the generated files in the diff view! Very neat

carterkozak (commented):
thanks, lgtm! @mpritham could you give final approval if you're happy with this?

mpritham (approved):
Looks great! Thank you!

PR #2462, opened 2025-02-25
Add Jakarta dependencies to integration test inputs

kunalrkak:
## Before this PR

We'd like to check-in generated code, including a test that creates object/service definitions with jakarta packages. This currently doesn't work since jakarta isn't a dependency of the integration inputs.

## After this PR

==COMMIT_MSG==
Adds a dependency on jakarta to integration inputs.
==COMMIT_MSG==


kunalrkak:
Pulled this change into a larger PR

PR #2464, opened 2025-02-26
[Enhancement] Strict Staged Builders Support Primitive Optimized Collections

bmarcaur:
## Before this PR
When using `useStrictStagedBuilders` and `nonNullCollections` the generated builders did not allow for the optimal (non-boxing) path (https://github.com/palantir/conjure-java/pull/2389), nor did it optimally handle Jackson deserialization (https://github.com/palantir/conjure-java/pull/2431).

## After this PR
Strict staged builders now expose additional, optimized setters and also leverage these setters for Jackson deserialization.

==COMMIT_MSG==
==COMMIT_MSG==

## Possible downsides?
IMO this class is becoming unwieldy. It has two similar, but somewhat divergent code paths between strict and non-strict builders. You cannot simply hoist the strictness to the top level because it has entry points like `addStagedBuilder` which you can call even when `useStrictStagedBuilders` is set to true? Kinda feels like we should have different builder implementations for each of the paths through and we then factor out the bit of shared behavior. IDK it felt bad adding all these `boolean strict` all over the place.

In addition, I added some complexity to support [adding the collection reset](https://github.com/palantir/conjure-java/blob/07e5ae9700fe95e45e9e8c9524c13ce10a9b3969/conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveStrictExample.java#L196). If we are comfortable allowing people to cast into the base builder and break shit.. Then I can remove some of these unwanted `boolean stricts`.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
@@ -172,9 +172,8 @@ public final class BeanBuilderGenerator {
         addJsonDeserializeUsingBuilderAnnotation(specBuilder, STAGED_BUILDER_IMPLEMENTATION_NAME);
 
         List<EnrichedField> allFields = enrichFields(typeDef.getFields());
-        List<TypeSpec> interfaces =
-                generateIntermediateStageInterfaces(objectClass, builderClass, typeMapper, allFields, safetyEvaluator);
-        interfaces.add(generateBuilderFinalStageInterface(List.of()));
+        List<TypeSpec> interfaces = generateIntermediateStageInterfaces(allFields, true);
+        interfaces.add(generateBuilderFinalStageInterface(List.of(), true));
         specBuilder.addTypes(interfaces);
 
         addBuilderInterfaceAndMethod(specBuilder, interfaces);
@@ -203,11 +202,12 @@ public final class BeanBuilderGenerator {
 
         addJsonDeserializeUsingBuilderAnnotation(specBuilder, STAGED_BUILDER_IMPLEMENTATION_NAME);
 
-        List<TypeSpec> interfaces = generateIntermediateStageInterfaces(
-                objectClass, builderClass, typeMapper, fieldsNeedingBuilderStage, safetyEvaluator);
-        interfaces.add(generateBuilderFinalStageInterface(allFields.stream()
-                .filter(field -> !fieldsNeedingBuilderStage.contains(field))
-                .collect(Collectors.toList())));
+        List<TypeSpec> interfaces = generateIntermediateStageInterfaces(fieldsNeedingBuilderStage, false);
+        interfaces.add(generateBuilderFinalStageInterface(
+                allFields.stream()
+                        .filter(field -> !fieldsNeedingBuilderStage.contains(field))
+                        .collect(Collectors.toList()),
+                false));
         specBuilder.addTypes(interfaces);
         addBuilderInterfaceAndMethod(specBuilder, interfaces);
 
@@ -264,12 +264,8 @@ public final class BeanBuilderGenerator {
                         .build());
     }
 
-    private static List<TypeSpec> generateIntermediateStageInterfaces(
-            ClassName objectClass,
-            ClassName builderClass,
-            TypeMapper typeMapper,
-            List<EnrichedField> fieldsNeedingBuilderStage,
-            SafetyEvaluator safetyEvaluator) {
+    private List<TypeSpec> generateIntermediateStageInterfaces(
+            List<EnrichedField> fieldsNeedingBuilderStage, boolean strict) {
         List<TypeSpec.Builder> interfaces = new ArrayList<>();
 
         PeekingIterator<EnrichedField> fieldPeekingIterator =
@@ -277,6 +273,7 @@ public final class BeanBuilderGenerator {
 
         while (fieldPeekingIterator.hasNext()) {
             EnrichedField field = fieldPeekingIterator.next();
+            Type type = field.conjureDef().getType();
             String nextBuilderStageName = fieldPeekingIterator.hasNext()
                     ? JavaNameSanitizer.sanitize(fieldPeekingIterator.peek().fieldName())
                     : "completed_";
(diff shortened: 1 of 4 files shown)


conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveStrictExample.java
```diff
@@ -164,14 +173,32 @@ public Builder ints(@Nonnull Iterable<Integer> ints) {
         }
 
         @Override
-        @JsonSetter(value = "doubles", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
+        @JsonSetter(value = "ints", nulls = Nulls.SKIP, contentNulls = Nulls.FAIL)
+        public Builder ints(@Nonnull int... ints) {
+            checkNotBuilt();
+            this.ints = ConjureCollections.newNonNullIntegerList();
```
bmarcaur: Interested in everyone's take on whether this level of defensiveness is strictly (heh) necessary. Anyone that is determined enough to break out of the type safety of the strict staged builder kinda DGAF anymore. I only ask because factoring to support this added a ton of `strict` sprawl.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/BeanBuilderGenerator.java
```diff
@@ -805,20 +829,13 @@ private boolean isByteBuffer(Type type) {
...
-            Map<com.palantir.conjure.spec.TypeName, TypeDefinition> typesMap,
-            boolean override) {
+    private List<MethodSpec> createAuxiliarySetters(EnrichedField enriched, boolean override) {
         Type type = enriched.conjureDef().getType();
         Optional<LogSafety> safety = safetyEvaluator.getUsageTimeSafety(enriched.conjureDef());
 
         ImmutableList.Builder<MethodSpec> builder = ImmutableList.builder();
 
         if (type.accept(TypeVisitor.IS_LIST)) {
-            if (isPrimitiveOptimized(type)) {
```
bmarcaur, line 811: Hoisted this up as having it nested in the auxiliary setters was a hold over from a time before `isPrimitiveOptimized` properly guarded for collection types only. Hence it needing to be nested inside a `type.accept(TypeVisitor.IS_LIST)`. However, this is not strictly a refactor for refactor's sake. I did need to add the primitive setter in more cases than what this method is used for.

conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveStrictExample.java
carterkozak: > Anyone that is determined enough to break out of the type safety of the strict staged builder kinda DGAF anymore

I think there's a bit of subtlety here because it's not only misuse of the strict builder that could trigger the case you're defending against, but also deserialization where we tell jackson `@JsonDeserialize(builder = PrimitiveStrictExample.DefaultBuilder.class)`. It's entirely reflective and doesn't know how to use staged builders, so the package-private implementation is used directly.

I'm supportive of being defensive here, I generally weigh the quality of the codegen output much higher than the quality of the codegen implementation code -- not that it should be _bad_, but I'd prefer to make the generated code as close to perfect as we can reasonably manage.

carterkozak: There's an argument for adding an overload of `public static List<Integer> newNonNullIntegerList(Iterable<Integer> iterable)` to do this in one method call `public static List<Integer> newNonNullIntegerList(int[] items)` and simplify the generated code a bit, but that would take a bit to propagate out. I don't think the upside is necessarily high enough to justify broadening the API and managing its rollout.

carterkozak:
I think this will need a rebase once https://github.com/palantir/conjure-java/pull/2463 merges given we've replaced some of the integrationInput test infrastructure.

@mpritham Could you give this a final review + approval once the test refactoring PR has merged and this has rebased?

conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveStrictExample.java
bmarcaur: I am likely gonna have to add another overloaded endpoint for something akin to `public static void addToDoubleList(double double)` for an optimized path to singular adds. I could throw this in as well and just allow it to bake. Im not in a rush to get this in. But I did want it out of my head.

bmarcaur:
> I think this will need a rebase once #2463 merges given we've replaced some of the integrationInput test infrastructure.
> 
> @mpritham Could you give this a final review + approval once the test refactoring PR has merged and this has rebased?

Yup, chatted with Pritham about this. I'm not in a rush so would rather not disrupt Kunal.

kunalrkak:
fysa #2463 has merged in, this should be good to rebase! running the test [here](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/ParameterizedConjureGenerationTest.java#L37) (with the recreate flag set) will regenerate `PrimitiveStrictExample` with your changes

bmarcaur:
> fysa #2463 has merged in, this should be good to rebase! running the test [here](https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/ParameterizedConjureGenerationTest.java#L37) (with the recreate flag set) will regenerate `PrimitiveStrictExample` with your changes

@kunalrkak @mpritham It rebased eerily cleanly. I still ran `gw :conjure-java-core:test -Drecreate=true` but it produced no changes so I think this is clean and GTG.

edit: I think I should still address https://github.com/palantir/conjure-java/pull/2464#discussion_r1970821403 before this merges. Its a good suggestion. It will only be a one line change in the generation code. So a review on this need not wait, but you can if you want.

conjure-java-core/src/integrationInput/java/com/palantir/product/PrimitiveStrictExample.java
bmarcaur: https://github.com/palantir/conjure-java/pull/2475

bmarcaur: I rebased on ^ so now this is a non-issue 👍🏼

bmarcaur:
Marking `DNM` until https://github.com/palantir/conjure-java/pull/2475 has soaked for a while

mpritham (approved):

PR #2467, opened 2025-02-28
Deserialize sets preserving order in union and alias types

kunalrkak:
## Before this PR

Jackson deserializes sets into `HashSet`s by default, which loses any element ordering and means roundtrip serde is equal but not identical. Instructing Jackson to deserialize sets as `LinkedHashSet`s instead maintains order.

## After this PR

==COMMIT_MSG==
Instructs Jackson to deserialize sets within union and alias types as `LinkedHashSet`s in order to preserve element ordering.
==COMMIT_MSG==

## Possible downsides?

We never guaranteed ordering of sets so modifying this shouldn't be an issue.

### DIFF
--- a/changelog/@unreleased/pr-2467.v2.yml
+++ b/changelog/@unreleased/pr-2467.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Instructs Jackson to deserialize sets within union and alias types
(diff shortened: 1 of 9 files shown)


carterkozak (approved):
Great tests, thanks!

PR #2473, opened 2025-03-03
[do-not-merge] Generate reachability-metadata.json files for generated object types

mpritham:
## Context
In order to use reflection in native executables created with [Graal's `native-image` feature](https://www.graalvm.org/latest/reference-manual/native-image/) we need to provide the Graal compiler a listing of the types and methods we'd like to be accessible for reflection. This is exposed via [reachability metadata files](https://www.graalvm.org/latest/reference-manual/native-image/metadata/) stored in `META-INF/native-image/**/reachability-metadata.json` files.

## Before this PR
No reflection metadata is generated. Jackson is unable to deserialize Conjure object types.

## After this PR
When creating the metadata files, instead of listing the methods accessed for reflection, we use the convenience functions `allDeclaredFields`, `allPublicFields` and `allDeclaredConstructors`.

==COMMIT_MSG==
Generate reachability-metadata.json files for generated object types
==COMMIT_MSG==

## Possible downsides?
We over-include certain methods: several methods in the `Builder` classes are convenience overrides and are not accessed via reflection by Jackson (e.g. `addAllX` methods for collections fields). This over-inclusion isn't preferred: 1) it'd be safer to enforce that only a certain fields are accessible via reflection to prohibit arbitrary uses of reflection, 2) including extra methods increases binary size. We are still in the early stages of exploring Java native executables and have opt'd to over-include methods accessible for reflection to iterate faster.


conjure-java-core/src/main/java/com/palantir/conjure/java/GeneratedFile.java
```diff
@@ -0,0 +1,33 @@
...
+import com.palantir.javapoet.JavaFile;
+
+public sealed interface GeneratedFile permits GeneratedJavaFile, GeneratedReachabilityMetadataFile {
+    record GeneratedJavaFile(JavaFile javaFile) implements GeneratedFile {
+        public static GeneratedJavaFile of(JavaFile javaFile) {
+            return new GeneratedJavaFile(javaFile);
+        }
+    }
+
+    record GeneratedReachabilityMetadataFile(String packageName, ReachabilityMetadata reachabilityMetadata)
```
mpritham, line 31: Generating one JSON per package here. An alternative is to create one JSON per type. This is what we'd want to do when generating metadata in an annotation library as it allows Gradle to make use of incremental compilation: skip generation of metadata for classes that have not changed.

For `conjure-java`, I can go either way. One JSON per type is a load of files and slightly more difficult to inspect manually so I opted to do this. (it's not really all that much more difficult - I can just `cat` all the files in a directory)

PR #2475, opened 2025-03-06
New ConjureCollections Methods for Future Optimizations

bmarcaur:
## Before this PR
There is no way to create a new optimized collection with an existing primitive array (think when deserializing from Jackson) nor was there a way to optimally add a single element.

Addresses the discussion in https://github.com/palantir/conjure-java/pull/2464#discussion_r1973729961

## After this PR
This PR stages those methods so that in a later PR, once most places have adopted this version, we can start generating code to allow end users to have these routines.


mpritham (approved):
Thanks!

PR #2476, opened 2025-03-06
Add All Primitive Collection Types to the Test Example

bmarcaur:
## Before this PR
I was making some changes in https://github.com/palantir/conjure-java/pull/2464 and was wondering if I had written a bug that might cause the other types to behave incorrectly. 

## After this PR
I added all the primitive collection types (optimized or not) so that when future changes land I can easily notice if I correctly targeted only the types I care about.

## Possible downsides?
I used a type documentation intentionally. I figured that in the future someone will notice the difference in the generated code and presume it is a bug. This way the docs are right there inline with the type.

### DIFF
--- a/conjure-java-core/src/test/resources/primitive-collections.yml
+++ b/conjure-java-core/src/test/resources/primitive-collections.yml
@@ -9,3 +9,16 @@ types:
           field: integer
           ints: list<integer>
           doubles: list<double>
+          # this is stored optimally but does not have boxing optimizations
+          longs:
+            type: list<safelong>
+            docs: |
+              This primitive type is stored optimally, but does not have boxing optimizations.
+              
+              This type was added in case we choose to change that later we have a comparison.
+          bools:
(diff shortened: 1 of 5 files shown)


conjure-java-core/src/test/resources/primitive-collections.yml
```diff
@@ -9,3 +9,16 @@ types:
           field: integer
           ints: list<integer>
           doubles: list<double>
+          # this is stored optimally but does not have boxing optimizations
```
mpritham: nit:

```suggestion
```

mpritham (dismissed):
Thanks!

conjure-java-core/src/test/resources/primitive-collections.yml
bmarcaur: BLAST!

mpritham (approved):

PR #2480, opened 2025-03-11
Add validation for snapshot folder contents

kunalrkak:
## Before this PR

Follow-up from https://github.com/palantir/conjure-java/pull/2456#discussion_r1961983538
Adds a test to validate the packages in the snapshot testing folder. This will fail if: 
1. A test case is removed but the corresponding package root in the snapshot testing directory is not deleted
2. Generated objects are checked into the directory but are not registered as a test case

This will just serve as a rough guardrail to make sure contributors use the testing framework when generating new reference objects.

## After this PR

==COMMIT_MSG==
Adds a test to ensure only registered test case content exists in the snapshot testing folder.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2480.v2.yml
+++ b/changelog/@unreleased/pr-2480.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Add a test to ensure only registered test case content exists in the
+    snapshot testing folder.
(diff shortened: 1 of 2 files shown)


mpritham (approved):

mpritham:
Thanks!

PR #2486, opened 2025-03-17
[Improvement] Primitive Optimized Singular Addition Implementation

bmarcaur:
## Before this PR
Before the singular additions methods e.g. `public Builder ints(int ints)` would leverage `boolean add(E e)` from the `Collection` interface. This would cause each addition to be boxed before it is unboxed at rest.

## After this PR
Now the implementation delegates to a non-boxing addition path when it is a primitive list. This brings singular additions inline with the bulk addition optimizations.

### DIFF
--- a/changelog/@unreleased/pr-2486.v2.yml
+++ b/changelog/@unreleased/pr-2486.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: '[Improvement] Primitive Optimized Singular Addition Implementation'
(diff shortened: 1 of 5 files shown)


bmarcaur:
Marking `DNM` until https://github.com/palantir/conjure-java/pull/2475 has soaked for a while

mpritham (dismissed):
Thanks!

mpritham (approved):

PR #2520, opened 2025-04-23
Handle remote peer exceptions

ERogalla:
## Before this PR
We have experienced many IO exceptions originating from this [undertow message](https://sourcegraph.com/github.com/undertow-io/undertow/-/blob/core/src/main/java/io/undertow/UndertowMessages.java?L427-L428).

## After this PR
Don't rethrow on this error.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
@@ -208,7 +208,8 @@ public enum ConjureExceptions implements ExceptionHandler {
 
         if (ioException.equals(UndertowMessages.MESSAGES.couldNotReadContentLengthData())) {
             log.info(
-                    "Remote peer closed connection before all data could be read. The request may have been aborted by the client",
+                    "Remote peer closed connection before all data could be read. The request may have been aborted by"
+                            + " the client",
                     ioException);
             return;
         }
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
@@ -80,12 +80,8 @@ public enum ConjureExceptions implements ExceptionHandler {
             deadlineExpiredException(exchange, (DeadlineExpiredException) throwable);
         } else if (throwable instanceof Error) {
             error(exchange, (Error) throwable);
-        } else if (throwable instanceof IOException
-                && (!exchange.getConnection().isOpen()
-                        || throwable.equals(UndertowMessages.MESSAGES.couldNotReadContentLengthData()))) {
-            log.info(
-                    "I/O exception from a closed connection. The request may have been aborted by the client",
-                    throwable);
+        } else if (throwable instanceof IOException) {
+            ioException(exchange, (IOException) throwable);
         } else {
             ServiceException exception = new ServiceException(ErrorType.INTERNAL, throwable);
             log(exception, throwable);
@@ -200,6 +196,29 @@ public enum ConjureExceptions implements ExceptionHandler {
         // writeResponse is called by UndertowDeadlineReasonResponseEncodingAdapter#setStatus
         DeadlineExpiredReasons.encodeToResponse(
                 exception, exchange, UndertowDeadlineReasonResponseEncodingAdapter.INSTANCE);
+    }
+
+    private static void ioException(HttpServerExchange exchange, IOException ioException) {
+        if (!exchange.getConnection().isOpen()) {
+            log.info(
+                    "I/O exception from a closed connection. The request may have been aborted by the client",
+                    ioException);
+            return;
+        }
+
+        if (ioException.equals(UndertowMessages.MESSAGES.couldNotReadContentLengthData())) {
+            log.info(
+                    "Remote peer closed connection before all data could be read. The request may have been aborted by the client",
+                    ioException);
+            return;
+        }
+
+        ServiceException exception = new ServiceException(ErrorType.INTERNAL, ioException);
+        log(exception, ioException);
+        writeResponse(
+                exchange,
+                Optional.of(ConjureError.fromServiceException(exception)),
+                exception.getErrorType().httpErrorCode());
     }
 
     private static void error(HttpServerExchange exchange, Error error) {
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
@@ -32,6 +32,7 @@ import com.palantir.deadlines.DeadlineExpiredReasons;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
+import io.undertow.UndertowMessages;
 import io.undertow.io.UndertowOutputStream;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.util.Headers;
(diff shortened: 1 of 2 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -79,7 +80,9 @@ public void handle(HttpServerExchange exchange, Throwable throwable) {
             deadlineExpiredException(exchange, (DeadlineExpiredException) throwable);
         } else if (throwable instanceof Error) {
             error(exchange, (Error) throwable);
-        } else if (throwable instanceof IOException && !exchange.getConnection().isOpen()) {
+        } else if (throwable instanceof IOException
+                && (!exchange.getConnection().isOpen()
+                        || throwable.equals(UndertowMessages.MESSAGES.couldNotReadContentLengthData()))) {
```
pkoenig10: nit: This should probably be pulled out to a helper method so we can give it a descriptive name and avoid cluttering this existing `instanceof` casing.

carterkozak: I don't think this works, I'd suggest taking another pass at the test to see what's happening in practice!

Creating a new IOException + gathering its stack trace every time we catch any sort of IOException isn't ideal.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -199,6 +198,34 @@ private static void deadlineExpiredException(HttpServerExchange exchange, Deadli
...
+                    ioException);
+            return;
+        }
+
+        if (ioException
+                .getMessage()
+                .equals(UndertowMessages.MESSAGES
+                        .couldNotReadContentLengthData()
+                        .getMessage())) {
+            log.info(
```
ERogalla, line 214: Does it make sense to not send an error here? Or should it be a 400 error. 
For reference, this error message is pulled directly from [UndertowMessages](https://sourcegraph.com/github.com/undertow-io/undertow/-/blob/core/src/main/java/io/undertow/UndertowMessages.java?L426-428)

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
ERogalla: Updated. @carterkozak you're right. The original test I added did not trigger the added codepath.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -79,10 +80,8 @@ public void handle(HttpServerExchange exchange, Throwable throwable) {
             deadlineExpiredException(exchange, (DeadlineExpiredException) throwable);
         } else if (throwable instanceof Error) {
             error(exchange, (Error) throwable);
-        } else if (throwable instanceof IOException && !exchange.getConnection().isOpen()) {
-            log.info(
-                    "I/O exception from a closed connection. The request may have been aborted by the client",
-                    throwable);
+        } else if (throwable instanceof IOException) {
+            ioException(exchange, (IOException) throwable);
```
schlosna: nit: can use instanceof pattern variable to avoid cast:

```suggestion
        } else if (throwable instanceof IOException ioException) {
            ioException(exchange, ioException);
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -199,6 +198,34 @@ private static void deadlineExpiredException(HttpServerExchange exchange, Deadli
...
+
+        if (ioException
+                .getMessage()
+                .equals(UndertowMessages.MESSAGES
+                        .couldNotReadContentLengthData()
+                        .getMessage())) {
+            log.info(
+                    "Remote peer closed connection before all data could be read. The request may have been aborted by"
+                            + " the client",
+                    ioException);
```
schlosna, line 217:
```suggestion
                    ioException);
            IoUtils.safeClose(exchange.getConnection());
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -199,6 +198,34 @@ private static void deadlineExpiredException(HttpServerExchange exchange, Deadli
...
+                    "I/O exception from a closed connection. The request may have been aborted by the client",
+                    ioException);
+            return;
+        }
+
+        if (ioException
+                .getMessage()
+                .equals(UndertowMessages.MESSAGES
+                        .couldNotReadContentLengthData()
+                        .getMessage())) {
```
schlosna: `ioException.getMessage()` could return `null` leading to a `NullPointerException` on this error handling path, so we need to handle that.

Additionally, `UndertowMessages.MESSAGES.couldNotReadContentLengthData()` creates a new `IOException` on each invocation and adds the expensive overhead of `Throwable#fillInStackTrace()`. We should resolve the message once to a static final field.

I've pushed https://github.com/palantir/conjure-java/pull/2523 as suggested changes on top of this branch

```suggestion
        if (Objects.equals(ioException.getMessage(), COULD_NOT_READ_CONTENT_LENGTH_MESSAGE)) {
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
schlosna: We could not read the full request from the client and client has closed its side of the connection so we can't send a response, but we need to close and clean up our exchange.

schlosna (commented):
I'd suggest https://github.com/palantir/conjure-java/pull/2523 on top of this PR

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
```diff
@@ -361,6 +362,13 @@ public void handlesErrorWithoutRethrowing() {
                 .doesNotThrowAnyException();
     }
 
+    @Test
+    public void handlesPeerRemoteExceptionsWithoutRethrowing() throws IOException {
+        exception = UndertowMessages.MESSAGES.couldNotReadContentLengthData();
+        HttpURLConnection connection = execute();
+        assertThat(connection.getResponseCode()).isEqualTo(200);
```
schlosna: there shouldn't be a response code if the request wasn't fully read and we threw before processing & writing response
```suggestion
        assertThatThrownBy(() -> connection.getResponseCode())
                .isInstanceOf(SocketException.class)
                .hasMessageContaining("Unexpected end of file from server");
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
pkoenig10: Let's adopt this convention for all the branches here for consistency.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -199,6 +198,34 @@ private static void deadlineExpiredException(HttpServerExchange exchange, Deadli
                 exception, exchange, UndertowDeadlineReasonResponseEncodingAdapter.INSTANCE);
     }
 
+    private static void ioException(HttpServerExchange exchange, IOException ioException) {
```
pkoenig10, line 205: nit: Move this below `error` to match the ordering of the cases above.

carterkozak:
Can you help me understand the core problem that we’re solving? The existing exception itself isn’t unreasonable imo, but it could certainly be handled better. Is that signal causing monitors to fire?

I’m a bit worried that we’re matching a specific failure where we should probably look for a class of failures, as there are other ways that requests can bail out depending on the transport encoding.

ERogalla:
This was not causing any monitors to fire, but was just spamming error service logs. I was hoping to clear out any noisy/incorrect error logs that aren't actual errors. We can instruct our team to ignore these errors, but it is occurring on seemingly random service endpoints.

FWIW I'm totally open to close this PR out if this isn't the root cause. I just assumed that this was just some unhandled (but expected) exception.

carterkozak:
In that case, I'd attempt to cast a more precise net: Instead of updating our exception handling to check messages, we may prefer to wrap the request body conduit/channel such that we can intercept I/O exceptions while reading data and handle them more precisely without impacting the flow of exceptions thrown within service implementations.

PR #2524, opened 2025-05-06
ConjureCollections uses forEach where possible

schlosna:
## Before this PR

`ConjureCollections` used iterator based for loops

## After this PR

==COMMIT_MSG==
ConjureCollections uses `Iterable#forEach` where possible to minimize allocations and speed up iteration over collections.
==COMMIT_MSG==

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
@@ -18,9 +18,11 @@ package com.palantir.conjure.java.lib.internal;
 
 import com.palantir.conjure.java.lib.SafeLong;
 import com.palantir.logsafe.Preconditions;
+import java.util.AbstractCollection;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
@@ -41,8 +43,8 @@ public final class ConjureCollections {
     /*
      * This is bizarre. Allow me to explain...
      *
-     * We do _not_ want to expose the Conjure*List types externally
-     * but we also want the optimizations they provide to make it thru
+     * We do _not_ want to expose the Conjure*List types externally,
+     * but we also want the optimizations they provide to make it through
      * to jackson for serialization. So the runtime type needs to be
      * preserved while also not exposing the type :phew:.
      *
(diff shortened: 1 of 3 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -51,79 +53,82 @@ private ConjureCollections() {
...
             // This special-casing allows us to take advantage of the more performant
             // ArrayList#addAll method which does a single System.arraycopy.
-            addTo.addAll((Collection<T>) elementsToAdd);
+            addTo.addAll(collection);
         } else {
-            for (T element : elementsToAdd) {
-                addTo.add(element);
-            }
+            Preconditions.checkNotNull(elementsToAdd, "elementsToAdd cannot be null")
+                    .forEach(addTo::add);
```
aldexis, line 79: It seems like the default `forEach` on Iterable does
```
    default void forEach(Consumer<? super T> action) {
        Objects.requireNonNull(action);
        for (T t : this) {
            action.accept(t);
        }
    }
```
So you know for which iterable types you expect this to be an improvement? (I've found a couple, for it's not always implemented differently)

(I still think it's better because in the worst case, it will be the same, but I don't believe this will necessarily net us benefits in most cases)

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -141,10 +146,7 @@ public static <T> List<T> newNonNullList() {
 
     public static <T> List<T> newNonNullList(Iterable<? extends T> iterable) {
         List<T> arrayList = newList(iterable);
-        for (T item : arrayList) {
-            Preconditions.checkNotNull(item, "iterable cannot contain null elements");
-        }
-
+        arrayList.forEach(ConjureCollections::checkNotNullElement);
```
aldexis, line 149: Note to self: ArrayList overrides forEach and does not allocate an iterator, so even though we already loop over the contents in `newList(iterable)`, in the worst case, this won't allocate two iterators

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -51,79 +53,82 @@ private ConjureCollections() {
...
 
     // Prefer to use newList(iterable)
     // explicitly need to return mutable list for generated builders
     @SuppressWarnings({"IllegalType", "unchecked", "NonApiType"})
     public static <T> ArrayList<T> newArrayList(Iterable<? extends T> iterable) {
-        Preconditions.checkNotNull(iterable, "iterable cannot be null");
-        if (iterable instanceof Collection) {
-            return new ArrayList<>((Collection<T>) iterable);
+        if (iterable instanceof Collection<? extends T> collection) {
+            return new ArrayList<>(collection);
```
aldexis, line 117: Not for this PR, but more for my understanding/knowledge, since we're looking at optimizing allocations: From what I can tell, this might allocate the underlying array twice?
This does
```
    public ArrayList(Collection<? extends E> c) {
        Object[] a = c.toArray();
        if ((size = a.length) != 0) {
            if (c.getClass() == ArrayList.class) {
                elementData = a;
            } else {
                elementData = Arrays.copyOf(a, size, Object[].class);
            }
        } else {
            // replace with empty array.
            elementData = EMPTY_ELEMENTDATA;
        }
    }
```
and `Collection#toArray` says
```
The returned array will be "safe" in that no references to it are maintained by this collection. (In other words, this method must allocate a new array even if this collection is backed by an array). The caller is thus free to modify the returned array.
```
Unfortunately, it seems like `addAll` also does the same 🤔 (so using `new ArrayList(collection.size())` and `addAll` wouldn't help either)
```
    public boolean addAll(Collection<? extends E> c) {
        Object[] a = c.toArray();
        modCount++;
        int numNew = a.length;
        if (numNew == 0)
            return false;
        Object[] elementData;
        final int s;
        if (numNew > (elementData = this.elementData).length - (s = size))
            elementData = grow(s + numNew);
        System.arraycopy(a, 0, elementData, s, numNew);
        size = s + numNew;
        return true;
    }
```
Curious if you have any knowledge about this and why we would be allocating the same array twice

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -162,10 +164,7 @@ public static <T> Set<T> newNonNullSet() {
 
     public static <T> Set<T> newNonNullSet(Iterable<? extends T> iterable) {
         Set<T> set = newSet(iterable);
-        for (T item : set) {
-            Preconditions.checkNotNull(item, "iterable cannot contain null elements");
-        }
-
+        set.forEach(ConjureCollections::checkNotNullElement);
```
aldexis, line 167: Fwiw, LinkedHashSet does not seem to override Iterable's default forEach implementation, so we'll still create the iterator here afaict

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -51,79 +53,82 @@ private ConjureCollections() {
...
+                    for (Object element : array) {
+                        checkNotNullElement(element);
+                    }
+                    return array;
+                }
+
+                @Override
+                public int size() {
+                    return collection.size();
+                }
```
carterkozak, line 104: In an ideal world I think we'd implement this within specialized non-null collections to avoid the new anonymous wrapper allocation on a per-call basis. Getting there would be a little tricky, because this API is fairly specialized, however it's in the `internal` package so we can make some assumptions about how it's used.
Probably not a substantial optimization.

carterkozak: I suppose another option is to make the assumption that `toArray` will be used, and explicitly handle that ourselves:
```java
T[] values = elementsToAdd.toArray();
addTo.ensureCapacity[IfPossible](values.length);
for (T element : values) { // no more iterator allocation because we're iterating over an array
    Preconditions.checkNotNull(element, "elementsToAdd cannot contain null elements");
    addTo.add(element);
}
```

PR #2536, opened 2025-05-13
Add a Bytes.Builder to avoid array copy

jonsyu1:
## Before this PR
The only way to construct `conjure.lib.Bytes` is to make a byte array copy.

## After this PR
==COMMIT_MSG==
Add a Bytes.Builder to avoid array copy
==COMMIT_MSG==

`Bytes.Builder` is inspired by guava's `ImmutableList.Builder` and `ByteArrayOutputStream`. It's purpose is to allow users to safely create their byte array without having to copy it once they're ready to produce `Bytes`.

## Possible downsides?
Users who do not know how much they want to write could end up making several allocations to reach the desired byte array size.

I see no existing limits on the maximum size of the `Bytes` byte array. I think it's reasonable to add a stricter limit on the builder, but it may be confusing for one construction method to have limits when others do not.

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
@@ -24,11 +24,14 @@ import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.palantir.logsafe.Preconditions;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import java.util.Objects;
 
 /** An immutable {@code byte[]} wrapper. */
 @JsonSerialize(using = Bytes.Serializer.class)
@@ -114,6 +117,14 @@ public final class Bytes {
         return new Bytes(safe);
     }
 
+    public static Builder builder(int capacity) {
+        return new Builder(capacity);
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
     static final class Serializer extends JsonSerializer<Bytes> {
         @Override
         public void serialize(Bytes value, JsonGenerator gen, SerializerProvider _serializer) throws IOException {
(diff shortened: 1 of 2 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
```diff
@@ -128,4 +139,56 @@ public Bytes deserialize(JsonParser parser, DeserializationContext _ctxt) throws
...
+            Preconditions.checkState(!isBuilt, "No writes are allowed because Immutable Bytes have already been built");
+            Objects.checkFromIndexSize(off, len, bytes.length);
+            ensureCapacity(count + len);
+            System.arraycopy(bytes, off, contents, count, len);
+            count += len;
+        }
+
+        public Bytes build() {
+            close();
+            return new Bytes(contents);
```
pkoenig10, line 177: This needs to be resized to remove uninitialized values.

Unless you know the exact size of the byte array ahead of time, I don't think we can avoid a copy here.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
```diff
@@ -128,4 +139,56 @@ public Bytes deserialize(JsonParser parser, DeserializationContext _ctxt) throws
...
+
+        public Builder() {
+            this(32);
+        }
+
+        public Builder(int capacity) {
+            Preconditions.checkArgument(capacity > 0);
+            this.contents = new byte[capacity];
+            this.count = 0;
+        }
```
pkoenig10, line 156: Constructors should be private

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
```diff
@@ -128,4 +139,56 @@ public Bytes deserialize(JsonParser parser, DeserializationContext _ctxt) throws
...
+        private byte[] contents;
+        private boolean isBuilt;
+        private int count;
+
+        public Builder() {
+            this(32);
+        }
+
+        public Builder(int capacity) {
+            Preconditions.checkArgument(capacity > 0);
```
pkoenig10, line 153: 0 is a valid size

conjure-lib/src/main/java/com/palantir/conjure/java/lib/Bytes.java
jonsyu1: A more broadly applicable approach may be to define a `from(ByteArrayOutputStream)`. `ByteArrayOutputStream::toByteArray` makes a copy via `Arrays::copyOf` so this returns an immutable copy.

pkoenig10: This wouldn't allow you do to anything that you couldn't do with the existing API.

jonsyu1: Sure, I can extend `ByteArrayOutputStream` to use the `protected` fields `buf[]` and `count` to construct `Bytes::from(byte[] array, int offset, int length)`. I'll do that for now.

PR #2548, opened 2025-06-16
Upgrade Eclipse Collections to v13, With Boxed Collection Support

bmarcaur:
This PR originally started out to test version `12.0` in support of https://github.com/eclipse-collections/eclipse-collections/issues/1737.

This bumps us to the latest version, `13.0` so that we can leverages some of the primitive optimized collections (e.g. `MutableDoubleList`). I was evaluating the likelihood of us addressing an improvement in our backlog which hinges on the availability of some new types introduced in the `v12+` RC's (specifically types like `BoxedMutableDoubleList`).

This _slightly_ reduces our custom boxed types. Although, as described in https://github.com/eclipse-collections/eclipse-collections/issues/1737#issuecomment-2978600032, I'll inline: 

One thing that I am noticing though is that `MutableIntCollection` has primitive overloads for `add()` (see code below) yet utilities like `BoxedMutableIntList`, which encloses a delegate `MutableIntCollection`, does not expose these sorts of functions. This prevents addition from being "non-boxing". We implemented a `BoxMutable*` like wrappers but then expanded it to [expose "non-boxing" paths](https://github.com/palantir/conjure-java/blob/640dbac596b5c9d7c3429ca790592e9a25141c5c/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java#L36-L42). This isn't feedback on the 12.0 _release_ necessarily, but it is something I noticed when upgrading. It may also be that this is an intentional narrowing of the interfaces.

```java
public interface MutableIntCollection extends IntIterable
{

    boolean add(int element);

    boolean addAll(int... source);

```

We added some optimizations (e.g. https://github.com/palantir/conjure-java/pull/2431 and https://github.com/palantir/conjure-java/pull/2486) in the interim between when we originally scoped out the desire to use these types and when they were released. Which relies on us having an optimized path through this type into the underlying Eclipse collection.

Either way this seems like nothing is lost and we have slightly less code to own.


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -169,12 +169,6 @@ public static <T> Set<T> newNonNullSet(Iterable<? extends T> iterable) {
         return set;
     }
 
-    /**
-     * The following Conjure boxed list wrappers for the eclipse-collections [type]ArrayList are temporary (except
-     * ConjureSafeLongList). In eclipse-collections 12, a BoxedMutable[type]List will be released. Once available,
-     * Conjure[type]List should be replaced with that.
-     */
-
```
bmarcaur, line 177: I'll reiterate inline here.

> We added some optimizations (e.g. https://github.com/palantir/conjure-java/pull/2431 and https://github.com/palantir/conjure-java/pull/2486) in the interim between when we originally scoped out the desire to use these types and when they were released. Which relies on us having an optimized path through this type into the underlying Eclipse collection.

Meaning these comments are out of date. Especially the Jackson optimization I think that will always need to be inline with the class until we have Project Valhalla complete.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
bmarcaur, line 1: Almost all of the removals in this class were simply hoisted into the newly available types `BoxedMutable*List`.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureIntegerList.java
```diff
@@ -17,74 +17,26 @@
...
-    }
-
-    @Override
-    public Integer set(int index, Integer element) {
-        return delegate.set(index, element);
-    }
-
-    ConjureIntegerList asUnmodifiable() {
-        return new ConjureIntegerList(delegate.asUnmodifiable());
-    }
```
aldexis, line 86: The eclipse-collections of `asUnmodifiable` seem to return a view wrapping the existing list. I don't imagine we consider this a problem that it would hold the original list (rather than its delegate `MutableIntList`) and prevent it from being garbage-collected?

aldexis (approved):
Double-checked the various methods for their hoisted variants and the code is pretty much identical, so this looks fine to me 👍

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureIntegerList.java
bmarcaur: I dont think so? They are all mostly behavior interfaces wrapping the underlying dense storage. Similar to how `ConjureIntegerList` was a wrapper over `UnmodifiableIntList`.

---

To run this to ground.

`AbstractMutableList` has an implementation of:

```java
public MutableList<T> asUnmodifiable() {
    return UnmodifiableMutableList.of(this);
}
```

And `UnmodifiableMutableList` delegates to the super type `AbstractUnmodifiableMutableCollection` which holds the real reference to `MutableCollection`.

---

Where as `IntArrayList` (the delegate type) has an implementation of:

```java
public MutableIntList asUnmodifiable()
{
    return new UnmodifiableIntList(this);
}
```

And `UnmodifiableIntList` delegates to the super type `AbstractUnmodifiableIntCollection` which holds the real reference to `MutableIntCollection`.

---

So realistically all we are losing is the primitive typing for some of the methods in that interface. Which is represented in the difference between `MutableList` and `MutableIntCollection`. But we were already wrapping it in a type that didn't further expose those underlying primitives.

```java
public static <T> List<T> unmodifiableList(List<T> list) {
    // Return the unmodifiable version of the Eclipse types
    if (list instanceof ConjureIntegerList conjureIntegerList) {
        return (List<T>) conjureIntegerList.asUnmodifiable();
    } else if (list instanceof ConjureDoubleList conjureDoubleList) {
        return (List<T>) conjureDoubleList.asUnmodifiable();
    } else if (list instanceof ConjureSafeLongList conjureSafeLongList) {
        return (List<T>) conjureSafeLongList.asUnmodifiable();
    } else {
        // Otherwise use the JDK types
        return Collections.unmodifiableList(list);
    }
}
```

So I don't think it changes the overhead, just different wrapping types, and does reduce a mild amount of code we own.

aldexis: Ah yes, that makes sense. In both cases, we're simply introducing a wrapper, just in a different location, so there is really just not extra overhead. All looks good then 👍

bmarcaur:
~Gonna mark as `do not merge` until I am back to monitor its rollout.~

PR #2549, opened 2025-06-17
Generate endpoint associated errors as `EndpointServiceException`s

mpritham:
## Before this PR
In #2401, endpoint errors were generated as `CheckedServiceExceptions` [(introduced here)](https://github.com/palantir/conjure-java-runtime-api/pull/1262). The choice to use checked exceptions has caused significant friction for clients of Conjure, particularly when using Java's functional APIs.

## After this PR
This change generates Conjure endpoint associated errors as `EndpointServiceExceptions` which are `RuntimeException`s.  The undertow interfaces are still generated with a `throws` clause, documenting each error that is associated the endpoint: this will help us in identifying the errors that an endpoint can throw when creating a static analysis check for detecting un-associated endpoint errors.

==COMMIT_MSG==
Generate endpoint associated errors as `EndpointServiceException`s
==COMMIT_MSG==

## Migration notes
`conjure-java` users that have adopted the feature already (which generated `CheckedServiceException`) should inspect their code-base for times they've handled each endpoint error. While it's not incorrect to handle `RuntimeException`s with `try-catch` blocks, they may have been added solely to be able to use the errors generated as checked exceptions - this is no longer needed.

## Possible downsides?
There is no longer any enforcement that errors associated with endpoints will only ever be associated with endpoints: an endpoint-associated error can be thrown from any endpoint regardless of it's association with that endpoint.


build.gradle
```diff
@@ -60,6 +60,7 @@ allprojects {
 
     repositories {
         mavenCentral() { metadataSources { mavenPom(); ignoreGradleMetadataRedirection() } }
+        mavenLocal()
```
bjlaub: reminder to remove this once https://github.com/palantir/conjure-java-runtime-api/pull/1352 merges.

conjure-java-core/src/integrationInput/java/dialogueendpointresulttypes/com/palantir/another/EndpointSpecificServerErrors.java
```diff
@@ -21,13 +21,13 @@ public static DifferentPackage differentPackage(@Nullable Throwable cause) {
      *
      * @param shouldThrow Cause the method to throw when true
      */
-    public static void throwIfDifferentPackage(boolean shouldThrow) throws DifferentPackage {
```
bjlaub, line 24: just curious, why does this `throws` disappear?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/EndpointErrorGenerator.java
```diff
@@ -165,7 +165,6 @@ private static List<MethodSpec> generateConditionalExceptionFactories(
                                     options,
                                     exceptionClassName,
                                     Optional.empty())
-                            .addException(exceptionClassName)
```
bjlaub, line 168: > The undertow interfaces are still generated with a throws clause

how does this relate to this statement? I was expecting to still see `throws` on methods in the generated integration input but it looks like they disappeared.

mpritham: > I was expecting to still see throws on methods in the generated integration input but it looks like they disappeared.

The undertow interfaces still have a `throws` clause:

https://github.com/palantir/conjure-java/blob/c8904d3c9f265741be1e33f965fb429d17bba4a5/conjure-java-core/src/integrationInput/java/undertow/com/palantir/product/UndertowErrorService.java#L23-L35

I've removed the `throws` clauses from the `throwIf` helper methods. These methods are used inside implementation code: I didn't want to add a `throws` statement here because clients would have to handle the thrown exception when they make a call to a `throwIf` method.

By keeping the `throws` clauses on the undertow interfaces, users who make calls to the undertow interface will need to handle the thrown exception as well: devs definitely do this, but I think this isn't all that common of a pattern, and the benefit of having the `throws` clause there (helps us the static analysis check know what errors can be thrown from an endpoint method) are worth it IMO.

bjlaub (approved):

PR #2564, opened 2025-07-09
Bump dropwizard dependency (and jersey-common to 3.x)

aldexis:
## Before this PR
Dropwizard is currently being used in tests, and its low version is blocking other dependency upgrades due to an incompatibility with newer junit versions: https://github.com/palantir/conjure-java/pull/2353

## After this PR

==COMMIT_MSG==
Bump dropwizard dependency (and jersey-common to 3.x)
==COMMIT_MSG==

Down the line, we might want to rewrite the tests to not rely on dropwizard at all, like was done in https://github.com/palantir/conjure-java-runtime/pull/2438 instead.

## Possible downsides?
This is bumping jersey to 3.x as well as moving a number of things to jakarta, but these should all have been done a while ago.


conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/TestCases.java
```diff
@@ -233,21 +233,25 @@ public final class TestCases {
                     .files(Path.of("ete-binary.yml"))
                     .options(Options.builder()
                             .requireNotNullAuthAndBodyParams(true)
+                            .jakartaPackages(true)
```
aldexis, line 236: Updating the tests to generate jakarta classes, so we can rip out the javax dependencies from our gradle

conjure-java-server-verifier/src/test/java/com/palantir/conjure/java/verification/server/undertest/JerseyServerUnderTestApplication.java
```diff
@@ -21,28 +21,27 @@
...
 
     @Override
-    public void initialize(Bootstrap<JerseyServerUnderTestConfiguration> bootstrap) {
+    public void initialize(Bootstrap<Configuration> bootstrap) {
         ObjectMapper remotingObjectMapper = ObjectMappers.newServerObjectMapper()
                 // needs discoverable subtype resolver for DW polymorphic configuration mechanism
                 .setSubtypeResolver(new DiscoverableSubtypeResolver())
                 .registerModule(new FuzzyEnumModule());
         bootstrap.setObjectMapper(remotingObjectMapper);
-        bootstrap.addBundle(new WebSecurityBundle());
```
aldexis, line 40: This seems to be an old thing which isn't compatible with newer dropwizard/jersey, and was archived 4 years ago: https://github.com/palantir/dropwizard-web-security

versions.props
```diff
@@ -42,14 +43,13 @@ org.mockito:* = 5.11.0
 org.slf4j:* = 2.0.17
 org.wildfly.common:wildfly-common = 1.7.0.Final
 jakarta.ws.rs:jakarta.ws.rs-api = 4.0.0
+jakarta.validation:jakarta.validation-api = 3.0.0
 
 # dependency-upgrader:OFF
-# Avoid forcing consumers to upgrade jackson
-com.fasterxml.jackson.*:jackson-* = 2.12.3
```
aldexis: This was outdated and effectively useless given the versions.lock pin was already bumped

kunalrkak (dismissed):
LGTM

aldexis:
@kunalrkak @bjlaub I was finally able to come back to this and test it on a couple of internal repos. As far as I can tell, this should be fine to merge and have virtually no impact anywhere. If you can give it a look and approve if it looks good, that would be great, as it will then unblock https://github.com/palantir/conjure-java/pull/2353

conjure-java-core/src/test/resources/test/api/TestService.java.jersey_require_not_null
```diff
@@ -18,7 +18,7 @@ import java.util.OptionalInt;
 import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.processing.Generated;
-import javax.validation.constraints.NotNull;
+import jakarta.validation.constraints.NotNull;;
```
bjlaub: nit: there's an extra semicolon here?

aldexis: Indeed 🤔 Removed it from all three files that had it (I think this might be a left-over from a bad search-replace I did at some point)

bjlaub (dismissed):

bjlaub:
👍

PR #2567, opened 2025-07-11
Deconflict `cause` conjure error arguments

aldexis:
## Before this PR
The conjure-java code generation for errors generates methods that take a `Throwable cause` parameter.

While [other places do parameter name deconfliction](https://github.com/palantir/conjure-java/blob/6ba3a53d2a04b2a1e65b0bf32f43e2c7ff799bb4/conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java#L62-L89), the generated code will fail to compile if a `cause` argument is added to the conjure error definition.

## After this PR

==COMMIT_MSG==
Deconflict `cause` conjure error arguments
==COMMIT_MSG==

You can see the reproduction of the problem in https://github.com/palantir/conjure-java/pull/2567/commits/a9ef2ac69539a6364b817503521d7f7592c25a4b with the fix and its changes in https://github.com/palantir/conjure-java/pull/2567/commits/b7f21c5824a6aec8da62ddaa53b1728613ae8626

### DIFF
--- a/conjure-java-core/src/integrationInput/java/dialogueendpointresulttypes/com/palantir/product/ConjureServerErrors.java
+++ b/conjure-java-core/src/integrationInput/java/dialogueendpointresulttypes/com/palantir/product/ConjureServerErrors.java
@@ -1,0 +1,38 @@
+package dialogueendpointresulttypes.com.palantir.product;
+
+import com.palantir.conjure.java.api.errors.EndpointServiceException;
+import com.palantir.logsafe.Safe;
+import com.palantir.logsafe.SafeArg;
+import javax.annotation.Nullable;
+import javax.annotation.processing.Generated;
+
+@Generated("com.palantir.conjure.java.types.EndpointErrorGenerator")
+public final class ConjureServerErrors {
+    private ConjureServerErrors() {}
+
+    public static ConflictingCauseSafeArg conflictingCauseSafeArg(@Safe String cause) {
+        return new ConflictingCauseSafeArg(cause, null);
+    }
+
+    public static ConflictingCauseSafeArg conflictingCauseSafeArg(@Safe String cause, @Nullable Throwable cause) {
+        return new ConflictingCauseSafeArg(cause, cause);
+    }
+
+    /**
+     * Throws a {@link ConflictingCauseSafeArg} when {@code shouldThrow} is true.
+     *
+     * @param shouldThrow Cause the method to throw when true
+     * @param cause
+     */
+    public static void throwIfConflictingCauseSafeArg(boolean shouldThrow, @Safe String cause) {
+        if (shouldThrow) {
+            throw conflictingCauseSafeArg(cause);
+        }
+    }
+
+    public static final class ConflictingCauseSafeArg extends EndpointServiceException {
+        private ConflictingCauseSafeArg(@Safe String cause, @Nullable Throwable cause) {
+            super(ConjureErrors.CONFLICTING_CAUSE_SAFE_ARG, cause, SafeArg.of("cause", cause));
+        }
+    }
+}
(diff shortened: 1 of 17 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
```diff
@@ -88,6 +90,11 @@ public String visitUnknown(String unknownType) {
         return value;
     }
 
+    /** Sanitizes parameter names for errors. */
+    public static String sanitizeErrorParameterName(String input) {
+        return SourceVersion.isName(input) && !RESERVED_ERROR_PARAMETER_NAMES.contains(input) ? input : escape(input);
```
mpritham, line 95: Don't think `SourceVersion.isName` is required. JavaPoet already does this check when[ we create a parameter.](https://github.com/palantir/javapoet/blob/c1ea09d451b7e82c13967e702a8dcb6684c06a44/javapoet/src/main/java/com/palantir/javapoet/ParameterSpec.java#L132-L145)

```suggestion
        return RESERVED_ERROR_PARAMETER_NAMES.contains(input) ? escape(input) : input;
```

conjure-java-core/src/integrationInput/java/dialogueendpointresulttypes/com/palantir/product/ConjureServerErrors.java
```diff
@@ -0,0 +1,38 @@
...
+        return new ConflictingCauseSafeArg(cause_, cause);
+    }
+
+    /**
+     * Throws a {@link ConflictingCauseSafeArg} when {@code shouldThrow} is true.
+     *
+     * @param shouldThrow Cause the method to throw when true
+     * @param cause
+     */
+    public static void throwIfConflictingCauseSafeArg(boolean shouldThrow, @Safe String cause) {
```
mpritham: We should also add `shouldThrow` to the list of reserved error parameter names.

aldexis: Nice catch!

conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
aldexis: Doesn't it [throw](https://github.com/palantir/javapoet/blob/c1ea09d451b7e82c13967e702a8dcb6684c06a44/javapoet/src/main/java/com/palantir/javapoet/ParameterSpec.java#L143) if `SourceVersion.isName` is false?

Fwiw, I'm not sure that `escape(input)` would make it return true, but I mostly pattern-matched from the existing other sanitization logic.

conjure-java-core/src/integrationInput/java/dialogueendpointresulttypes/com/palantir/product/ConjureServerErrors.java
aldexis: Fixed to also deconflict on `shouldThrow` (and added to the test cases)

conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
mpritham: `escape` appends an underscore - I think it wouldn't make the variable valid, and this will end up throwing when we construct the parameter.

I think when this throws later, it'd be slightly confusing to see the escaped name instead of just the offending name.

aldexis: Isn't that also true of any of the other cases above? Or is there something I'm missing that makes this unique to this case? If not, I'd prefer to keep the code consistent, and follow-up with a PR to clean this up.

mpritham: It's true for the other cases above as well. Sure, happy to merge as-is.

mpritham (approved):
Thanks!

conjure-java-core/src/main/java/com/palantir/conjure/java/util/JavaNameSanitizer.java
aldexis: Sweet - will follow-up with another refactor

PR #2577, opened 2025-07-17
Add `dangerousDoNotUseEnableEndpointAssociatedErrors` option to gate endpoint error generation

mpritham:
## Before this PR
When errors are associated with endpoints, they are generated as subtypes of `EndpointServiceException`. The parameters of these exceptions are serialized as JSON, while previously error parameters were serialized with `Object.toString`. This wire break can cause failures when clients parse error parameters expecting the previous `Object.toString` serialization format. 

We'd like to **temporarily** stop supporting usage of the endpoint associated error feature and build some tooling to enable migration of _all_ error parameters to JSON.

## After this PR

==COMMIT_MSG==
Add `dangerousEnableEndpointAssociatedErrors` option to gate endpoint
==COMMIT_MSG==

## Rollout
Currently, endpoint associated error generation is supported. Initially, the value of the `dangerousEnableEndpointAssociatedErrors` flag is set to true, matching the current behavior. In a future PR, after updating the consumers of this feature to set the flag to `true` explicitly, I will change the default value to `false`.

Eventually, I'd like to remove this flag once we have migrated all error parameters to be serialized as JSON.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
@@ -203,10 +203,17 @@ public interface Options {
     }
 
     /**
-     * Enabling this option will cause parameters of errors that are associated with an endpoint to be serialized as
-     * JSON. This is a wire break and can cause errors for clients that parse error parameters. Do not enable this.
+     * Enabling this option will enable associating errors with endpoints, which currently causes parameters of these
+     * errors to be serialized as JSON. This is a wire break and can cause errors for clients that parse error
+     * parameters. Do not enable this.
      * <p>
-     * For rollout, this value is initially set to true. In the future, this value will be set to false.
+     * This flag has been temporarily added to help us iterate on a safer rollout strategy for endpoint-associated
+     * errors.
+     * <p>
+     * It is initially set to true, but will quickly be set to false.
+     * <p>
+     * It will also eventually be removed, and you should not use it without letting the Conjure maintainers know.
+     * Otherwise, a future upgrade will break your builds.
      */
     @Value.Default
     default boolean dangerousDoNotUseEnableEndpointAssociatedErrors() {
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
@@ -209,7 +209,7 @@ public interface Options {
      * For rollout, this value is initially set to true. In the future, this value will be set to false.
      */
     @Value.Default
-    default boolean dangerousEnableEndpointAssociatedErrors() {
+    default boolean dangerousDoNotUseEnableEndpointAssociatedErrors() {
         return true;
     }
 
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
@@ -203,6 +203,17 @@ public interface Options {
     }
 
     /**
+     * Enabling this option will cause parameters of errors that are associated with an endpoint to be serialized as
+     * JSON. This is a wire break and can cause errors for clients that parse error parameters. Do not enable this.
+     * <p>
+     * For rollout, this value is initially set to true. In the future, this value will be set to false.
+     */
+    @Value.Default
+    default boolean dangerousEnableEndpointAssociatedErrors() {
+        return true;
+    }
+
+    /**
      * Makes immutable copies of collections for union and alias types, respecting the --nonNullCollections flag.
      */
     @Value.Default
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/EndpointErrorGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/EndpointErrorGenerator.java
@@ -62,7 +62,7 @@ public final class EndpointErrorGenerator implements Generator {
         SafetyEvaluator safetyEvaluator = new SafetyEvaluator(types);
         TypeMapper typeMapper = new TypeMapper(types, options);
         DeclaredEndpointErrors endpointErrors = DeclaredEndpointErrors.from(definition);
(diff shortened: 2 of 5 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
```diff
@@ -202,6 +202,17 @@ default boolean generateDialogueEndpointErrorResultTypes() {
         return false;
     }
 
+    /**
+     * Enabling this option will cause parameters of errors that are associated with an endpoint to be serialized as
+     * JSON. This is a wire break and can cause errors for clients that parse error parameters. Do not enable this.
+     * <p>
+     * For rollout, this value is initially set to true. In the future, this value will be set to false.
```
kunalrkak: Mind adding a note indicating that this option is temporary and will be removed in a future release?

kunalrkak (dismissed):

conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
mpritham: Makes sense, will add.

aldexis: Proposal on a slightly more explicit documentation:
```suggestion
     * Enabling this option will enable associating errors with endpoints, which currently causes parameters of these errors to be serialized as
     * JSON. This is a wire break and can cause errors for clients that parse error parameters. Do not enable this.
     * <p>
     * This flag has been temporarily added to help us iterate on a safer rollout strategy for endpoint-associated
     * errors.
     * <p>
     * It is initially set to true, but will quickly be set to false.
     * <p>
     * It will also eventually be removed, and you should not use it without letting the Conjure maintainers know.
     * Otherwise, a future upgrade will break your builds.
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/EndpointErrorGenerator.java
```diff
@@ -55,10 +58,18 @@ public EndpointErrorGenerator(Options options) {
 
     @Override
     public Stream<JavaFile> generate(ConjureDefinition definition) {
+        DeclaredEndpointErrors endpointErrors = DeclaredEndpointErrors.from(definition);
+        if (!options.dangerousDoNotUseEnableEndpointAssociatedErrors()
+                && !endpointErrors.errors().isEmpty()) {
+            List<String> errorNames =
+                    endpointErrors.errors().stream().map(ErrorTypeName::getName).toList();
+            throw new SafeIllegalStateException(
+                    "Errors are associated with endpoints. This feature is currently not supported.",
```
aldexis: Proposal on a slightly more explicit message, to help with potential OSS consumers that would have already enabled this
```suggestion
                    "Errors are associated with endpoints. This feature has been temporarily disabled while we iterate through issues with its rollout. If this is blocking you or you need this immediately, please reach out to the Conjure maintainers at https://github.com/palantir/conjure-java/issues so we can help support your needs.",
```

conjure-java-core/src/test/java/com/palantir/conjure/java/types/EndpointErrorGeneratorTests.java
```diff
@@ -0,0 +1,89 @@
...
+
+        // Verify that an exception is thrown, and that it contains the name of the error associated with the endpoint
+        assertThatThrownBy(() -> generator.generate(DEFINITION))
+                .isInstanceOf(SafeIllegalStateException.class)
+                .hasMessageContaining("Errors are associated with endpoints. This feature is currently not supported.")
+                .hasMessageContaining(ERROR_NAME);
+    }
+
+    @Test
+    public void testNoExceptionThrownWhenOptionIsTrueAndEndpointErrorsExist() {
```
aldexis, line 101: nit: we should also have a test that no exception is thrown if option is false, but there are no endpoint errors

conjure-java-core/src/test/java/com/palantir/conjure/java/types/EndpointErrorGeneratorTests.java
```diff
@@ -0,0 +1,89 @@
...
+    public void testThrowsExceptionWhenOptionIsFalseAndEndpointErrorsExist() {
+        // Create an EndpointErrorGenerator with the option set to false
+        EndpointErrorGenerator generator = new EndpointErrorGenerator(Options.builder()
+                .dangerousDoNotUseEnableEndpointAssociatedErrors(false)
+                .build());
+
+        // Verify that an exception is thrown, and that it contains the name of the error associated with the endpoint
+        assertThatThrownBy(() -> generator.generate(DEFINITION))
+                .isInstanceOf(SafeIllegalStateException.class)
+                .hasMessageContaining("Errors are associated with endpoints. This feature is currently not supported.")
```
aldexis: vnit: You could make the message a static const on the `EndpointErrorGenerator` class to reuse it here, rather than copying the message

aldexis (commented):
We're going to need to pipe this through gradle-conjure as well, right? (or do we have some other way we could set it without changes there?

conjure-java-core/src/test/java/com/palantir/conjure/java/types/EndpointErrorGeneratorTests.java
mpritham: Added. fwiw copilot also suggested this :D

kunalrkak (dismissed):

mpritham:
> We're going to need to pipe this through gradle-conjure as well, right? (or do we have some other way we could set it without changes there?

We don't need to update `gradle-conjure`. The options are not listed or stored statically in `gradle-conjure` - the Conjure CLI's `generate` command is run with the arguments set from the Gradle options block ([reading this src](https://github.com/palantir/gradle-conjure/blob/f9c3fe6eb9fda832ce7c299ef9703990a7bbcc5c/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureGeneratorTask.java#L105-L114)).

kunalrkak (approved):

PR #2588, opened 2025-07-29
Respect `Accept-Conjure-Error-Parameter-Format`

mpritham:
## Before this PR
Need to merge the [PR the introduces the Accept-Conjure-Error-Parameter-Format header first](https://github.com/palantir/conjure-java-runtime-api/pull/1398).

Currently, error parameters are serialized using `Objects.toString`. We'd like to enable clients to receive JSON parameters as well. In order to safely migrate error parameter serialization formats, clients should be able to specify the parameter format they expect from servers.

## After this PR

==COMMIT_MSG==
When the `Accept-Conjure-Error-Parameter-Format`  header is `JSON`, servers serialize error parameters as JSON, instead of using `Objects.toString`.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowEteResource.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowEteResource.java
@@ -21,6 +21,7 @@ import com.palantir.conjure.java.undertow.lib.BinaryResponseBody;
 import com.palantir.ri.ResourceIdentifier;
 import com.palantir.tokens.auth.AuthHeader;
 import com.palantir.tokens.auth.BearerToken;
+import errors.com.palantir.product.ConjureErrors;
 import jakarta.validation.constraints.NotNull;
 import jakarta.ws.rs.core.StreamingOutput;
 import java.nio.charset.StandardCharsets;
@@ -170,6 +171,14 @@ public final class UndertowEteResource implements UndertowEteService {
     @Override
     public SimpleEnum enumHeader(AuthHeader _authHeader, SimpleEnum headerParameter) {
         return headerParameter;
+    }
+
+    @Override
+    public String jsonErrorsHeader(AuthHeader authHeader, String headerParameter) {
+        if (headerParameter.equals("JSON") || headerParameter.equals("TOSTRING")) {
+            throw ConjureErrors.invalidServiceDefinition("my-service-string", Optional.of(SimpleEnum.VALUE));
+        }
+        return "hello!";
     }
 
     @Override
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowEteResource.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowEteResource.java
@@ -175,10 +175,7 @@ public final class UndertowEteResource implements UndertowEteService {
 
     @Override
     public String jsonErrorsHeader(AuthHeader authHeader, String headerParameter) {
-        if (headerParameter.equals("JSON") || headerParameter.equals("TOSTRING")) {
-            throw ConjureErrors.invalidServiceDefinition("my-service-string", Optional.of(SimpleEnum.VALUE));
-        }
-        return "hello!";
+        throw ConjureErrors.invalidServiceDefinition("my-service-string", Optional.of(SimpleEnum.VALUE));
     }
 
     @Override
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
@@ -18,6 +18,9 @@ package com.palantir.conjure.java.undertow.runtime;
 
 import com.google.common.util.concurrent.RateLimiter;
 import com.palantir.conjure.java.api.errors.CheckedServiceException;
+import com.palantir.conjure.java.api.errors.ConjureErrorParameterFormat;
+import com.palantir.conjure.java.api.errors.ConjureErrorParameterFormats;
+import com.palantir.conjure.java.api.errors.ConjureErrorParameterFormats.ConjureErrorParameterFormatRequestDecodingAdapter;
 import com.palantir.conjure.java.api.errors.EndpointServiceException;
 import com.palantir.conjure.java.api.errors.ErrorType;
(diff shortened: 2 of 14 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -115,11 +119,38 @@ private static void endpointServiceException(HttpServerExchange exchange, Endpoi
...
 
     private static void serviceException(HttpServerExchange exchange, ServiceException exception) {
-        log(exception);
+        Optional<ConjureErrorParameterFormat> maybeErrorParamFormatHeader =
+                ConjureErrorParameterFormats.parseFromRequest(
+                        exchange.getRequestHeaders(), ConjureErrorParamsDecoder.INSTANCE);
+
+        boolean isHeaderPresent = maybeErrorParamFormatHeader.isPresent();
+        boolean isJsonErrorParameterValueFormat =
+                isHeaderPresent && maybeErrorParamFormatHeader.get().equals(ConjureErrorParameterFormat.JSON_FORMAT);
```
bjlaub: thoughts on using `equalsIgnoreCase` here? i can see `json` and `JSON` being easy to mix up.

or perhaps encapsulating this in `ConjureErrorParameterFormats` directly? e.g.
```
public static <T> boolean isJsonFormat(T request, ConjureErrorParameterFormatRequestDecodingAdapter<? super T> adapter) {
    return parseFromRequest(request, adapter).map(v -> v.equalsIgnoreCase(ConjureErrorParameterFormat.JSON_FORMAT).orElse(false);
}
```

bjlaub (dismissed):
approving for RC

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
mpritham: Makes sense. I just updated the current logic.

bjlaub (dismissed):
approving for RC

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowEteResource.java
```diff
@@ -172,6 +173,14 @@ public SimpleEnum enumHeader(AuthHeader _authHeader, SimpleEnum headerParameter)
         return headerParameter;
     }
 
+    @Override
+    public String jsonErrorsHeader(AuthHeader authHeader, String headerParameter) {
+        if (headerParameter.equals("JSON") || headerParameter.equals("TOSTRING")) {
```
aldexis: nit: let's just throw if the header is set? (or even just always throw?)
```suggestion
        if (!headerParameter.isEmpty()) {
```

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -115,11 +119,41 @@ private static void endpointServiceException(HttpServerExchange exchange, Endpoi
...
+            log.info(
+                    "Unrecognized Conjure error parameter format header",
+                    SafeArg.of("headerValue", maybeErrorParamFormatHeader.get()));
+        }
+
+        // Use the appropriate serialization based on format
+        logWithSerializationFormat(exception, isJsonErrorParameterValueFormat);
+        ConjureError conjureError = isJsonErrorParameterValueFormat
+                ? ConjureError.fromServiceExceptionWithJsonSerializedParameterValues(exception)
+                : ConjureError.fromServiceException(exception);
```
aldexis, line 144: If we had the stringified format too, we could do something like:
```suggestion
        ConjureErrorParameterFormat errorParamFormatHeader =
                ConjureErrorParameterFormats.parseFromRequest(
                        exchange.getRequestHeaders(), ConjureErrorParamsDecoder.INSTANCE)
                        .orElse(DEFAULT_FORMAT);

        // Log unrecognized format if present but not JSON
        if (ConjureErrorParameterFormats.isKnown(errorParamFormatHeader)) {
            log.info(
                    "Unrecognized Conjure error parameter format header - defaulting to {}",
                    SafeArg.of("headerValue", errorParamFormatHeader.get()),
                    SafeArg.of("default", DEFAULT_FORMAT));
            errorParamFormatHeader = DEFAULT_FORMAT;
        }
        
        boolean useJsonErrorParameterValueFormat = errorParamFormatHeader
                        .equalsIgnoreCase(ConjureErrorParameterFormat.JSON_FORMAT);

        // Use the appropriate serialization based on format
        logWithSerializationFormat(exception, useJsonErrorParameterValueFormat);
        ConjureError conjureError = useJsonErrorParameterValueFormat
                ? ConjureError.fromServiceExceptionWithJsonSerializedParameterValues(exception)
                : ConjureError.fromServiceException(exception);
```
with `DEFAULT_FORMAT` being `ConjureErrorParameterFormat.STRING_FORMAT` (and potentially passed down as a parameter?)

aldexis (dismissed):

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
mpritham: I think we can add the `DEFAULT_FORMAT` later on if we find that folks want to keep using `Objects.toString` parameters. Right now, if there is no value provided for the header, the default will be the `Objects.toString` parameters. Folks can undo the change (if they find there is some issue with using the JSON params) by unsetting the flag on their dialogue clients.

I think introducing the `DEFAULT_FORMAT` header as an option for devs will push them towards using it, and will make our migration harder to roll out because we'd have to track down all usages of `DEFAULT_FORMAT` and push teams towards the `JSON` format.

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowEteResource.java
mpritham: Updated to always throw. Initially I thought I'd test that we're able to get the standard return type from this endpoint, but this can be handled with other tests and doesn't relate to this PR.

bjlaub (approved):

PR #2608, opened 2025-08-12
Support generation of Dialogue interfaces that throw `RemoteException` subclasses with rich parameter types

mpritham:
## Before this PR

Error parameters from `ServiceException`s are serialized with `Objects.toString`. This is difficult to parse: clients must write code that understands the details of Java's `toString` method. 

Endpoint associated errors' parameters are serialized as JSON. While useful, associating errors with endpoints on a server immediately changes the error parameter serialization format for all clients. Clients that have relied on custom error parameter parsing logic can fail. Previously, when introducing endpoint associated errors and serializing their parameters as JSON, we had considered this a reasonable break that adopting servers would be responsible for rolling out. The current stance is that asking servers to track down all their client usage and ensure the migration is safe is not reasonable. 

We'd like to create a _safe_ migration path towards using JSON serialization for all error parameters (meaning: servers will not break clients). In the long term, we'd like to not use `Objects.toString` parameters at all for `ServiceException`s.

## After this PR
When the `generateErrorParameterFormatRespectingDialogueInterfaces` flag is set to `true`:

- Conjure will generate Dialogue interfaces:
  - that will populate the `Accept-Conjure-Error-Parameter-Format` header with the value of `bodySerDe`'s `errorParameterFormat` field when present
  - which have endpoint deserializers that:
    -  will create custom exceptions when the error parameter payload is JSON. These exceptions will be subtypes of `RemoteException` whose `SerializableError` field will be populated so existing client code will continue to work as expected.
    - will fall back to creating `RemoteException` when the client cannot deserialize the error parameter payload.
      - flag: it's possible that the error parameter payload's `toString` representation matches the JSON representation (e.g. no parameters, only an integer parameter, .etc.). In this case, the new exception type will be created.
- for each `ServiceException` Conjure will generate a record for the parameters of the error, a class representing the serialized error, and a class which is a sub-type of `RemoteException` for use in the Dialogue interfaces described above.

For example, if we had an error defined like:
```
MyError:
  namespace: Conjure
  code: INTERNAL
  safe-args:
    optionalParameter: optional<string>
```

A client could add logic to their code that handles remote exceptions like so:

```diff
// Initialize client to set the errorParameterFormat to JSON
try {
    client.call(...);
} catch (RemoteException e) {
    Optional<String> maybeValue = optionalParsingHelper(e.getError().parameters().get("optionalParameter"));
    // Do something with value

+    if (e.getCause() instanceof MyErrorException myErrorException) {
+        Optional<String> maybeValueFromNewException =
+                myErrorException.error().parameters().optionalParameter();
+        // Do something with maybeValue
+    }
}
```

Note: clients are still required to handle `RemoteException`. 1) Errors other than those arising from servers throwing a Conjure defined error can be surfaced as `RemoteException`, 2) Clients could be hitting servers that may not be able to understand the `Accept-Conjure-Error-Parameter-Format` header. 

==COMMIT_MSG==
Support generation of Dialogue interfaces that throw `RemoteException` subclasses with rich parameter types
==COMMIT_MSG==

## Possible downsides?
The `generateErrorParameterFormatRespectingDialogueInterfaces` is not strictly required: our deserializers are backward compatible. I'd still like to flag the generation of clients that use these new deserializers to have finer control over roll out, being extra cautious of any implementation bugs as these are critical code paths affecting every endpoint.

We're generating more code for every `ServiceException` (when the flag is set).


conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/TestCases.java
```diff
@@ -176,7 +176,10 @@ public final class TestCases {
                     .files(Path.of("ete-service.yml"))
                     .files(Path.of("ete-binary.yml"))
                     .files(Path.of("example-errors.yml"))
-                    .options(Options.builder().apiVersion("1.2.3").build())
+                    .options(Options.builder()
+                            .apiVersion("1.2.3")
+                            .generateErrorParameterFormatRespectingDialogueInterfaces(true)
```
mpritham, line 191: By enabling this flag here, instead of creating a new test case, we are no longer testing the code gen of the current Dialogue interfaces with this test.

I will add a separate test case before merging. I didn't want to do this while we're initially reviewing and cutting RCs: there are loads of files that are generated, which makes reviewing a bit more tedious.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
```diff
@@ -161,18 +172,129 @@ private JavaFile generateErrorTypesForNamespace(
...
+                .addModifiers(Modifier.PUBLIC, Modifier.STATIC, Modifier.FINAL)
+                .superclass(
+                        ParameterizedTypeName.get(ClassName.get(AbstractSerializableError.class), parametersClassName))
+                .addMethod(constructor)
+                .addMethod(toSerializableErrorBuilder.build());
+
+        return classBuilder.build();
+    }
+
     private static List<TypeSpec> generateErrorParameterRecords(
```
mpritham, line 301: Note to self: when the record fields are reserved words like "cause" or "shouldThrow", we use `cause_` and `shouldThrow_`. But we should expose the record field as "cause"/"shouldThrow".

conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandlerTest.java
```diff
@@ -138,13 +138,15 @@ static Stream<Arguments> differentPackageExceptions() {
                 Arguments.of((Supplier<Exception>) () -> new DifferentPackageEndpointException(null)));
     }
 
+    @SuppressWarnings("removal")
```
mpritham, line 141: This should be removed when we remove the `CheckedServiceException`. The constructors of `CheckedServiceException` have been deprecated.

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -246,6 +246,16 @@ public static final class GenerateCommand implements Runnable {
                         + "This feature is currently not supported.")
         private boolean dangerousDoNotUseEnableEndpointAssociatedErrors;
 
+        @CommandLine.Option(
+                names = "--generateErrorParameterFormatRespectingDialogueInterfaces",
+                defaultValue = "false",
+                description =
+                        "Enabling this flag will lead to the generation of Dialogue client interfaces that are able to"
+                            + " send a header to servers specifying the Conjure error parameter serialization format"
+                            + " they expect, and create custom exceptions foreach error defined in the service's"
```
aldexis:
```suggestion
                            + " they expect, and create custom exceptions for each error defined in the service's"
```

conjure-java-core/src/integrationInput/java/undertow/com/palantir/product/ErrorServiceEndpoints.java
```diff
@@ -60,7 +60,7 @@ private static final class TestBasicErrorEndpoint implements HttpHandler, Endpoi
 
         @Override
         public void handleRequest(HttpServerExchange exchange)
-                throws IOException, TestServerErrors.InvalidArgument, ConjureServerErrors.ConflictingCauseSafeArg {
+                throws IOException, TestServerErrors.InvalidArgument, ConjureServerErrors.ConflictingCauseSafeArgErr {
```
aldexis, line 63: Not quite related to this PR, but I'm actually surprised we still generate the throws statement here, since these are not checked exceptions (I didn't realize we could even do this for unchecked exceptions). Is this because we were looking to have static analysis verify the thrown errors?

conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
```diff
@@ -202,6 +202,17 @@ default boolean dangerousDoNotUseEnableEndpointAssociatedErrors() {
...
+    /**
+     * Enabling this flag will lead to the generation of Dialogue client interfaces that are able to send a header to
+     * servers specifying the Conjure error parameter serialization format they expect, and create custom exceptions for
+     * each error defined in the service's package.
+     */
+    @Value.Default
+    @Beta
+    default boolean generateErrorParameterFormatRespectingDialogueInterfaces() {
+        return false;
+    }
```
aldexis, line 214: If I understand/remember things correctly, the rollout strategy is:
- Enabling this will make it generate dialogue clients that _can_ deserialize both legacy and new formats for errors, but will _not_ send the header by default
- Once it's enabled on a project producing dialogue clients, the consumers _can_ be updated to send the new header (but won't by default)
- We're going to roll things out such that both are done, but this gives us some way to safely roll out the new dialogue clients, as well as its adoption, and two escape hatches (per-service and per-client) if things go wrong?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -127,6 +142,45 @@ public MethodSpec generate(ServiceDefinition def) {
         return method;
     }
 
+    private MethodSpec createHelperToConstructExceptionDeserializerArgs(String serviceDefinitionPackageName) {
+        List<ErrorDefinition> errorDefinitions = packageToErrorDefinitionsMapping.get(serviceDefinitionPackageName);
```
aldexis: Mentioned in the tests PR, but I'm not sure we should limit ourselves to the errors from the same package as the service? (if anything, I could see devs adding the errors under a specific `.errors` package)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -127,6 +142,45 @@ public MethodSpec generate(ServiceDefinition def) {
...
+
+        CodeBlock.Builder exceptions = CodeBlock.builder()
+                .add("return $T.<$T>builder()", ExceptionDeserializerArgs.class, typeVariableT)
+                .add(".returnType(returnType)");
+        // Add exceptions from error definitions
+        for (ErrorDefinition errorDef : errorDefinitions) {
+            String errorName = errorDef.getErrorName().getName();
+            ClassName errorClass = ClassName.get(
+                    Packages.getPrefixedPackage(errorDef.getErrorName().getPackage(), options.packagePrefix()),
+                    ErrorGenerationUtils.errorTypesClassName(errorDef.getNamespace()),
```
aldexis: style nit: these are reused, so let's extract (might also consider moving the various `ClassName` construction pieces to helper methods within `ErrorDefinition`?)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -221,14 +294,27 @@ private MethodSpec clientImpl(EndpointDefinition def) {
...
                                         : ".bodySerDe().inputStreamDeserializer()"))
                         .orElseGet(() -> def.getEndpointName().get() + "Deserializer"));
 
         methodBuilder.addCode(request);
+        if (options.generateErrorParameterFormatRespectingDialogueInterfaces()) {
+            // If the BodySerDe.errorParameterFormat field is set, add the header
+            methodBuilder.addCode(CodeBlock.builder()
+                    .beginControlFlow("if ($L.bodySerDe().errorParameterFormat().isPresent())", RUNTIME)
+                    .add(
+                            "$L.putHeaderParams($S," + " $L.bodySerDe().errorParameterFormat().get().toString());",
```
aldexis:
```suggestion
                            "$L.putHeaderParams($S, $L.bodySerDe().errorParameterFormat().get().toString());",
```

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -221,14 +294,27 @@ private MethodSpec clientImpl(EndpointDefinition def) {
...
+                            REQUEST,
+                            "Accept-Conjure-Error-Parameter-Format", // TODO(pm): make this public.
```
aldexis: Specifically flagging your TODO here (I imagine this is a constant in dialogue already? not sure which PR is adding it and it doesn't seem to be on develop)

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowEteResource.java
```diff
@@ -178,6 +197,109 @@ public String jsonErrorsHeader(AuthHeader authHeader, String headerParameter) {
         throw ConjureErrors.invalidServiceDefinition("my-service-string", Optional.of(SimpleEnum.VALUE));
     }
 
+    @Override
+    public String errorParameterSerialization(AuthHeader authHeader, String headerParameter) {
+        if (headerParameter.startsWith("JSON") || headerParameter.equals("TOSTRING")) {
```
aldexis, line 202: Any reason not to always throw?

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/ParameterizedConjureGenerationTest.java
```diff
@@ -68,6 +69,7 @@ void testGeneratedCode(ParameterizedTestCase testCase) throws IOException {
 
     @Test
     @Order(2)
+    @Disabled
```
aldexis: Why is this disabled?

conjure-java-core/src/test/resources/ete-service.yml
```diff
@@ -217,6 +217,18 @@ services:
             type: string
         returns: string
 
+      errorParameterSerialization:
+        http: GET /errors/serialization
+        docs: |
+          This endpoint is used to test that that error parameters serialized as JSON or using `Objects.toString` are
```
aldexis:
```suggestion
          This endpoint is used to test that error parameters serialized as JSON or using `Objects.toString` are 
```

conjure-java-core/src/test/resources/ete-service.yml
```diff
@@ -217,6 +217,18 @@ services:
             type: string
         returns: string
 
+      errorParameterSerialization:
+        http: GET /errors/serialization
+        docs: |
+          This endpoint is used to test that that error parameters serialized as JSON or using `Objects.toString` are 
+          both handled correctly by the clients that return "rich" exceptions (which are sub-types of `RemoteException`).
```
aldexis:
```suggestion
          both handled correctly by the clients that can deserialize "rich" exceptions (which are sub-types of `RemoteException`).
```

conjure-java-core/src/main/java/com/palantir/conjure/java/util/ErrorGenerationUtils.java
```diff
@@ -145,15 +176,27 @@ public static void addAllParametersWithSafetyAnnotationsToMethodBuilder(
...
-    private static ParameterSpec.Builder buildParameterWithSafetyAnnotationInternal(
+    public static ParameterSpec buildUnsanitizedParameterWithSafetyAnnotationAndJsonProperty(
             TypeMapper typeMapper, FieldDefinition argDefinition, boolean isSafe) {
+        ParameterSpec.Builder parameterBuilder =
+                buildParameterWithSafetyAnnotationInternal(typeMapper, argDefinition, isSafe, false);
+        parameterBuilder.addAnnotation(AnnotationSpec.builder(JsonProperty.class)
+                .addMember("value", "$S", argDefinition.getFieldName().get())
+                .build());
+        return parameterBuilder.build();
+    }
```
aldexis, line 174: This should probably live next to `buildParameterWithSafetyAnnotationAndJsonProperty` and share more of the logic? Seems like the only difference is sanitized vs not (maybe we don't need _both_?)

conjure-java-core/src/integrationInput/java/undertow/com/palantir/product/ErrorServiceEndpoints.java
mpritham: Yes. I wanted to keep these to aid any static analysis. I don't think it hurts to have them here.

conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
mpritham: Yes that's right.

This first flag is arguably not required but I'm being cautious.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
mpritham: updated.

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowEteResource.java
mpritham: I wanted to also add a test to ensure that we're able to deserialize the return type correctly.

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/ParameterizedConjureGenerationTest.java
mpritham: typo. removed.

conjure-java-core/src/main/java/com/palantir/conjure/java/util/ErrorGenerationUtils.java
mpritham: [I removed `buildParameterWithSafetyAnnotationAndJsonProperty` when I reverted the addition of result type generation.](https://github.com/palantir/conjure-java/pull/2618/files#diff-00db4e48da0239df397b4da39d8f8bc571fce350e3b5080a3a5caa60aacc1456L200)

conjure-java-core/src/test/java/com/palantir/conjure/java/JerseyEteResource.java
```diff
@@ -179,6 +179,12 @@ public String jsonErrorsHeader(AuthHeader authHeader, String _headerParameter) {
         return "hello!";
     }
 
+    @Override
+    public String errorParameterSerialization(AuthHeader authHeader, String headerParameter) {
+        // Conjure-Java has not supported Jersey for a while. We are not testing the error serialization feature here.
+        return "hello!";
```
aldexis: nit: should we just explicitly throw to indicate that this is not meant to be tested? (that way, we'd catch more easily that this endpoint somehow ended up in a test somewhere)

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -183,6 +238,21 @@ private CodeBlock constructDeserializer(Optional<Type> type, TypeName className)
...
+        CodeBlock.Builder builder = CodeBlock.builder();
+        if (type.isEmpty()) {
+            builder.add("emptyBodyDeserializer(");
+        } else if (returnTypes.isBinary(className)) {
+            builder.add("inputStreamDeserializer(");
+        } else if (returnTypes.isOptionalBinary(className)) {
+            builder.add("optionalInputStreamDeserializer(");
+        } else {
+            builder.add("deserializer(");
+        }
```
aldexis: Is there a way we could make this more resilient to adding some other cases in the future? e.g. easier to discover we need to update this code location?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
```diff
@@ -153,21 +165,164 @@ private JavaFile generateErrorTypesForNamespace(
...
+            ClassName errorsClassName, List<ErrorDefinition> errorDefinitions) {
+        return errorDefinitions.stream()
+                .map(def -> generateRemoteExceptionType(errorsClassName, def))
+                .toList();
+    }
+
+    private static TypeSpec generateRemoteExceptionType(ClassName errorsClassName, ErrorDefinition errorDefinition) {
+        String remoteExceptionClassName = errorDefinition.getErrorName().getName() + "Exception";
+        ClassName serializableErrorClassName =
+                errorsClassName.nestedClass(errorDefinition.getErrorName().getName() + "SerializableError");
```
aldexis, line 200: Fwiw, I'm not fond of the `+ "Exception"`, `+ "SerializableError"` etc being encoded in multiple places. This feels like it ought to belong to a utility method in `ErrorGenerationUtils` like `errorParametersClassName`? (e.g. `errorExceptionName(ErrorDefinition errorDefinition)`?)

conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java
```diff
@@ -153,21 +165,164 @@ private JavaFile generateErrorTypesForNamespace(
...
+    private static List<TypeSpec> generateSerializableErrors(List<ErrorDefinition> errorDefinitions) {
+        return errorDefinitions.stream()
+                .map(ErrorGenerator::generateSerializableError)
+                .toList();
+    }
+
+    private static TypeSpec generateSerializableError(ErrorDefinition errorDefinition) {
+        String serializableErrorClassName = errorDefinition.getErrorName().getName() + "SerializableError";
+        String parameterClassNameString = ErrorGenerationUtils.errorParametersClassName(
+                errorDefinition.getErrorName().getName());
```
aldexis: vnit: could streamline these lines a bit by making `ErrorGenerationUtils.errorParametersClassName` take an `ErrorDefinition` (unless it's not always the same field that we're getting?)

conjure-java-core/src/main/java/com/palantir/conjure/java/util/ErrorGenerationUtils.java
```diff
@@ -62,6 +63,10 @@ public static String errorTypesClassName(ErrorNamespace namespace) {
         return namespace.get() + "Errors";
     }
 
+    public static String errorParametersClassName(String errorName) {
```
aldexis: Mentioned above, let's have this take `ErrorDefinition`

conjure-java-core/src/main/java/com/palantir/conjure/java/util/ErrorGenerationUtils.java
```diff
@@ -145,15 +150,27 @@ public static void addAllParametersWithSafetyAnnotationsToMethodBuilder(
 
     public static ParameterSpec buildParameterWithSafetyAnnotation(
             TypeMapper typeMapper, FieldDefinition argDefinition, boolean isSafe) {
-        return buildParameterWithSafetyAnnotationInternal(typeMapper, argDefinition, isSafe)
+        return buildParameterWithSafetyAnnotationInternal(typeMapper, argDefinition, isSafe, true)
                 .build();
     }
 
-    private static ParameterSpec.Builder buildParameterWithSafetyAnnotationInternal(
+    public static ParameterSpec buildUnsanitizedParameterWithSafetyAnnotationAndJsonProperty(
```
aldexis, line 166: nit: not necessarily asking for a change, but the long-name of this method, which seems to do a lot and is really hard to read, makes me think that we ought to rely more directly on the builder pattern here, and e.g. have _separate_ methods that do
- Build the parameter spec
- sanitize
- add safety annotation
- add json property

Then these can be combined in whichever way we need where need to.

For instance, maybe we could:
- Remove `WithSafetyAnnotation` from the method names (it seems to always be there?)
- Expose `buildParameterWithSafetyAnnotationInternal` as `buildParameter`
  - We could remove the existing `buildParameterWithSafetyAnnotation` which doesn't bring much by itself
- Add a `addJsonProperty` util method that takes the builder and arg definition

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -555,12 +557,24 @@ public void testListOfNull() {
...
             exceptionThrowingClient.jsonErrorsHeader(AuthHeader.valueOf("authHeader"), "JSON");
         } catch (RemoteException e) {
             assertThat(e.getError().parameters())
                     .containsExactlyInAnyOrderEntriesOf(
                             Map.of("serviceName", "my-service-string", "serviceDef", SimpleEnum.VALUE.toString()));
+            // Assert that error parameters can be re-serialized as JSON.
+            assertThat(e).isInstanceOfSatisfying(SerializableErrorProvider.class, errorProvider -> {
+                try {
+                    String serialized = objectMapper.writeValueAsString(
+                            errorProvider.error().parameterMap());
```
aldexis:
```suggestion
                    String serialized = objectMapper.writeValueAsString(errorProvider.error());
```

conjure-java-core/src/main/java/com/palantir/conjure/java/util/ErrorGenerationUtils.java
```diff
@@ -63,7 +63,16 @@ public static String errorTypesClassName(ErrorNamespace namespace) {
...
     }
 
-    public static String errorParametersClassName(String errorName) {
+    public static String serializableErrorClassName(String errorName) {
+        return errorName + "SerializableError";
+    }
+
+    public static String errorExceptionClassName(String errorName) {
+        return errorName + "Exception";
+    }
```
aldexis, line 72: nit: for consistency
```suggestion
    public static String serializableErrorClassName(ErrorDefinition errorDefinition) {
        return errorDefinition.getErrorName().getName() + "SerializableError";
    }

    public static String errorExceptionClassName(ErrorDefinition errorDefinition) {
        return errorDefinition.getErrorName().getName() + "Exception";
    }
```

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -240,14 +242,11 @@ private CodeBlock constructDeserializer(Optional<Type> type, TypeName className)
...
-            builder.add("inputStreamDeserializer(");
-        } else if (returnTypes.isOptionalBinary(className)) {
-            builder.add("optionalInputStreamDeserializer(");
-        } else {
-            builder.add("deserializer(");
+        switch (getDeserializerType(type, className)) {
+            case STANDARD -> builder.add("deserializer(");
+            case EMPTY_BODY -> builder.add("emptyBodyDeserializer(");
+            case BINARY -> builder.add("inputStreamDeserializer(");
+            case OPTIONAL_BINARY -> builder.add("optionalInputStreamDeserializer(");
```
aldexis, line 249: Nice, thanks!

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -539,4 +538,23 @@ private static MethodSpec toStringMethod(ClassName className) {
                         StaticFactoryMethodGenerator.ENDPOINT_CHANNEL_FACTORY)
                 .build();
     }
+
+    private enum DeserializerType {
+        STANDARD,
+        EMPTY_BODY,
+        BINARY,
+        OPTIONAL_BINARY;
+    }
```
aldexis, line 547: Do we not already have such an enum defined somewhere else? 🤔 Or was it a different repo?

aldexis (dismissed):
Approving, with just one main comment about the new test, which uses the now-defunct `parameterMap` method

aldexis (approved):

PR #2613, opened 2025-08-15
Add example error with several parameter types

mpritham:
## Before this PR
In #2608, I test the ser/de of various error parameter types. Several thousand lines of generated code is required for the Conjure objects I'm checking in. I'd like to checkin these definitions and generated files in a separate (this) PR to make reviewing #2608 easier.

## After this PR

==COMMIT_MSG==
Add example error with several parameter types
==COMMIT_MSG==


bjlaub (approved):
lgtm

Issue #2325, opened 2024-06-21
~10x perf degradation between 5.67.0 and 6.11.0 (and later)


michal-draftwise:
## What happened?

When upgrading from `5.67.0` to `8.6.0` we noticed about 10x longer task times for `:compileConjureObjects`. I checked different versions in between and it turns out the task was failing for versions `6.0.0` through to `6.10.0`, it started succeeding in `6.11.0` and it already suffered from described slowness.

I checked the most recent version (`8.23.0` at the tine of writing) but it is still slow.

## What did you want to happen?

Ideally we would have wanted for tasks to get quicker, but we'd be happy with them not being that significantly slower ;-)

CRogers:
Should be fixed by https://github.com/palantir/conjure-java/pull/2614

gmanis-draftwise:
Hey @CRogers -- thanks for the fix. However, it looks like it's still a pretty large perf degredation from 5.x; in our project running locally, the old version of conjure compiles in 15 seconds. Before the improvement, newer versions were ~2 minutes 30 seconds, and after, it is about 2 minutes 10 seconds.

kunalrkak:
@gmanis-draftwise apologies for the slowness as I didn't see this reply - sounds like there are still additional issues beyond what the PR fixed. Could you share if you're using the Conjure CLI directly or running through the gradle-conjure plugin + would it be possible to create a minimal reproduction that I could use to test/validate a potential fix with? For me, compiling objects with the latest conjure-java versions locally run in a few seconds as expected.

gmanis-draftwise:
Hi @kunalrkak -- thanks for getting back to me.

I tested this again and it took two and a half minutes but noticed

```[goethe] The current JVM does not have the right JVM arguments for direct formatting. Falling back to equivalent but slower bootstrapping formatter. Required exports: [jdk.compiler/com.sun.tools.javac.api, jdk.compiler/com.sun.tools.javac.file, jdk.compiler/com.sun.tools.javac.parser, jdk.compiler/com.sun.tools.javac.tree, jdk.compiler/com.sun.tools.javac.util]```

Adding 

```org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m \
--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED \
--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED \
[...]

PR #2614, opened 2025-08-15
Optimize CLI dist performance by adding JVM exports

kunalrkak:
## Before this PR

Addressing #2325. 

After transitioning from Java 8 to newer versions, the JPMS restricted access to certain compiler APIs that Conjure-Java relies on for efficient formatting. When running conjure-java through gradle environments, these restrictions were bypassed due to the `--add-exports` arguments specified in our [gradle.properties](https://github.com/palantir/conjure-java/blob/develop/gradle.properties#L4C1-L5C1), allowing the use of the efficient [DirectFormatterFacade](https://github.com/palantir/goethe/blob/develop/goethe/src/main/java/com/palantir/goethe/DirectFormatterFacade.java). However, when running the standalone CLI tool, it fell back to the slower [BootstrappingFormatterFacade](https://github.com/palantir/goethe/blob/develop/goethe/src/main/java/com/palantir/goethe/BootstrappingFormatterFacade.java), which launches a separate process for each formatting operation - we noted the slowness [here](https://github.com/palantir/goethe/blob/develop/goethe/src/main/java/com/palantir/goethe/FormatterFacadeFactory.java#L31-L33), but for some reason this message doesn't make its way to my stderr.

Testing locally on a sample input conjure IR, my execution time drops from ~53s to ~4s. 

## After this PR

==COMMIT_MSG==
Adds JVM exports for the `jdk.compiler` by default to conjure-java's CLI distribution. 
==COMMIT_MSG==

### DIFF
--- a/conjure-java/build.gradle
+++ b/conjure-java/build.gradle
@@ -18,6 +18,16 @@ apply plugin: 'com.palantir.external-publish-application-dist'
 
 mainClassName = 'com.palantir.conjure.java.cli.ConjureJavaCli'
 
+application {
+    applicationDefaultJvmArgs.addAll([
+            '--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED',
+            '--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED',
+            '--add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED',
+            '--add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED',
+            '--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED'
+    ])
+}
+
 dependencies {
     implementation project(':conjure-java-core')
     implementation 'com.fasterxml.jackson.core:jackson-annotations'


crogoz:
Thanks @kunalrkak, we will look into changing the [`com.palantir.baseline-module-jvm-args`](https://github.com/palantir/gradle-baseline/tree/develop?tab=readme-ov-file#baseline-java-code-quality-plugins) such that we will automatically propagate the necessary exports from the `palantir-java-format` to the application layer. This will fix other instances where projects use pjf as a dependency and they need the extra add-exports to be propagated.

conjure-java/build.gradle
```diff
@@ -18,6 +18,16 @@ apply plugin: 'com.palantir.external-publish-application-dist'
...
 
+application {
+    applicationDefaultJvmArgs.addAll([
+            '--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED',
+            '--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED',
+            '--add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED',
+            '--add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED',
+            '--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED'
+    ])
+}
```
CRogers, line 29: The optimal way to fix this would be get [`BaselineModuleJvmArgs`](https://github.com/palantir/gradle-baseline/blob/bb7a3628d1fc82a659c325ed1f67357961f0f80f/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineModuleJvmArgs.java#L65-L71) to support setting these values based on discovered required jvm args from the jars in the runtimeClasspath. I tried at [this attempt](https://github.com/palantir/gradle-baseline/compare/callumr/support-application-in-modulejvmargs) but I not sure it's possible with these setters/getters and no lazy properties or `CommandLineArgumentProvider`s. So best to merge this kind of fix.

conjure-java/build.gradle
```diff
@@ -18,6 +18,16 @@ apply plugin: 'com.palantir.external-publish-application-dist'
 
 mainClassName = 'com.palantir.conjure.java.cli.ConjureJavaCli'
 
+application {
+    applicationDefaultJvmArgs.addAll([
```
CRogers, line 22: The API for this is `Iterable<String> getApplicationDefaultJvmArgs();`. This code is relying on the fact the default value is an `ArrayList` which is a bit risky.

CRogers: But fine I guess as nothing else is adding to it here.

CRogers:
:+1:

PR #2622, opened 2025-08-25
Revert "Upgrade Eclipse Collections to v13, With Boxed Collection Support (#2548)"

bmarcaur:
This reverts commit 71aa627058cea7e08404ece632ac9e029ba05260.

Related issue: https://github.com/eclipse-collections/eclipse-collections/issues/1053

## Before this PR
Internal discussion: https://pl.ntr/eclipse-collections-issue-discussion

Turns out that the Eclipse Collections do not implement `AbstractList` and [do not implement all of the methods on collections](https://github.com/eclipse-collections/eclipse-collections/blob/72eaeaa1dbcf13fa5cee436273ba41ff4c927b4e/eclipse-collections-code-generator/src/main/resources/impl/list/mutable/primitiveArrayList.stg#L1034).

## After this PR
Reverts back to the old implementation but keeps the Eclipse Collections 13 upgrade. Also throws in some notes about not moving to the new implementation.

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
@@ -169,12 +169,6 @@ public final class ConjureCollections {
         return set;
     }
 
-    /**
-     * The following Conjure boxed list wrappers for the eclipse-collections [type]ArrayList are temporary (except
-     * ConjureSafeLongList). In eclipse-collections 12, a BoxedMutable[type]List will be released. Once available,
-     * Conjure[type]List should be replaced with that.
-     */
-
     // This method returns a list that can't handle nulls. Do not use this unless the nonNullCollections flag is set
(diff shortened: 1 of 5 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -169,6 +169,12 @@ public static <T> Set<T> newNonNullSet(Iterable<? extends T> iterable) {
         return set;
     }
 
+    /**
+     * The following Conjure boxed list wrappers for the eclipse-collections [type]ArrayList are temporary (except
+     * ConjureSafeLongList). In eclipse-collections 12, a BoxedMutable[type]List will be released. Once available,
+     * Conjure[type]List should be replaced with that.
+     */
+
```
bmarcaur: Gonna remove these docs from the revert, they kinda don't make sense in light of this change.

mpritham (approved):

PR #2623, opened 2025-08-26
Add tests for #2608

mpritham:
Separating #2608 to make it easier to review. This PR should be updated to create separate test files instead of using the existing `dialogue` test case, but using the existing test case makes it easy to see what concretely changes when the new functionality isused.


conjure-java-core/src/integrationInput/java/dialogue/com/palantir/another/ConjureErrors.java
```diff
@@ -39,4 +44,39 @@ public static boolean isDifferentPackageError(RemoteException remoteException) {
...
+            return SerializableError.builder()
+                    .errorCode(errorCode())
+                    .errorName(errorName())
+                    .errorInstanceId(errorInstanceId())
+                    .build();
+        }
+    }
+
+    public static final class DifferentPackageErrorException extends RemoteException {
+        private DifferentPackageErrorSerializableError error;
```
aldexis: `RemoteException` is serializable, due to being a subclass of `Throwable` (and contains a `serialVersionUid`). I'm wondering whether this means we need to make all these exceptions and the `AbstractSerializableError` also serializable (which would raise the question of how to maintain compatibility upon changes)

I don't think we would be relying on this anywhere, but this might be worth to explicitly acknowledge this?

conjure-java-core/src/integrationInput/java/dialogue/com/palantir/product/EmptyPathServiceAsync.java
```diff
@@ -34,11 +35,46 @@ static EmptyPathServiceAsync of(EndpointChannelFactory _endpointChannelFactory,
...
+                                new TypeMarker<ConjureErrors.InvalidServiceDefinitionSerializableError>() {},
+                                new TypeMarker<ConjureErrors.InvalidServiceDefinitionException>() {})
+                        .exception(
+                                ConjureErrors.INVALID_TYPE_DEFINITION.name(),
+                                new TypeMarker<ConjureErrors.InvalidTypeDefinitionSerializableError>() {},
+                                new TypeMarker<ConjureErrors.InvalidTypeDefinitionException>() {})
+                        .exception(
+                                ConjureJavaErrors.JAVA_COMPILATION_FAILED.name(),
+                                new TypeMarker<ConjureJavaErrors.JavaCompilationFailedSerializableError>() {},
+                                new TypeMarker<ConjureJavaErrors.JavaCompilationFailedException>() {})
```
aldexis, line 74: Should this not contain the error from the different package too? It _can_ be thrown by the service that defined it, right?

conjure-java-core/src/integrationInput/java/dialogue/com/palantir/product/EmptyPathServiceBlocking.java
```diff
@@ -33,11 +34,46 @@ static EmptyPathServiceBlocking of(EndpointChannelFactory _endpointChannelFactor
...
+                        .exception(
+                                ConjureErrors.INVALID_TYPE_DEFINITION.name(),
+                                new TypeMarker<ConjureErrors.InvalidTypeDefinitionSerializableError>() {},
+                                new TypeMarker<ConjureErrors.InvalidTypeDefinitionException>() {})
+                        .exception(
+                                ConjureJavaErrors.JAVA_COMPILATION_FAILED.name(),
+                                new TypeMarker<ConjureJavaErrors.JavaCompilationFailedSerializableError>() {},
+                                new TypeMarker<ConjureJavaErrors.JavaCompilationFailedException>() {})
+                        .build();
+            }
```
aldexis: It's bugging me a bit that we're defining the exact same method in each service, and then calling it to create new deserializer args for each endpoint (which will be the same if they share a return type).
Do you envision this being some kind of problem or maybe instead a benefit down the line? (e.g. whether on memory allocation, since we're copying the same map of possible hundreds of errors on each endpoint, which might be many as well, or our ability to change the code-gen later on)

Maybe it _is_ better this way, because if we need to later have different exceptions deserializable for different services, we don't need to break the ABI by removing a method that creates the generic deserializer. With that said, I'm not sure when we'd want to _not_ support all errors, but only a subset.

conjure-java-core/src/integrationInput/java/dialogue/com/palantir/product/EteBinaryServiceAsync.java
```diff
@@ -92,33 +148,38 @@ public ListenableFuture<InputStream> postBinaryThrows(
...
+                    _request.putHeaderParams(
+                            "Accept-Conjure-Error-Parameter-Format",
+                            _runtime.bodySerDe().errorParameterFormat().get().toString());
+                }
                 return _runtime.clients()
-                        .call(
-                                getOptionalBinaryPresentChannel,
-                                _request.build(),
-                                _runtime.bodySerDe().optionalInputStreamDeserializer());
+                        .call(getOptionalBinaryPresentChannel, _request.build(), getOptionalBinaryPresentDeserializer);
```
aldexis, line 177: side-note: the name of this endpoint is so confusing, because I though `getOptionalBinaryPresentChannel` and `getOptionalBinaryPresentDeserializer` were methods (e.g. suppliers), when they were in fact fields

conjure-java-core/src/integrationInput/java/dialogue/test/api/CookieServiceAsync.java
```diff
@@ -33,13 +35,24 @@ static CookieServiceAsync of(EndpointChannelFactory _endpointChannelFactory, Con
...
 
-            private final Deserializer<Void> eatCookiesDeserializer =
-                    _runtime.bodySerDe().emptyBodyDeserializer();
+            private final Deserializer<Void> eatCookiesDeserializer = _runtime.bodySerDe()
+                    .emptyBodyDeserializer(createExceptionDeserializerArgs(new TypeMarker<Void>() {}));
+
+            private <T> ExceptionDeserializerArgs<T> createExceptionDeserializerArgs(TypeMarker<T> returnType) {
+                return ExceptionDeserializerArgs.<T>builder()
+                        .returnType(returnType)
+                        .build();
```
aldexis: Is `CookieServiceAsync` considered as part of a different conjure project? If not, should it also support deserializing all errors?

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
```diff
@@ -587,12 +587,11 @@ public void testStringParametersDoNotChangeWhenUsingJsonAndJavaString() {
...
             // .getError() returns the SerializableError which should contain the legacy parameters sent over the wire.
             jsonParams = e.getError().parameters();
-            // e should be an instance of ErrorWithComplexArgsException, which has rich parameters as well.
-            //            assertThat(e.getCause()).isInstanceOfSatisfying(ErrorWithComplexArgsException.class, exception
-            // -> {
-            //                assertThat(exception.error().parameters().optionalExample().getOptionalString())
-            //                        .contains("optional-value");
-            //            });
+            // e should be an instance of ErrorWithComplexArgsException, which has rich parameters as well
+            assertThat(e.getCause()).isInstanceOfSatisfying(ErrorWithComplexArgsException.class, exception -> {
```
aldexis: Shouldn't this be
```suggestion
            assertThat(e).isInstanceOfSatisfying(ErrorWithComplexArgsException.class, exception -> {
```
rather than `e.getCause()`?

conjure-java-core/src/integrationInput/java/dialogue/com/palantir/product/EmptyPathServiceBlocking.java
mpritham: One solution to this could be to create a new file that just has this method which all generated services can use. From the comment above regarding adding all errors in the Conjure definitions, we'd have only one helper method definition in this file.

So concretely, I'm thinking of a file `ServiceErrorUtils` that is generated which has one method `createExceptionDeserializerArgs`.

If we end up changing how the deserializers are created, I think we shouldn't have an ABI break?

> With that said, I'm not sure when we'd want to not support all errors, but only a subset.

I think this is related to the comment above: whether we filter errors on defined in the same package, or just include _every_ error in the Conjure definition provided. I implemented the former, but I'm sort of learning toward the latter to make migration easier. Some folks end up creating errors in a separate `errors` package.

conjure-java-core/src/integrationInput/java/dialogue/test/api/CookieServiceAsync.java
mpritham: This is related to the comments above regarding the filtering around what errors are considered. In the implementation, I added filters so services would be able to handle all errors defined in the same package. After looking at some usages, it's probably safest to include all errors defined - some folks put errors in a separate package `com.palantir.x.errors`.

Even if they didn't, it's possible, like you said, for a service to throw errors defined in a different package.

conjure-java-core/src/integrationInput/java/dialogue/com/palantir/product/EmptyPathServiceBlocking.java
mpritham: So the changes I want to make here:
- take out the filtering on package - consider _every_ error defined for every service
- add a separate utility file for this helper method

With the separate utility file, we will still end up creating loads of maps. One idea is to introduce a cache from `returnType` -> `exceptionDeserializerArgs`.

conjure-java-core/src/integrationInput/java/dialogue/com/palantir/another/ConjureErrors.java
mpritham: yeah, I'm not sure of the history here, but we use jackson to deserialize some subclass of `AbstractSerializableError`, and construct this class - it's never deserialized to directly off the wire. I'm not sure that `RemoteException` is either.

I can acknowledge this in a comment.

aldexis: All `Throwable`s are `Serializable`, so it's really just a matter of whether we care. It sounds like `RemoteException` actually has a `serialVersionUid`. It sounds like it was added form the get-go though https://github.com/palantir/conjure-java-runtime-api/pull/9 and I can't imagine this being used as serializable anywhere.

With that said, it might be worth just adding a default `serialVersionUid` on the subclasses nonetheless? 🤔 (adding it after the fact is a bit annoying after all)

conjure-java-core/src/integrationInput/java/dialogue/com/palantir/product/EmptyPathServiceBlocking.java
aldexis: We chatted about this yesterday. There unfortunately isn't really a great place to put such a file, since it can't be in a predefined package, to avoid conflicts with other dialogue client libraries. I think it's fine to leave it as one method per client for now, and get rid of it if we realize this causes problems. This is also why we have various safe rollout mechanisms after all.

This (and the fact that we can't deserialize exceptions that aren't known at codegen time) makes me wonder however whether we could instead try to discover exceptions at runtime (e.g. through some autoservice or other mechanism, making it possible to build a centralized map of all exceptions, instead of having each client and deserializer to know about them).

conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java
aldexis: Discussed about this yesterday. This is because of the pattern from https://github.com/palantir/dialogue/blob/c17a28c5673cd3988056a4181d38f5f1eae43f59/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java#L148.

We ought to either replicate this pattern with the new remote exceptions, or instead get rid of it entirely (and do [what is done for other types of exceptions](https://github.com/palantir/dialogue/blob/c17a28c5673cd3988056a4181d38f5f1eae43f59/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java#L134))

aldexis (commented):
Didn't do a very deep review of the code changes (i.e. I didn't inspect each line, because they seem mostly duplicates of each other), so I could have missed one specific change that is problematic among all the others, but this seems fine to me 👍

mpritham:
I'm going to modify #2608 to create a new test case (instead of setting the flag to an existing test case like I did in this PR). No need to merge this PR into #2608.

PR #2649, opened 2025-09-25
[Feature] Conjure unions are generated as sealed classes

kunalrkak:
## Before this PR
Using sealed classes for conjure unions allows consumers to use switch statements instead of verbose visitors.
Heavily influenced by the work and existing discussion in #1838.

## After this PR

==COMMIT_MSG==
Allow conjure unions to be generated as sealed classes.
==COMMIT_MSG==


conjure-java-core/src/integrationInput/java/nongenerated/com/palantir/product/api/ExampleSealedInterfaceUnion.java
```diff
@@ -0,0 +1,125 @@
...
+import nongenerated.com.palantir.product.api.ExampleSealedInterfaceUnion.UnknownVariant;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type", defaultImpl = UnknownVariant.class)
+@JsonSubTypes({
+    @JsonSubTypes.Type(value = Foo.class, name = "foo"),
+    @JsonSubTypes.Type(value = Bar.class, name = "bar"),
+    @JsonSubTypes.Type(value = Baz.class, name = "baz")
+})
+@JsonIgnoreProperties(ignoreUnknown = true)
+public sealed interface ExampleSealedInterfaceUnion {
```
kunalrkak: fysa: this class was handrolled to roughly represent the ideal end state for initial testing

conjure-java-core/src/integrationInput/java/nongenerated/com/palantir/product/api/ExampleSealedInterfaceUnion.java
```diff
@@ -0,0 +1,125 @@
...
+            Preconditions.checkNotNull(value, "baz cannot be null");
+        }
+
+        @Override
+        public String toString() {
+            return "ExampleSealedInterfaceUnion.Baz{value: " + value + '}';
+        }
+    }
+
+    record UnknownVariant(@JsonProperty("type") String type, @JsonProperty("value") Map<String, Object> value)
```
mpritham: What would the value be? E.g. if a server added a new variant to the union "newType: safelong". I guess the type would be "safelong". 

If the motivation to provide more details to the client on what variant was unhandled, maybe just the JSON key is sufficient. It's also possible that the actual type is not safe to log, and we don't have safety information to correctly handle the values?

conjure-java-core/src/integrationInput/java/nongenerated/com/palantir/product/api/ExampleSealedInterfaceUnion.java
```diff
@@ -0,0 +1,125 @@
...
+import nongenerated.com.palantir.product.api.ExampleSealedInterfaceUnion.Foo;
+import nongenerated.com.palantir.product.api.ExampleSealedInterfaceUnion.UnknownVariant;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type", defaultImpl = UnknownVariant.class)
+@JsonSubTypes({
+    @JsonSubTypes.Type(value = Foo.class, name = "foo"),
+    @JsonSubTypes.Type(value = Bar.class, name = "bar"),
+    @JsonSubTypes.Type(value = Baz.class, name = "baz")
+})
+@JsonIgnoreProperties(ignoreUnknown = true)
```
mpritham: hm, if we enable this, are we still able to construct an "unknown" variant when there's some key in the JSON that's not recognized?

conjure-java-core/src/integrationInput/java/nongenerated/com/palantir/product/api/ExampleSealedInterfaceUnion.java
```diff
@@ -0,0 +1,125 @@
...
+        @Override
+        public String toString() {
+            return "ExampleSealedInterfaceUnion.Baz{value: " + value + '}';
+        }
+    }
+
+    record UnknownVariant(@JsonProperty("type") String type, @JsonProperty("value") Map<String, Object> value)
+            implements ExampleSealedInterfaceUnion {
+
+        public UnknownVariant {
```
mpritham: Just for clarity what is the desired client usage? Are they matching on `ExampleSealedInterfaceUnion` or `Known`?

hm, should we be matching `UnknownVariant`? Something like this maybe confusing:

```java
switch (exampleSealedInterfaceUnion) {
   Foo val -> {...}
   Bar val -> {...}
   Baz val -> {...}
   UnknownVariant val -> {...}
   default -> {...}
}
```

If we don't permit the `default` case (i.e. via an errorprone check), then introducing a new variant would lead to a compile break. (maybe this is desired but that's arguable)

conjure-java-core/src/integrationInput/java/nongenerated/com/palantir/product/api/ExampleSealedInterfaceUnion.java
mpritham: > It's also possible that the actual type is not safe to log, and we don't have safety information to correctly handle the values?

I see we do this for the unknown unions currently generated.

conjure-java-core/src/integrationInput/java/nongenerated/com/palantir/product/api/ExampleSealedInterfaceUnion.java
kunalrkak: I wrote up two examples of client consumption [here](https://github.com/palantir/conjure-java/pull/2649/commits/2734bda9244de33ba633154d6e0c19541675bdc2) - I think they could actually match on either `ExampleSealedInterfaceUnion` or `Known`. Also tested the example you created - if the switch statement is exhaustive, the IDE gives a hint that `'default' branch is unnecessary `. I wonder if that can be codified into errorprone?

conjure-java-core/src/integrationInput/java/nongenerated/com/palantir/product/api/ExampleSealedInterfaceUnion.java
kunalrkak: I think this is a misnomer - "type" here refers to what union variant it is, not the java type. But yes I think the value is generic just to be able to deserialize anything passed to it

conjure-java-core/src/integrationInput/java/nongenerated/com/palantir/product/api/ExampleSealedInterfaceUnion.java
kunalrkak: Will write a test for this!

kunalrkak: Had an issue with JsonTypeInfo not passing the "type" back to the UnknownVariant constructor, but fixed that and wrote some tests [here](https://github.com/palantir/conjure-java/pull/2649/commits/3bd4c1c7ab764a1285e6e21af849e610f578344e#diff-9173b1b67f907da9802c544ef1a52a03426d8a02685830f63ce07c2152e8073aR235-R262)

conjure-java-core/src/test/java/com/palantir/conjure/java/types/UnionTests.java
```diff
@@ -207,6 +213,53 @@ void unionWithDefensiveCollectionsIsImmutable_map() {
         assertThatExceptionOfType(UnsupportedOperationException.class).isThrownBy(() -> internal.put("test", "update"));
     }
 
+    @Test
+    void sealedUnionCanBeConsumedWithSwitchStatement() {
+        String expected = "foo";
+
+        String s =
+                switch (SimpleUnion.foo(expected).throwOnUnknown()) {
```
mpritham: hm why do we need the `throwOnUnknown` in `SimpleUnion`? We're constructing the object so I think this is unneeded (unless you were planning on creating some sort of visitor for back compat?)

conjure-java-core/src/integrationInput/java/nongenerated/com/palantir/product/api/ExampleSealedInterfaceUnion.java
mpritham: I'm thinking about how to roll this out without the risk of ABI breaks. Concretely the error case I have in mind is:
- server 1 publishes objects with the new sealed interface unions
- library has some code using the old unions
- server 2 upgrades the objects JAR for server 1 and makes a call to the library, leading to a runtime failure


One idea is to update the current implementation of unions so that the generated `Base` type is 1) public and attainable via `getValue`, 2) is a sealed interface permitted all of the variants. I think this would work: clients can continue to use the visitor, and clients that want to use the sealed unions will just have to match on the `Base` type with a call to `getValue()`. What are your thoughts here?

kunalrkak:
Notes from chatting with @mpritham offline:

- Add explicit permits clause for union base class i.e. `permits SimpleUnion.Known, SimpleUnion.UnknownVariant`
- Prefer exhaustive switch statements via errorprone, to force consumers to update "visitors" when new union variants are added
- Can we test if there's any issue with making the union base an interface instead of a class? Potentially reflection logic that relies on the latter?

mpritham:
> Add explicit permits clause for union base class i.e. permits SimpleUnion.Known, SimpleUnion.UnknownVariant

Yeah. It's implicitly added because of [JLS 9.1.4](https://docs.oracle.com/javase/specs/jls/se21/html/jls-9.html#jls-9.1.4). 

But also I wonder if we need the `Known` type - the rationale for not having it is that we'll be able to force consumers to consider all the union variants. By supporting the `Known` type, this will be considered an exhaustive pattern matching switch:

```java
switch (simpleUnion) {
  Known k -> {...}
  Unknown _ -> {...}
}
```
and there will be no compile breaks if the API were to update the `Known` interface with another variant.

mpritham:
I see this piece of code in [Dialogue that uses reflection to deserialize `null` into a type.](https://github.com/palantir/dialogue/blob/d9293d4b203b3200c72964d3fb7b3e93f4dcce0e/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java#L141) It looks explicitly for a `Class`. 

Could we add an ETE test that tries to have Dialogue deserialize an endpoint that has a union result type and return `null`? We could probably add this in `UndertowServiceEteTest` [(src)](https://github.com/palantir/conjure-java/blob/b7812ec52cde1ef1fedea6a5807107a21855196d/conjure-java-core/src/test/java/com/palantir/conjure/java/UndertowServiceEteTest.java)

conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
```diff
@@ -188,36 +188,44 @@ default boolean preferObjectBuilders() {
...
+    default boolean sealedUnions() {
+        return false;
+    }
+
+    /**
+     * If {@link #sealedUnions} is enabled, this enables visitor generation for back-compatibility.
+     */
+    @Value.Default
+    default boolean sealedUnionVisitors() {
         return false;
```
mpritham: Thoughts on having this be `true` by default? This way devs don't have to set two flags to enable this feature. When the visitors are removed, it will be really clear because there will be a flag in the Gradle file setting this flag to false.

conjure-java-core/src/test/java/com/palantir/conjure/java/types/UnionTests.java
```diff
@@ -207,6 +214,79 @@ void unionWithDefensiveCollectionsIsImmutable_map() {
...
+    }
+
+    @Test
+    public void testCreateUnknownType_sealedUnion() {
+        String expectedUnknownType = "qux";
+        List<String> expectedUnknownValue = List.of("quux", "quuz");
+        SimpleUnion union = SimpleUnion.unknown(expectedUnknownType, expectedUnknownValue);
+
+        switch (union) {
+            // UnknownVariant is a bit verbose - should we just call this "Unknown"?
```
mpritham: 👍 Yeah, I think just having it be "unknown" is cleaner.

conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java
```diff
@@ -240,21 +247,10 @@ public static final class GenerateCommand implements Runnable {
...
-                names = "--generateErrorParameterFormatRespectingDialogueInterfaces",
-                defaultValue = "false",
-                description =
-                        "Enabling this flag will lead to the generation of Dialogue client interfaces that are able to"
-                            + " send a header to servers specifying the Conjure error parameter serialization format"
-                            + " they expect, and create custom exceptions for each error defined in the service's"
-                            + " package.")
-        private boolean generateErrorParameterFormatRespectingDialogueInterfaces;
+                description = "Uses sealed classes for union type generation.")
+        private boolean sealedUnions;
```
mpritham, line 265: We also need to add the sealed union visitors flag here.

build.gradle
```diff
@@ -47,7 +48,7 @@ apply plugin: 'com.palantir.baseline-java-versions'
 apply plugin: 'com.palantir.jdks.latest'
 
 javaVersions {
-    libraryTarget = 17
+    libraryTarget = 21
```
mpritham: reminder to remove.

readme.md
```diff
@@ -51,9 +51,17 @@ The recommended way to use conjure-java is via a build tool like [gradle-conjure
...
+                     This is an experimental feature that may cause compilation failures! If enabled, endpoints that 
+                     have associated errors will return a result type: a sealed interface permitting subclasses for the 
+                     endpoint's return value, and each endpoint error. Producing JARs with this feature enabled will 
+                     result in a compile-time break when consumers bump their dependency on the JAR. This is because the
+                     return types of every endpoint with associated errors will change.
         --defensiveCollections
                       Makes immutable copies of collections for union and alias types, respecting the 
                       --nonNullCollections flag.
+        --sealedUnions
+                      Uses sealed interfaces for union type generation.
```
mpritham: Reminder to add sealed union visitors flag.

conjure-java-core/src/test/java/com/palantir/conjure/java/parameterized/TestCases.java
```diff
@@ -342,6 +322,37 @@ public final class TestCases {
...
+                    .generatorTypes(List.of(
+                            GeneratorType.OBJECT,
+                            GeneratorType.ERROR,
+                            GeneratorType.CHECKED_ERROR,
+                            GeneratorType.DIALOGUE,
+                            GeneratorType.UNDERTOW))
+                    .build())
+            .add(ParameterizedTestCase.builder()
+                    .name("sealed-unions")
+                    .docs("Testing generating union types as sealed interfaces")
```
mpritham: nit:
```suggestion
                    .docs("Testing generating union types as sealed classes")
```

conjure-java-core/src/test/resources/example-unions.yml
```diff
@@ -0,0 +1,25 @@
+types:
+  definitions:
+    default-package: com.palantir.product
+    objects:
+      SimpleUnion:
+        union:
+          foo: string
```
mpritham, line 7: SInce we're no longer suffixing the class with `Wrapper`, what happens if we have a field with the same name as the union name (`simpleUnion` in this case)? We probably need to sanitize that.

conjure-java-core/src/test/java/com/palantir/conjure/java/types/UnionTests.java
```diff
@@ -207,6 +214,79 @@ void unionWithDefensiveCollectionsIsImmutable_map() {
         assertThatExceptionOfType(UnsupportedOperationException.class).isThrownBy(() -> internal.put("test", "update"));
     }
 
+    @Test
+    void sealedUnionCanBeConsumedWithSwitchStatement() {
+        String expected = "foo";
+
+        String actual =
+                switch (SimpleUnion.foo(expected).throwOnUnknown()) {
+                    case Foo foo -> foo.value();
```
mpritham: Thoughts on making these exhaustive?

```java
switch (SimpleUnion.foo(expected).throwOnUnknown()) {
    case Foo foo -> foo.value();
    case Bar _bar -> throw new SafeIllegalStateException();
    case Baz _baz -> throw new SafeIllegalStateException();
};
```

readme.md
```diff
@@ -51,9 +51,17 @@ The recommended way to use conjure-java is via a build tool like [gradle-conjure
...
+        --generateDialogueEndpointErrorResultTypes
+                     This is an experimental feature that may cause compilation failures! If enabled, endpoints that 
+                     have associated errors will return a result type: a sealed interface permitting subclasses for the 
+                     endpoint's return value, and each endpoint error. Producing JARs with this feature enabled will 
+                     result in a compile-time break when consumers bump their dependency on the JAR. This is because the
+                     return types of every endpoint with associated errors will change.
         --defensiveCollections
                       Makes immutable copies of collections for union and alias types, respecting the 
                       --nonNullCollections flag.
+        --sealedUnions
```
mpritham, line 57: Also, did we want to add a comment about the risks or considerations folks will have to take when reverting this flag?

conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
kunalrkak: yup good call - I think we actually want this default on when folks are switching over, we can set it to false by default in new repos

build.gradle
kunalrkak: hmm I thought I needed this, I guess none of the generated code actually requires 21 though, but some of the tests use the pattern matching from 21

conjure-java-core/src/test/resources/example-unions.yml
kunalrkak: you're totally correct, I'll track this!

readme.md
kunalrkak: I have that in the [Options javadoc](https://github.com/palantir/conjure-java/blob/b676104d5c22b96acafc29b99c86165f6097c02c/conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java#L216-L217), do you think its useful to add to the readme?

mpritham (dismissed):
Approving for an RC

mpritham (dismissed):
approving for rc

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/EmptyUnion.java
```diff
@@ -0,0 +1,167 @@
...
+        switch (Preconditions.checkNotNull(type, "Type is required")) {
+            default:
+                return new Unknown(type, Collections.singletonMap(type, value));
+        }
+    }
+
+    public Known throwOnUnknown() {
+        throw new SafeIllegalArgumentException(
+                "Unknown variant of the 'EmptyUnion' union", SafeArg.of("unknownType", ((Unknown) this).type()));
+    }
```
kunalrkak: @mpritham & @aldexis - curious for your thoughts on how this should be handled for empty union types. I refactored the code so it unconditionally throws if there are no known variants, but that's mainly to keep the method present. we could also just special case and not generate this for empty unions, since its not really useful.

aldexis: This seems like a reasonable approach, since it provides information about the "unexpected" type.

I would however get rid of `public interface Known {}` because you can't make it sealed afterwards without breaking ABI (though I wouldn't expect people to implement it, but we should simply not have it)

aldexis: Alternatively, I wonder whether it should be valid in the first place to have an empty union...

kunalrkak: It's valid today, so regardless we have to support empty union generation

Hmm, good point about the Known interface - if we remove this, what would `throwOnUnknown` return? Could we just make Known private in this case?

aldexis: Oh good point, I had missed that it was the return type 🤔 
How would you even use the `throwOnUnknown` method if its return type is private though? At this point, we might be better off just removing the method entirely? What is the expected usage in this instance?

kunalrkak: Yeah removing the method as well as the Known interface seems least confusing. Keeping it really only makes sense for API consistency, but I don't think the case of 1+ variants -> empty union is ABI safe anyways.

mpritham (dismissed):
LGTM. Thanks!

mpritham (approved):
Thanks!

PR #2650, opened 2025-09-26
Support optional auth parameters in conjure-undertow-annotations

pkoenig10:
There are some legitimate cases where endpoints do not require authentication, but will use credentials if they are provided. For example, a logout endpoint.

Currently, the `Authorization` header parameters of these endpoints must be declared using `@Header` and are interpreted as normal header parameters instead of an authorization parameter. This means that we won't use `AuthorizationExtractor` to deserialize them and thus don't populate the MDC with the metadata from the provided credential.

This PR updates conjure-undertow-annotations to support `Optional<AuthHeader>` and `@Cookie Optional<BearerToken>` parameters and treat them as auth parameters.

### DIFF
--- a/conjure-undertow-annotations/readme.md
+++ b/conjure-undertow-annotations/readme.md
@@ -77,8 +77,7 @@ public interface MyService {
 }
 ```
 
-When using an HTTP authentication header in the form of `Bearer [token]`, the token can be automatically injected when
-using an [`AuthHeader`](https://github.com/palantir/auth-tokens/blob/develop/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java) parameter:
+To access a bearer token provided using the HTTP `Authorization` header in the form of `Bearer [token]`, you can use an unannotated [`AuthHeader`](https://github.com/palantir/auth-tokens/blob/develop/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java) or `Optional<AuthHeader>` parameter, which will be deserialized using the [`AuthorizationExtractor`](../conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/AuthorizationExtractor.java):
 
 ```java
 public interface MyService {
@@ -88,14 +87,30 @@ public interface MyService {
 }
 ```
 
-Similarly, you can access a bearer token from a cookie when using the `@Handle.Cookie` annotation together with the
-[`BearerToken`](https://github.com/palantir/auth-tokens/blob/develop/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java) type which also sets the respective auth state:
+```java
+public interface MyService {
+
+    @Handle(method = HttpMethod.GET, path = "/path")
+    void myEndpoint(Optional<AuthHeader> authHeader);
+}
+```
+
+Similarly, to access a bearer token provided using a cookie, you can use the `@Handle.Cookie` annotation on a 
+[`BearerToken`](https://github.com/palantir/auth-tokens/blob/develop/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java) or `Optional<BearerToken>` parameter, which will be deserialized using the [`AuthorizationExtractor`](../conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/AuthorizationExtractor.java):
 
 ```java
 public interface MyService {
 
     @Handle(method = HttpMethod.GET, path = "/path")
     void myEndpoint(@Handle.Cookie("AUTH_TOKEN") BearerToken token);
+}
+```
+
+```java
+public interface MyService {
+
+    @Handle(method = HttpMethod.GET, path = "/path")
+    void myEndpoint(@Handle.Cookie("AUTH_TOKEN") Optional<BearerToken> token);
 }
 ```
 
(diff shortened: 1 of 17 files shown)


conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/EndpointDefinitions.java
```diff
@@ -132,6 +133,13 @@ public Optional<EndpointDefinition> tryParseEndpointDefinition(
             return Optional.empty();
         }
 
+        long numAuthParameters = argumentDefinitions.stream()
+                .filter(argument -> argument.paramType().match(IsAuthVisitor.INSTANCE))
+                .count();
+        if (numAuthParameters > 1) {
+            errorContext.reportError("Methods cannot have multiple auth parameters", element);
+        }
```
pkoenig10, line 142: Resolves this TODO

https://github.com/palantir/conjure-java/blob/3abe19157b0bf06b3f3100057f4f7c4d1f9e588c/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java#L290

If this ends up blocking some upgrades, that's probably a good thing, since it means the MDC set for a particular requests is somewhat unclear.

Although I think it's pretty unlikely that there is existing code that would violate this check. Partly because our internal auditing annotation processor library also does this validation.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
```diff
@@ -388,24 +389,18 @@ public Void cookie(
...
-                                .build())
-                        .constructorInitializer(CodeBlock.builder()
-                                .addStatement(
-                                        "this.$N = new $T(runtime, $S)",
-                                        deserializerFieldName,
-                                        BearerTokenCookieDeserializer.class,
-                                        cookieName)
-                                .build())
-                        .build());
-                handlerBuilder.addStatement("$T $N = $L", paramType, variableName, invokeDeserializer(def));
```
pkoenig10, line 408: The indirection through `BearerTokenCookieDeserializer` seems unnecessary and is inconsistent with the code generated for `AuthHeader`, where we invoke `runtime.auth()` directly.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
```diff
@@ -629,15 +634,37 @@ public CodeBlock cookie(
...
             public CodeBlock authHeader(String _variableName) {
                 return CodeBlock.of("this.$N.auth().header($N)", RUNTIME_NAME, EXCHANGE_NAME);
             }
 
+            @Override
+            public CodeBlock optionalAuthHeader(String _variableName) {
+                return CodeBlock.of(
+                        "$2N.getRequestHeaders().contains($3T.AUTHORIZATION) "
+                                + "? Optional.of(this.$1N.auth().header($2N)) "
+                                + ": Optional.empty()",
```
pkoenig10, line 662: This is a little gross, but otherwise we have to add new methods to `AuthorizationExtractor`. Doing so is difficult because this is a public interface that is used outside this repo, so this has some non-trivial ABI concerns.

We can always improve this in the future if we want do.

bjlaub:
Can we update the docs here describing the use case and showing an example? https://github.com/palantir/conjure-java/tree/develop/conjure-undertow-annotations#header-and-cookie-values

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/EndpointDefinitions.java
pkoenig10: It turns out that our internal auditing annotation processor library already supports optional `AuthHeader` and cookie `BearerToken` parameters.

conjure-undertow-annotations/readme.md
```diff
@@ -77,8 +77,7 @@ public interface MyService {
 }
 ```
 
-When using an HTTP authentication header in the form of `Bearer [token]`, the token can be automatically injected when
-using an [`AuthHeader`](https://github.com/palantir/auth-tokens/blob/develop/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java) parameter:
+To access a bearer token provided using the HTTP `Authorization` header in the form of `Bearer [token]`, you can use an unannotated [`AuthHeader`](https://github.com/palantir/auth-tokens/blob/develop/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java) or `Optional<AuthHeader>` parameter, which will be deserialized using the [`AuthorizationExtractor`](../conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/AuthorizationExtractor.java):
```
pkoenig10, line 80: Updated these docs to make it clear that these parameters, unlike other `@Header` or `@Cookie` parameters, are deserialized using `AuthorizationExtractor`, which is the thing that sets the MDC state.

bjlaub: Cool, thanks. I wonder if we should clarify that the runtime will return a 401 in the case of an endpoint with a non-optional `AuthHeader` receiving a missing/bogus token, but will not automatically do that in the case of an endpoint that declares `Optional<AuthHeader>`? It's kind of an implementation detail so maybe not worth it. 🤷

bjlaub (approved):
lgtm, thanks

conjure-undertow-annotations/readme.md
pkoenig10: This is an implementation detail of the particular `AuthorizationExtractor` implementation used by the default `UndertowRuntime`. It's something that is universally true for any use of `conjure-undertow-annotations` (although any reasonable `AuthorizationExtractor` implementation would return a 401 in this case).

PR #2660, opened 2025-10-07
Support deserialization of double aliases from strings

mpritham:
## Before this PR
Aliases of doubles are valid keys in maps. Keys in maps are serialized as strings. Aliases of doubles can only be deserialized from `NaN`, `Infinity`, and `-Infinity`. We should support deserialization of other valid double values.

## After this PR

==COMMIT_MSG==
Support deserialization of double aliases from strings
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/allexamples/com/palantir/product/DoubleAliasExample.java
+++ b/conjure-java-core/src/integrationInput/java/allexamples/com/palantir/product/DoubleAliasExample.java
@@ -77,7 +77,16 @@ public final class DoubleAliasExample implements Comparable<DoubleAliasExample> 
             case "-Infinity":
                 return DoubleAliasExample.of(Double.NEGATIVE_INFINITY);
             default:
-                throw new IllegalArgumentException("Cannot deserialize string into double: " + value);
+                try {
+                    double doubleValue = Double.parseDouble(value);
+                    if (Double.toString(doubleValue).equals(value)) {
+                        return DoubleAliasExample.of(doubleValue);
+                    } else {
+                        throw new IllegalArgumentException("Cannot deserialize string into double: " + value);
+                    }
+                } catch (NumberFormatException e) {
+                    throw new IllegalArgumentException("Cannot deserialize string into double: " + value);
+                }
         }
     }
 }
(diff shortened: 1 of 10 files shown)


conjure-java-core/src/integrationInput/java/allexamples/com/palantir/product/DoubleAliasExample.java
```diff
@@ -77,7 +77,16 @@ public static DoubleAliasExample of(String value) {
             case "-Infinity":
                 return DoubleAliasExample.of(Double.NEGATIVE_INFINITY);
             default:
-                throw new IllegalArgumentException("Cannot deserialize string into double: " + value);
+                try {
+                    double doubleValue = Double.parseDouble(value);
```
mpritham, line 81: Why not just call `Double.parseDouble`? Because we don't want to accept things like hexadecimals and 1e10 (but we can accept 1.0E10 because that's a properly formatted double).

aldexis: Maybe add this as a comment on the codegen code? I could certainly see this be asked in a couple of years and someone changing this because they fail to find your PR comment

conjure-java-core/src/integrationInput/java/allexamples/com/palantir/product/DoubleAliasExample.java
```diff
@@ -77,7 +77,16 @@ public static DoubleAliasExample of(String value) {
             case "-Infinity":
                 return DoubleAliasExample.of(Double.NEGATIVE_INFINITY);
             default:
-                throw new IllegalArgumentException("Cannot deserialize string into double: " + value);
```
aldexis, line 80: Why is this specific to map keys? How are we deserializing non-map-key double aliases if not through this code? (I'm asking specifically because I'm curious about the perf cost of this change - it might not matter if it's for double alias map keys, but maybe more so if that's for all double alias deserializations?)

conjure-java-core/src/test/resources/example-types.yml
```diff
@@ -261,7 +261,8 @@ types:
           rids: map<RidAliasExample, ManyFieldExample>
           bearertokens: map<BearerTokenAliasExample, ManyFieldExample>
           integers: map<IntegerAliasExample, ManyFieldExample>
-          # doubles: map<DoubleAliasExample, ManyFieldExample> # typescript freaks out with the 'NaN'
+          doubles: map<DoubleAliasExample, ManyFieldExample> # typescript freaks out with the 'NaN'
```
aldexis, line 264: nit: I would probably just remove the typescript comment too, as I'm unclear what value it provides in the conjure-java repo
```suggestion
          doubles: map<DoubleAliasExample, ManyFieldExample>
```

PR #2696, opened 2025-10-27
Generate sealed interfaces to match against endpoint associated errors

mpritham:
## Before this PR
In #2401 we added support for defining endpoint associated errors. This allowed Conjure service owners to define the errors that clients should expect to handle. This work was extended in #2435 where Dialogue client interfaces were generated to return sealed interfaces permitting records that corresponded to each endpoint error, and a "success" type, acting as a "result" type.

Initially the stance was that endpoint errors defined by services *need* to be handled by clients: using result types (or checked exceptions) would nudge callers of a client's endpoint to handle all of the potential errors a service has associated with the endpoint. (This could be worked around by utility methods that extract specific "success" types out of the provided result type but that would be an intentional choice made by clients.) In practice, this is a restrictive stance that hurts adoption: client code does not always need to handle the errors that endpoints have associated but if they would like to, they should have a way to know what errors should be handled.

Another reason to not change the return type of all endpoints in all Dialogue client interfaces is that it could lead to ABI breaks.

## After this PR

For each endpoint that has associated errors in its definition, there will be a new sealed interface generated that allows devs to convert a generic `RemoteException` into one of the error-specific exceptions. The sealed interface allows devs on Java 21+ to use exhaustive pattern matching switch statements providing compile-time guarantees that they have handled all possible endpoint associated errors.

For example, consider this endpoint definition:

```yaml
ErrorService:
  name: Error Service
  package: com.palantir.product
  default-auth: header
  base-path: /errors
  endpoints:
    testBasicError:
      http: POST /basic
      args:
        shouldThrowError: boolean
      returns: string
      errors:
        - InvalidArgument
        - ConflictingCauseSafeArgErr
```

The following Dialogue client interface will be generated:

```java
public interface ErrorServiceBlocking {
    /** @apiNote {@code POST /errors/basic} */
    @ClientEndpoint(method = "POST", path = "/errors/basic")
    String testBasicError(AuthHeader authHeader, boolean shouldThrowError);
    
    // <various other methods not relevant to this feature>
    
    sealed interface TestBasicErrorErrors {
        static TestBasicErrorErrors from(RemoteException e) {
            if (e instanceof TestErrors.InvalidArgumentException ex) {
                return new InvalidArgument(ex);
            } else if (e instanceof ConjureErrors.ConflictingCauseSafeArgErrException ex) {
                return new ConflictingCauseSafeArgErr(ex);
            } else {
                return new Unknown(e);
            }
        }

        record InvalidArgument(TestErrors.InvalidArgumentException e) implements TestBasicErrorErrors {}

        record ConflictingCauseSafeArgErr(ConjureErrors.ConflictingCauseSafeArgErrException e)
                implements TestBasicErrorErrors {}

        record Unknown(RemoteException e) implements TestBasicErrorErrors {}
    }
}
```

This allows clients to catch `RemoteException`s thrown by the call to `testBasicError` and convert it into a `TestBasicErrorErrors` object.

==COMMIT_MSG==
Generate sealed interfaces to match against endpoint associated errors
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
+++ b/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
@@ -16,7 +16,6 @@ import com.palantir.dialogue.PlainSerDe;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Serializer;
 import com.palantir.dialogue.TypeMarker;
-import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.palantir.tokens.auth.AuthHeader;
 import java.io.InputStream;
 import java.lang.Boolean;
@@ -274,7 +273,7 @@ public interface ErrorServiceBlocking {
             } else if (e instanceof ConjureErrors.ConflictingCauseSafeArgErrException ex) {
                 return new ConflictingCauseSafeArgErr(ex);
             } else {
-                throw new SafeIllegalArgumentException("Not an error associated with testBasicError", e);
+                return new Unknown(e);
             }
         }
 
@@ -282,6 +281,8 @@ public interface ErrorServiceBlocking {
 
         final record ConflictingCauseSafeArgErr(ConjureErrors.ConflictingCauseSafeArgErrException e)
                 implements TestBasicErrorErrors {}
+
+        final record Unknown(RemoteException e) implements TestBasicErrorErrors {}
     }
 
     sealed interface TestImportedErrorErrors {
@@ -289,12 +290,14 @@ public interface ErrorServiceBlocking {
             if (e instanceof EndpointSpecificErrors.EndpointErrorException ex) {
                 return new EndpointError(ex);
             } else {
-                throw new SafeIllegalArgumentException("Not an error associated with testImportedError", e);
+                return new Unknown(e);
             }
         }
 
         final record EndpointError(EndpointSpecificErrors.EndpointErrorException e)
                 implements TestImportedErrorErrors {}
+
+        final record Unknown(RemoteException e) implements TestImportedErrorErrors {}
     }
 
     sealed interface TestMultipleErrorsAndPackagesErrors {
@@ -312,7 +315,7 @@ public interface ErrorServiceBlocking {
             } else if (e instanceof TestErrors.ComplicatedParametersException ex) {
                 return new ComplicatedParameters(ex);
             } else {
-                throw new SafeIllegalArgumentException("Not an error associated with testMultipleErrorsAndPackages", e);
+                return new Unknown(e);
             }
         }
 
@@ -330,6 +333,8 @@ public interface ErrorServiceBlocking {
 
         final record ComplicatedParameters(TestErrors.ComplicatedParametersException e)
                 implements TestMultipleErrorsAndPackagesErrors {}
+
+        final record Unknown(RemoteException e) implements TestMultipleErrorsAndPackagesErrors {}
     }
 
     sealed interface TestEmptyBodyErrors {
@@ -337,11 +342,13 @@ public interface ErrorServiceBlocking {
             if (e instanceof TestErrors.InvalidArgumentException ex) {
                 return new InvalidArgument(ex);
             } else {
-                throw new SafeIllegalArgumentException("Not an error associated with testEmptyBody", e);
+                return new Unknown(e);
             }
         }
 
         final record InvalidArgument(TestErrors.InvalidArgumentException e) implements TestEmptyBodyErrors {}
+
+        final record Unknown(RemoteException e) implements TestEmptyBodyErrors {}
     }
 
     sealed interface TestBinaryErrors {
@@ -349,11 +356,13 @@ public interface ErrorServiceBlocking {
             if (e instanceof TestErrors.InvalidArgumentException ex) {
                 return new InvalidArgument(ex);
             } else {
-                throw new SafeIllegalArgumentException("Not an error associated with testBinary", e);
+                return new Unknown(e);
             }
         }
 
         final record InvalidArgument(TestErrors.InvalidArgumentException e) implements TestBinaryErrors {}
+
+        final record Unknown(RemoteException e) implements TestBinaryErrors {}
     }
 
     sealed interface TestOptionalBinaryErrors {
@@ -361,10 +370,12 @@ public interface ErrorServiceBlocking {
             if (e instanceof TestErrors.InvalidArgumentException ex) {
                 return new InvalidArgument(ex);
             } else {
-                throw new SafeIllegalArgumentException("Not an error associated with testOptionalBinary", e);
+                return new Unknown(e);
             }
         }
 
         final record InvalidArgument(TestErrors.InvalidArgumentException e) implements TestOptionalBinaryErrors {}
+
+        final record Unknown(RemoteException e) implements TestOptionalBinaryErrors {}
     }
 }
--- a/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
+++ b/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
@@ -277,12 +277,12 @@ public interface ErrorServiceBlocking {
             }
         }
 
-        final record InvalidArgument(TestErrors.InvalidArgumentException e) implements TestBasicErrorErrors {}
+        record InvalidArgument(TestErrors.InvalidArgumentException e) implements TestBasicErrorErrors {}
 
-        final record ConflictingCauseSafeArgErr(ConjureErrors.ConflictingCauseSafeArgErrException e)
+        record ConflictingCauseSafeArgErr(ConjureErrors.ConflictingCauseSafeArgErrException e)
                 implements TestBasicErrorErrors {}
 
-        final record Unknown(RemoteException e) implements TestBasicErrorErrors {}
+        record Unknown(RemoteException e) implements TestBasicErrorErrors {}
     }
 
     sealed interface TestImportedErrorErrors {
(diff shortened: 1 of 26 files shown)


conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
```diff
@@ -0,0 +1,378 @@
...
+
+        record InvalidArgument(TestErrors.InvalidArgumentException e) implements TestBasicErrorErrors {}
+
+        record ConflictingCauseSafeArgErr(ConjureErrors.ConflictingCauseSafeArgErrException e)
+                implements TestBasicErrorErrors {}
+
+        record Unknown(RemoteException e) implements TestBasicErrorErrors {}
+    }
+
+    sealed interface TestImportedErrorErrors {
```
mpritham: Probably nicer for the suffix to be singular.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueInterfaceGenerator.java
```diff
@@ -191,4 +202,67 @@ private MethodSpec apiMethod(EndpointDefinition endpointDef, StaticFactoryMethod
...
+        CodeBlock.Builder codeBlock = CodeBlock.builder();
+        for (EndpointError error : endpointDef.getErrors()) {
+            String errorName = error.getError().getName();
+            String errorPackage = error.getError().getPackage();
+            ClassName exceptionClassName = ClassName.get(
+                    Packages.getPrefixedPackage(errorPackage, options.packagePrefix()),
+                    ErrorGenerationUtils.errorTypesClassName(error.getError().getNamespace()),
+                    ErrorGenerationUtils.errorExceptionClassName(errorName));
+            TypeSpec errorRecord = createRecordForEndpointErrorUtility(errorName, exceptionClassName, utiltyClassName);
+            // Not strictly required. Seems a bit cleaner to not have it.
```
mpritham: This is because of [JLS-9.1.4](https://docs.oracle.com/javase/specs/jls/se21/html/jls-9.html#jls-9.1.4). I'll remove the comment.

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
mpritham: For reviewers: this file (or the async interface) contains the relevant newly generated code.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueInterfaceGenerator.java
mpritham: I suspect these sealed interfaces are small so it might be quite easy to just see all of the sub-interfaces. For things like the generated union classes (which were much larger), I think it's nice to have the list of permitted interfaces in one line at a glance. I can be convinced that these sealed interfaces could be large for endpoints with loads of associated errors and we should add them. Not super opinionated here.

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
```diff
@@ -0,0 +1,378 @@
...
+            if (e instanceof EndpointSpecificErrors.EndpointErrorException ex) {
+                return new EndpointError(ex);
+            } else {
+                return new Unknown(e);
+            }
+        }
+
+        record EndpointError(EndpointSpecificErrors.EndpointErrorException e) implements TestImportedErrorErrors {}
+
+        record Unknown(RemoteException e) implements TestImportedErrorErrors {}
```
mpritham: I initially did not have this `Unknown` type, and just threw an exception if the `RemoteException` wasn't an endpoint associated error, but I think this will lead to nested `try-catch` blocks in code. Devs will likely want to handle the "unknown" case anyway so creating a separate record for it that they can match on makes sense.

conjure-java-core/src/test/java/com/palantir/conjure/java/endpointerrors/UndertowServiceEteTest.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    public void test_endpoint_error_utility() {
+        try {
+            errorServiceClient.testMultipleErrorsAndPackages(AUTH_HEADER, Optional.of("invalidArgument"));
+        } catch (RemoteException e) {
+            TestBasicErrorErrors error = ErrorServiceBlocking.TestBasicErrorErrors.from(e);
+            assertThat(error).isInstanceOfSatisfying(TestBasicErrorErrors.InvalidArgument.class, invalidArgument -> {
+                assertThat(invalidArgument.e().error().parameters().field()).isEqualTo("field");
```
mpritham: I can change this to `.exception`.

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
kunalrkak: Singular definitely makes sense since the actual instance will be a single type. Also thoughts on `ErrorType` or something for the suffix? Was also thinking `AssociatedError`, but that feels like a mouthful

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
```diff
@@ -0,0 +1,378 @@
...
+            return ErrorServiceBlocking.of(endpointChannelFactory, runtime);
+        }
+    }
+
+    sealed interface TestBasicErrorErrors {
+        static TestBasicErrorErrors from(RemoteException e) {
+            if (e instanceof TestErrors.InvalidArgumentException ex) {
+                return new InvalidArgument(ex);
+            } else if (e instanceof ConjureErrors.ConflictingCauseSafeArgErrException ex) {
+                return new ConflictingCauseSafeArgErr(ex);
```
kunalrkak: Aren't the series of `instanceof` checks unideal from a performance standpoint? If we were on source 21+ I would say a switch statement would be an easy replacement over a visitor

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
```diff
@@ -0,0 +1,378 @@
...
+            } else {
+                return new Unknown(e);
+            }
+        }
+
+        record InvalidArgument(TestErrors.InvalidArgumentException e) implements TestBasicErrorErrors {}
```
kunalrkak: nit: we should name the parameter `error` or something so that the getter method is not `.e()`

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueInterfaceGenerator.java
```diff
@@ -191,4 +202,67 @@ private MethodSpec apiMethod(EndpointDefinition endpointDef, StaticFactoryMethod
...
     }
+
+    private Optional<TypeSpec> endpointErrorUtilityType(
+            EndpointDefinition endpointDef, String packageName, ClassName className) {
+        ClassName utiltyClassName = ClassName.get(
+                packageName,
+                className.simpleName(),
+                CaseFormat.LOWER_CAMEL.to(
+                                CaseFormat.UPPER_CAMEL,
+                                endpointDef.getEndpointName().get()) + "Errors");
```
kunalrkak: I think you could use `ClassName#nestedClass` here

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
aldexis: Do we have an option without `instanceof` in the absence of source 21+? Visitor patterns with overloads would have to rely on the `RemoteException` having an `accept` method to take a visitor, which simply won't work here 🤔

aldexis (commented):
Didn't look through the codegen pieces, but at the generated code. The new helpers seem like a good addition and shouldn't break anything.

The only ABI risks I could find are 1. having it adopted and reverting usage of the flag 2. Removing associations between errors and endpoints (even with the error still existing) I think this is probably fine.

I'm not entirely sure the abi checker would protect us from the removal in all cases however, specifically if the usage looks like
```
switch(TestImportedErrorErrors.from(exception)) {
  EndpointError -> log.error("Found an endpoint error");
  Unknown -> log.error("Found an unknown error");
}
```
where the class itself isn't used anywhere

Beyond this, I agree with Kunal's comments and will leave the deeper review to him

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
mpritham: Yeah, I think we can actually match on the error name here. I can make that change.

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceBlocking.java
mpritham: hm, what are your thoughts on just keeping the name `<EndpointName>Error`? Feel like it's cleaner.

kunalrkak: sure works for me

kunalrkak (dismissed):
Thanks!

kunalrkak (approved):

PR #2697, opened 2025-10-30
Log 401 exceptions at debug

pkoenig10:
Unauthenticated requests are not interesting to log and can trivially be created by malicious users who aren't even able to log in. I tend to view unauthenticated requests similar to requests against unmatched paths, which we also do not emit logs for.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
@@ -94,7 +94,7 @@ public enum ConjureExceptions implements ExceptionHandler {
                     throwable);
         } else {
             ServiceException exception = new ServiceException(ErrorType.INTERNAL, throwable);
-            log(exception, throwable);
+            log(exception, throwable, false);
             writeResponse(
                     exchange,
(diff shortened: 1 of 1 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java
```diff
@@ -191,12 +191,21 @@ private static boolean qosExceptionHasAdditionalMetadata(QosException qosExcepti
     // error types don't propagate through other services.
     private static void remoteException(HttpServerExchange exchange, RemoteException remoteException) {
         if (remoteException.getStatus() == 401 || remoteException.getStatus() == 403) {
-            log.info(
-                    "Encountered a remote exception",
-                    SafeArg.of("errorInstanceId", remoteException.getError().errorInstanceId()),
-                    SafeArg.of("errorName", remoteException.getError().errorName()),
-                    SafeArg.of("statusCode", remoteException.getStatus()),
-                    remoteException);
+            if (remoteException.getStatus() == 401 && log.isDebugEnabled()) {
```
schlosna: Wouldn't this still log 401s at INFO? Do we want:

```suggestion
            if (remoteException.getStatus() == 401 || log.isDebugEnabled()) {
```

schlosna (approved):

PR #2698, opened 2025-10-30
Optimize bulk operations on primitive lists

westinrm:
## Before this PR
Lots of boxing and slow iteration when operating on primitive lists.

## After this PR
==COMMIT_MSG==
Optimize bulk operations on primitive lists

- addAllAndCheckNonNull skips null checks for primitive lists
- inserting a primitive list in the middle of a primitive list creates intermediate int[] using arraycopy
==COMMIT_MSG==

### DIFF
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
@@ -83,6 +83,13 @@ public final class ConjureCollections {
 
     @SuppressWarnings("unchecked")
     public static <T> void addAllAndCheckNonNull(Collection<T> addTo, Iterable<? extends T> elementsToAdd) {
+        if (elementsToAdd instanceof ConjureIntegerList
+                || elementsToAdd instanceof ConjureDoubleList
+                || elementsToAdd instanceof ConjureSafeLongList) {
+            // Primitive lists will never have null elements.
+            addAll(addTo, elementsToAdd);
+            return;
+        }
         Preconditions.checkNotNull(elementsToAdd, "elementsToAdd cannot be null");
         // If we know the number of elements we are adding and the addTo Collection is an ArrayList, we can eagerly
         // resize it to only do one grow() of the array.
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
@@ -57,6 +57,11 @@ final class ConjureDoubleList extends AbstractList<Double> implements RandomAcce
 
     @Override
     public boolean addAll(int index, Collection<? extends Double> collection) {
+        if (collection instanceof ConjureDoubleList primitives) {
+            return index == size() - 1
+                    ? delegate.addAll(primitives.delegate)
+                    : delegate.addAllAtIndex(index, primitives.delegate);
+        }
         double[] target = new double[collection.size()];
         Iterate.forEachWithIndex(collection, (each, parameter) -> target[parameter] = each.doubleValue());
         return delegate.addAllAtIndex(index, target);
--- a/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
+++ b/conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
@@ -58,9 +58,7 @@ final class ConjureDoubleList extends AbstractList<Double> implements RandomAcce
     @Override
     public boolean addAll(int index, Collection<? extends Double> collection) {
         if (collection instanceof ConjureDoubleList primitives) {
-            return index == size() - 1
-                    ? delegate.addAll(primitives.delegate)
-                    : delegate.addAllAtIndex(index, primitives.delegate);
+            return delegate.addAllAtIndex(index, primitives.delegate);
         }
         double[] target = new double[collection.size()];
         Iterate.forEachWithIndex(collection, (each, parameter) -> target[parameter] = each.doubleValue());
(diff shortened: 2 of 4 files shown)


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
```diff
@@ -57,6 +57,11 @@ public Double get(int index) {
 
     @Override
     public boolean addAll(int index, Collection<? extends Double> collection) {
+        if (collection instanceof ConjureDoubleList primitives) {
+            return index == size() - 1
+                    ? delegate.addAll(primitives.delegate)
```
westinrm: this check is unfortunate, but only the append method is optimized.  If we wanted to track down a change to eclipse collections to make `addAllAtIndex` do that optimization we could.  I'd also be okay skipping the append optimization, because practically it's likely to never happen due to the mentioned unmodifiable issues.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -83,6 +83,13 @@ public static <T> void addAll(Collection<T> addTo, Iterable<? extends T> element
 
     @SuppressWarnings("unchecked")
     public static <T> void addAllAndCheckNonNull(Collection<T> addTo, Iterable<? extends T> elementsToAdd) {
+        if (elementsToAdd instanceof ConjureIntegerList
```
mpritham, line 86: It feels a bit weird to have this method be named `addAllAndCheckNonNull` and have it _not_ check that elements are non-null. It'd be nice if we can change the callers of `addAllAndCheckNonNull` to use `addAll` if we're using primitive collections.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
```diff
@@ -83,6 +83,13 @@ public static <T> void addAll(Collection<T> addTo, Iterable<? extends T> element
...
+                || elementsToAdd instanceof ConjureDoubleList
+                || elementsToAdd instanceof ConjureSafeLongList) {
+            // Primitive lists will never have null elements.
+            addAll(addTo, elementsToAdd);
```
mpritham, line 90: We're skipping [this logic](https://github.com/palantir/safe-logging/blob/a055c3e3ced0b243ac226f5dfa160804eb6a0401/preconditions/src/main/java/com/palantir/logsafe/Preconditions.java#L320) for each element in the list. I wonder if this is actually a significant source of latency here? It'd be cool to have a jmh benchmark to validate that this gives us a performance gain we're expecting.

westinrm: That logic isn't the slow part -- the slow part is the adapting `long` to `SafeLong` since it doesn't appear to get JIT'd to a no-op like boxing.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
westinrm: That's not possible because the type is always `Collection<T>` in the signatures (and it _must_ be since the conjure collections are package private). So it's possible to know statically what the type is.

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
aldexis: Is this right? Shouldn't it be
```suggestion
            return index == size()
                    ? delegate.addAll(primitives.delegate)
```
?
for instance, taking `size == 1`, you'd be appending if trying to insert at index 0, which seems wrong?

If `index == size()`, what are you trying to optimize? Looking at the code for `IntArrayList` for instance, this would do
```
    public boolean addAllAtIndex(int index, int... source)
    {
        if (index > this.size || index < 0)
        {
            this.throwOutOfBounds(index);
        }
        if (source.length == 0)
        {
            return false;
        }
        int sourceSize = source.length;
        int newSize = this.size + sourceSize;
        this.ensureCapacity(newSize);
        this.shiftElementsAtIndex(index, sourceSize);
        System.arraycopy(source, 0, this.items, index, sourceSize);
        this.size = newSize;
        return true;
    }
```
and
```
    private void shiftElementsAtIndex(int index, int sourceSize)
    {
        int numberToMove = this.size - index;
        if (numberToMove > 0)
        {
            System.arraycopy(this.items, index, this.items, index + sourceSize, numberToMove);
        }
    }
```
`numberToMove` would be 0 here, so there would be nothing to do, and we'd end up doing effectively the same thing as `addAll`?

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureCollections.java
aldexis: Could you clarify which adapting you're referring to here? Which scenario exactly are you optimizing for?

westinrm: inlining some of the code here you get this:
```java
for (int i = 0; i < longArrayList.size(); i++) {
  long raw = longArrayList.get(i);
  SafeLong adapted = SafeLong.of(raw); /* It doesn't appear that the JIT erases this allocation */
  Preconditions.checkNotNull(adapted);
}
```
or this:
```java
for (int i = 0; i < intArrayList.size(); i++) {
  int raw = intArrayList.get(i);
  Long adapted = raw; /* It appears like the JIT erases this allocation */
  Preconditions.checkNotNull(adapted);
}
```

conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureDoubleList.java
westinrm: good catch.  I did look at that method and thought about "is it size-1 or size" but apparently I still messed it up.

westinrm: since this isn't tested code I just removed this optimization, since it doesn't really matter due to:
> appending a primitive list _still_ uses an intermediate int[] array due to the `MutableIntList.addAll` method only optimizing adding mutable IntArrayList, not UnmodifiableIntList

westinrm:
the `toArray` allocation used by ECJ's addAll method is likely just as problematic as the boxing/SafeLong.of wrapping done using `addAllAndCheckNonNull` so I'm going to close this.  If someone tracks down optimizations to ECJ's addAll method then reopening this is probably worthwhile.

PR #2707, opened 2025-11-07
Fix sealed union deserialization for camel case fields

kunalrkak:
## Before this PR

We were converting field names to fully lowercase for the subtypes annotation - which prevented Jackson from matching on the correct sub type at deserialization time.

## After this PR

==COMMIT_MSG==
Generated `@JsonSubTypes.Type` annotation has correct camel casing for field names instead of lowercase. This allows sealed unions to deserialize correctly.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
@@ -231,8 +231,7 @@ public final class UnionGenerator {
                                 "name",
                                 "$S",
                                 // "unknown" is valid here since UnknownVariant is only used as a default
-                                StringUtils.lowerCase(
-                                        memberTypeDef.getFieldName().get())) // non-capitalized
+                                memberTypeDef.getFieldName().get())
                         .build())
                 .toList();
         AnnotationSpec.Builder annotationBuilder = AnnotationSpec.builder(JsonSubTypes.class);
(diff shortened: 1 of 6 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -231,8 +231,7 @@ private static AnnotationSpec generateJsonSubTypes(ClassName unionClass, List<Fi
                                 "name",
                                 "$S",
                                 // "unknown" is valid here since UnknownVariant is only used as a default
-                                StringUtils.lowerCase(
-                                        memberTypeDef.getFieldName().get())) // non-capitalized
+                                memberTypeDef.getFieldName().get())
```
aldexis, line 234: Was there a reason we had made this lowerCase in the first place?

aldexis (approved):
Approving to unblock, but curious whether there was a specific reason we had thought of lowercasing in the first place

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
kunalrkak: I think this was just an oversight - at first we were passing the subvariant's class name here and needed to "uncapitalize", but since I was just testing with short names I forgot about camel casing, so it was incorrect anyways. After we switched back to passing the field name this just got left over.

PR #2716, opened 2025-11-12
Create `@ConjureGenerated` Annotation

kunalrkak:
## Before this PR

Existing `@Generated` annotations have a retention policy of `SOURCE` and are discarded after compilation.
This makes it hard to later tell if an object is conjure generated or not, forcing us to try to pattern match based
on known characteristics, i.e. palantir/gradle-baseline#3344. Using an annotation with a `CLASS` retention 
policy would allow us to just look for the presence of that annotation on a class instead.

## After this PR

==COMMIT_MSG==
Replace `@Generated` annotations with `@ConjureGenerated` in generated code.
==COMMIT_MSG==


conjure-lib/src/main/java/com/palantir/conjure/java/lib/internal/ConjureGenerated.java
```diff
@@ -0,0 +1,35 @@
...
+ * Marks types that were generated by conjure-java code generators.
+ * This annotation uses {@code CLASS} retention to remain visible to tools that analyze compiled bytecode.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.TYPE)
+public @interface ConjureGenerated {
+    /**
+     * The fully qualified name of the code generator that produced the annotated type.
+     */
+    String value();
```
kunalrkak, line 34: I left this as a String to mirror the existing usage of `@Generated` but given we're exposing this more broadly, perhaps we should switch to an enum of supported types?

kunalrkak:
It looks like ErrorProne is wired up (with this [config](https://github.com/palantir/suppressible-error-prone/blob/8d37b22d7dd6d7083745ca9a213df0593d843362/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java#L196)) to skip over code tagged with the `@Ganerated` annotation, which is why it's now complaining about some of the integration input files.

PR #2718, opened 2025-11-13
Add test utility for error-specific `RemoteException` and `SerializableError`

mpritham:
## Before this PR
Previously devs would mock interactions with Dialogue clients that threw a `RemoteException` by creating a Mockito mock returning a constructed `RemoteException`. Now, clients throw subclasses of `RemoteException`. It is not possible to construct these error-specific `RemoteException`s because the error's `SerializableError` constructor (a required parameter when constructing the `RemoteException` subclass) is package private. I'd like this to remain package-private: devs should not be constructing these errors in their production code. Still, there should be an easy way for folks to construct these error-specific `RemoteException`s for use in unit tests.

## After this PR
For each error, generate a test utility class that is able to construct error specific `RemoteException`s and `SerializableError`s.

==COMMIT_MSG==
Add test utility for error-specific `RemoteException` and `SerializableError`
==COMMIT_MSG==

## Possible downsides?

I would like folks to only use this in tests. I've named the class `TestUtility` and provided a JavaDoc telling folks to only use this in tests. Still, it's possible for folks to construct and throw these Conjure-generated exceptions in their code if they really wanted to.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/TestErrors.java
+++ b/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/TestErrors.java
@@ -160,4 +160,57 @@ public final class TestErrors {
             return error;
         }
     }
+
+    /**
+     * This utility class is provided to make creating error-specific exceptions easier for tests. It should not be used
+     * outside of tests!
+     */
+    public static final class ComplicatedParametersTestUtility {
+        public static ComplicatedParametersSerializableError createSerializableError(
+                String errorInstanceId, ComplicatedParametersParameters parameters) {
+            return new ComplicatedParametersSerializableError(
+                    COMPLICATED_PARAMETERS.code().name(), COMPLICATED_PARAMETERS.name(), errorInstanceId, parameters);
+        }
+
+        public static ComplicatedParametersException createException(
+                String errorInstanceId, ComplicatedParametersParameters parameters) {
+            return new ComplicatedParametersException(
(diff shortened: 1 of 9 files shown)


conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/TestErrors.java
```diff
@@ -160,4 +160,57 @@ public NotFoundSerializableError error() {
             return error;
         }
     }
+
+    /**
+     * This utility class is provided to make creating error-specific exceptions easier for tests. It should not be used
+     * outside of tests!
+     */
+    public static final class ComplicatedParametersTestUtility {
```
aldexis, line 168: Another approach would be to have one `TestUtility` class per `XErrors` class. Makes the "Test" part of the name a bit more prominent.
I don't have strong feelings either way.

aldexis (commented):
If we want to make sure this doesn't get used outside of tests, we could go one step further, and create a separate `xxx-test-utilities` library, but this seems much more involved, so this feels reasonable to me

aldexis (dismissed):

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/TestErrors.java
mpritham: Think I'll merge with the current state. With the approach where we have one test utility, I think we'd have to name each method like `createXSerializableError`. I think it'd be slightly easier to search for usages of places folks have created these errors by having all the constructors have the same name.

aldexis (approved):

PR #2735, opened 2025-12-02
Remove unnecessary permits clauses

pkoenig10:
There is no need to use a `permits` clause for direct subclasses in the same compilation unit. These subclasses are always permitted.

https://docs.oracle.com/javase/specs/jls/se21/html/jls-8.html#jls-8.1.6


mpritham:
We [had decided](https://github.com/palantir/conjure-java/pull/2649#issuecomment-3353690025) to keep the permits clauses, despite not being necessary, to serve as documentation: when a dev navigates to the definition of a union in their IDE, they can see all of the unions variants at a glance without having to read through the implementation.

PR #2752, opened 2025-12-15
Add test to check that sealed unions are serialized the same way as legacy unions

mpritham:
## Before this PR
Add a small test to validate that the same Conjure union object is serialized the same way whether using the sealed classes implementation or the legacy implementation.

## After this PR

==COMMIT_MSG==
Add test to check that sealed unions are serialized the same way as legacy unions
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/test/java/com/palantir/conjure/java/types/UnionTests.java
+++ b/conjure-java-core/src/test/java/com/palantir/conjure/java/types/UnionTests.java
@@ -249,6 +249,17 @@ class UnionTests {
         assertThat(simpleUnion.toString()).contains(expected);
     }
 
+    @Test
+    void sealedUnionsAreSerializedTheSameAsLegacyUnions() throws IOException {
+        String expected = "foo";
+        SimpleUnion simpleUnion = SimpleUnion.foo(expected);
+        undertow.com.palantir.product.SimpleUnion simpleLegacyUnion =
+                undertow.com.palantir.product.SimpleUnion.foo(expected);
+        String serialized = MAPPER.writeValueAsString(simpleUnion);
+        String serializedLegacy = MAPPER.writeValueAsString(simpleLegacyUnion);
+        assertThat(serialized).isEqualTo(serializedLegacy);
+    }
+
     private Void failOnKnownType(String type, Object value) {
         Fail.fail(
                 "Visited known type when expected unknown type",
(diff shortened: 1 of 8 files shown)


conjure-java-core/src/test/java/com/palantir/conjure/java/types/UnionTests.java
```diff
@@ -249,6 +249,17 @@ void sealedUnionCanUseAllExistingPatterns() {
         assertThat(simpleUnion.toString()).contains(expected);
     }
 
+    @Test
+    void sealedUnionsAreSerializedTheSameAsLegacyUnions() throws IOException {
```
mpritham: Here is the test. The remainder of the code is modifying `SimpleUnion` in `ete-service.yml` to match the one in `example-unions.yml`

kunalrkak:
@mpritham We have a bunch of these serde tests [here](https://github.com/palantir/conjure-java/blob/e27a5407443ee17e6bff5aed40420d26973ea851/conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java#L865-L875), with the expected output hardcoded (since it should not change) - but if you want to add a object <> object test, that file feels like a better place for it!

mpritham:
Sounds good. Moved.

conjure-java-core/src/test/java/com/palantir/conjure/java/types/WireFormatTests.java
```diff
@@ -937,6 +937,12 @@ void testSealedInterfaceUnionType_deserialize_CamelCase() throws JsonProcessingE
                 .isEqualTo(CamelCaseUnion.camelCasedField("test"));
     }
 
+    @Test
+    void testSealedInterfaceUnionType_wireFormatBackCompatWithLegacyImpl() throws IOException {
+        assertThat(mapper.writeValueAsString(SimpleUnion.foo("foo")))
+                .isEqualTo(mapper.writeValueAsString(undertow.com.palantir.product.SimpleUnion.foo("foo")));
```
aldexis: It's not immediately clear which one is the sealed interface and which one isn't. Should we add some verification that one is sealed and one isn't? (to avoid test refactors breaking our assumptions by making both sealed or both unsealed)

mpritham: I'm importing the `SimpleUnion` from the `sealedunions` package now

aldexis (approved):

PR #2753, opened 2025-12-15
[Sealed unions] Make `toString` backward compatible

mpritham:
## Before this PR
The union objects generated when the `sealedUnions` flag is set create `toString` methods that are not backward-compatible with objects generated without the `sealedUnions` flag. While in general `toString` method implementation is _not_ part of the Conjure API, it's current use in the serialization of error parameters makes rolling out changes to `toString` a wire break. 

## After this PR

==COMMIT_MSG==
[Sealed unions] Make `toString` backward compatible
==COMMIT_MSG==

## Possible downsides?
Clunky and inaccurate. The `toString` representation of union objects no longer accurately represents the generated object: there will be `<Variant`>Wrapper` strings in the `toString` representation which do not correspond to any class in the sealed class implementation.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
@@ -165,9 +165,9 @@ public final class UnionGenerator {
 
         ClassName baseClass = unionClass.nestedClass("Base");
 
-        List<FieldSpec> fields =
-                ImmutableList.of(FieldSpec.builder(baseClass, VALUE_FIELD_NAME, Modifier.PRIVATE, Modifier.FINAL)
-                        .build());
+        FieldSpec valueField = FieldSpec.builder(baseClass, VALUE_FIELD_NAME, Modifier.PRIVATE, Modifier.FINAL)
+                .build();
+        List<FieldSpec> valueFieldList = ImmutableList.of(valueField);
         List<AnnotationSpec> safety =
                 ConjureAnnotations.safety(safetyEvaluator.evaluate(TypeDefinition.union(typeDef)));
 
@@ -176,7 +176,7 @@ public final class UnionGenerator {
                 .addAnnotations(safety)
                 .addAnnotation(ConjureAnnotations.getConjureGeneratedAnnotation(UnionGenerator.class))
                 .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
-                .addFields(fields)
+                .addFields(valueFieldList)
                 .addMethod(generateConstructor(baseClass))
                 .addMethod(generateGetValue(baseClass))
                 .addMethods(
@@ -194,11 +194,11 @@ public final class UnionGenerator {
                         typeMapper, typesMap, baseClass, visitorClass, typeDef.getUnion(), options))
                 .addType(generateUnknownWrapper(baseClass, visitorClass, options))
                 .addMethod(MethodSpecs.createEquals(unionClass))
-                .addMethod(MethodSpecs.createEqualTo(unionClass, fields))
-                .addMethod(MethodSpecs.createHashCode(fields))
+                .addMethod(MethodSpecs.createEqualTo(unionClass, valueFieldList))
+                .addMethod(MethodSpecs.createHashCode(valueFieldList))
                 .addMethod(MethodSpecs.createToString(
                                 unionClass.simpleName(),
-                                fields.stream()
+                                valueFieldList.stream()
                                         .map(fieldSpec -> FieldName.of(fieldSpec.name()))
                                         .collect(Collectors.toList()))
                         .toBuilder()
@@ -211,6 +211,36 @@ public final class UnionGenerator {
                 .skipJavaLangImports(true)
                 .indent("    ")
                 .build();
+    }
+
+    private static MethodSpec createToString(
+            boolean sealedUnions, ClassName baseClassName, ClassName wrapperClassName) {
+        if (!sealedUnions) {
+            return MethodSpec.methodBuilder("toString")
+                    .addAnnotation(Override.class)
+                    .addModifiers(Modifier.PUBLIC)
+                    .returns(ClassName.get(String.class))
+                    .addCode(CodeBlock.builder()
+                            .add("return $S\n", wrapperClassName.simpleName() + '{' + VALUE_FIELD_NAME + ": ")
+                            .add(" + $N", VALUE_FIELD_NAME)
+                            .add(" + '}';")
+                            .build())
+                    .build();
+        } else {
+            return MethodSpec.methodBuilder("toString")
+                    .addAnnotation(Override.class)
+                    .addModifiers(Modifier.PUBLIC)
+                    .returns(ClassName.get(String.class))
+                    .addCode(CodeBlock.builder()
+                            .add(
+                                    "return $S\n",
+                                    baseClassName.simpleName() + '{' + VALUE_FIELD_NAME + ": "
+                                            + wrapperClassName.simpleName() + '{' + VALUE_FIELD_NAME + ": ")
+                            .add(" + $N", VALUE_FIELD_NAME)
+                            .add(" + \"}}\";")
+                            .build())
+                    .build();
+        }
     }
 
     private static AnnotationSpec generateJsonTypeInfo(ClassName unknownVariant) {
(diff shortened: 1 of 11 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -194,11 +194,11 @@ public static JavaFile generateUnionType(
...
                 .addType(generateUnknownWrapper(baseClass, visitorClass, options))
                 .addMethod(MethodSpecs.createEquals(unionClass))
-                .addMethod(MethodSpecs.createEqualTo(unionClass, fields))
-                .addMethod(MethodSpecs.createHashCode(fields))
+                .addMethod(MethodSpecs.createEqualTo(unionClass, valueFieldList))
+                .addMethod(MethodSpecs.createHashCode(valueFieldList))
                 .addMethod(MethodSpecs.createToString(
                                 unionClass.simpleName(),
-                                fields.stream()
+                                valueFieldList.stream()
```
mpritham: Unneeded traversal.

aldexis: Not sure I understand this comment, nor why you made the above change, as it doesn't seem like you're using `valueField` anywhere?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -213,6 +213,46 @@ public static JavaFile generateUnionType(
...
+                    .addAnnotation(Override.class)
+                    .addModifiers(Modifier.PUBLIC)
+                    .returns(ClassName.get(String.class))
+                    .addCode(CodeBlock.builder()
+                            .add("return $S\n", wrapperClassName.simpleName() + '{' + VALUE_FIELD_NAME + ": ")
+                            .add(" + $N", VALUE_FIELD_NAME)
+                            .add(" + '}';")
+                            .build())
+                    .build();
+        }
```
aldexis: _Something_ slightly bugs me in that we are hardcoding the expectation that there is only a single field. I'm not seeing this change in the near-term, but I'm wondering whether we could somehow rely on `MethodSpecs.createToString` or at least have a similar logic?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -213,6 +213,46 @@ public static JavaFile generateUnionType(
...
+                            .add(
+                                    "return $S\n",
+                                    baseClassName.simpleName() + '{' + VALUE_FIELD_NAME + ": "
+                                            + wrapperClassName.simpleName() + '{' + VALUE_FIELD_NAME + ": ")
+                            .add(" + $N", VALUE_FIELD_NAME)
+                            .add(" + \"}}\";")
+                            .build())
+                    .addJavadoc(
+                            "This method is not part of Conjure API. Users should not rely on consistent generation of"
+                                    + " the toString method between versions of Conjure.")
```
aldexis: Should we just always have this javadoc?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -213,6 +213,46 @@ public static JavaFile generateUnionType(
...
 
+    /** Creates a backward compatible toString method. While in general, Conjure-Java does not provide any guarantees on
+     * the toString representation of generated objects, the toString method is currently used when serializing error
+     * parameters when JSON is not explicitly used. This is clunky because the sealedUnions implementation does not
+     * generate any classes suffixed with `Wrapper`, yet such classes show up in the `toString` implementation. In a
+     * future Conjure-Java change, we can update the toString method here to better represent the variant in the
+     * sealedUnions implementation.
+     */
+    private static MethodSpec createToString(
+            boolean sealedUnions, ClassName baseClassName, ClassName wrapperClassName) {
```
aldexis: Can we add some tests to guarantee that this stays true?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -213,6 +213,46 @@ public static JavaFile generateUnionType(
...
+                                    baseClassName.simpleName() + '{' + VALUE_FIELD_NAME + ": "
+                                            + wrapperClassName.simpleName() + '{' + VALUE_FIELD_NAME + ": ")
+                            .add(" + $N", VALUE_FIELD_NAME)
+                            .add(" + \"}}\";")
+                            .build())
+                    .addJavadoc(
+                            "This method is not part of Conjure API. Users should not rely on consistent generation of"
+                                    + " the toString method between versions of Conjure.")
+                    .build();
+        } else {
```
kunalrkak: why would we not continue to use MethodSpecs#createToString for at least the old union types?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
kunalrkak: I'm not sure we should include this javadoc in the generation, feels a bit out of place just being on sealed union types. As long as we have a plan to migrate away from the old toString, we should just revert this later once safe.

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
mpritham: Updated. I was trying to create a method that handles only the case where we have `value` as a field, but it ended up being not as clean as just using the old logic in `MethodSpecs`

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
mpritham: A few ideas:
- Have some nested structure passed to `MethodSpecs.createToString` that can represent arbitrarily nested lists

```java
sealed interface ToStringField permits SimpleField, NestedField {}
record SimpleField(FieldName name) implements ToStringField {}
record NestedField(FieldName name, List<ToStringField> children) implements ToStringField {}
```

I like the design of this, but I didn't follow through with the idea because it made things more complicated for just this one use-case. I think we should explore this if we need to make other nested `toString` methods.

- Pass in a map that provides a `CodeBlock` to`MethodSpecs.createToString` containing the representation of the field.
```java
private static CodeBlock toStringConcatenation(
          String thisClassName,
          List<FieldName> fieldNames,
          Map<FieldName, CodeBlock> customFieldBlocks) {...}
```

This feels like a worse version of the idea above: the caller has to copy the logic in `MethodSpecs.toStringConcatenation`.

Maybe there's another thing you had in mind?

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
aldexis: This makes it somewhat explicit that people shouldn't be relying on toString's format in general, rather than just because of this specific case

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
aldexis: Having special handling for the legacy case seems fine actually. I can't seem to figure out a good way that isn't over-engineered. Relying on the existing method for the normal case seems better though (which I see you've done).

One thing we _could_ add would be a `MethodSpecs.createToStringBuilder(CodeBlock) -> MethodSpec.Builder` to encode the 
```java
MethodSpec.methodBuilder("toString")
                .addAnnotation(Override.class)
                .returns(ClassName.get(String.class))
```
part of the logic, but I don't feel strongly about it and it does clutter the code a bit

aldexis (commented):
The current approach seems reasonable to me, provided we do want to do this. If/when we confirm and you remove the do-not-merge, I'm happy to approve

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
mpritham: Adding this javadoc to all object `toString` methods is quite a large diff and somewhat unrelated to this PR. I can open a separate PR adding the javadoc to the generated code in other object types.

kunalrkak: As a counter point, there's a lot of things explicitly encoded in our conjure spec that aren't directly included in code/javadocs. Given this is true for all `toString`s, it feels better to document this once as opposed to on every conjure object's toString method. I think generally generated code should be just pure output based on the input conjure.

aldexis: That makes sense. And it's general bad practice to rely on a specific toString format too, so maybe let's not bother

kunalrkak (approved):
Thanks!

PR #2759, opened 2025-12-19
Fix unit test import

kunalrkak:
## Before this PR

We have other tests which expect`SimpleUnion` to be a sealed union

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


mpritham (approved):
Thanks!

PR #2761, opened 2026-01-05
Rethrow cookie parsing failure as IllegalArgumentException

pkoenig10:
## Before this PR

If a request contains more cookies than the configured maximum (200 by default), then the `HttpServerExchange` methods to obtain cookies will throw an `IllegalStateException`. This can result in false positives for internal errors monitors.

```
java.lang.IllegalStateException: {throwable0_message}
	at io.undertow.util.Cookies.createCookie(Cookies.java:397)
	at io.undertow.util.Cookies.parseCookie(Cookies.java:291)
	at io.undertow.util.Cookies.parseRequestCookies(Cookies.java:246)
	at io.undertow.util.Cookies.parseRequestCookies(Cookies.java:224)
	at io.undertow.util.Cookies.parseRequestCookies(Cookies.java:215)
	at io.undertow.server.HttpServerExchange.requestCookies(HttpServerExchange.java:1248)
	at io.undertow.server.HttpServerExchange.getRequestCookie(HttpServerExchange.java:1229)
	at com.palantir.conjure.java.undertow.runtime.ConjureContexts$ConjureServerRequestContext.cookie(ConjureContexts.java:85)
	...
```

## After this PR

Catch these exceptions and re-throw them as `SafeIllegalArgumentException`.

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
@@ -23,6 +23,7 @@ import com.palantir.conjure.java.undertow.lib.Contexts;
 import com.palantir.conjure.java.undertow.lib.Endpoint;
 import com.palantir.conjure.java.undertow.lib.RequestContext;
 import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import io.undertow.server.HttpServerExchange;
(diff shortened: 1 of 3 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
```diff
@@ -82,7 +83,17 @@ public Optional<String> firstHeader(String headerName) {
...
+            try {
+                cookie = exchange.getRequestCookie(cookieName);
+            } catch (IllegalStateException e) {
+                // An IllegalStateException is thrown if the number of request cookies exceeds the configured maximum
+                throw new SafeIllegalArgumentException("Failed to parse request cookies", e);
+            }
+            if (cookie == null) {
+                return Optional.empty();
+            }
+            return Optional.of(cookie.getValue());
```
kunalrkak, line 96: Maybe `Optional#ofNullable` here?

kunalrkak (commented):
One nit, otherwise lgtm!

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
pkoenig10: This value should never be null in a parsed cookie.

kunalrkak: hmm, how is that better implied in this code block vs using ofNullable? seems like we're handling it in the same way, just more verbose

pkoenig10: I personally find this way easier to read. It's consistent with the code structure other places we read cookies (the files touched in this PR) and avoids the lambda allocation.

kunalrkak (approved):
Thanks!

PR #2771, opened 2026-01-13
Add RequestContext method to read attachments

daanjanschouten:
## Before this PR

Adds an `attachment()` method to `RequestContext` to read `HttpServerExchange` attachments. This enables retrieving current request size from Witchcraft's `BytesReadHandler` in application code.

`RequestContext` is internal-only per its javadoc, so adding methods should be safe.

## After this PR

==COMMIT_MSG==
Add `RequestContext` method to read attachments.
==COMMIT_MSG==

### DIFF
--- a/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
+++ b/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
@@ -20,6 +20,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ListMultimap;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.Unsafe;
+import io.undertow.util.AttachmentKey;
 import java.security.cert.Certificate;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
```diff
@@ -50,6 +51,12 @@ public interface RequestContext {
      */
     Optional<String> firstHeader(String headerName);
 
+    /**
+     * Returns the attachment associated with the given {@code key}.
+     * An {@link Optional#empty()} is returned if no such attachment exists.
+     */
+    <T> Optional<T> attachment(AttachmentKey<T> key);
```
pkoenig10, line 58: I'm not sure this is appropriate for `RequestContext`.

`RequestContext` is designed to be a read-only interface for getting requests data not supported by the tool used to generate the endpoint implementations (either conjure or conjure-undertow-annotations). From the Javadoc for this interface:

> Interface providing a view over data provided by the the original HTTP request including request headers and query parameters.

This method is meaningfully different. It's not exposing information about the request - it's exposing information from the underlying `HttpServerExchange` for this request.

pkoenig10, line 58: > This enables retrieving current request size from Witchcraft's `BytesReadHandler` in application code.

I'm also not convinced that we want to make this part of the public API of Witchcraft. This increases the surface area of the Witchcraft API and removes a degree of freedom for evolving our implementation in the future if necessary.

PR #2812, opened 2026-02-23
Memoize safety evaluator

xanderbailey:
### Summary
- `SafetyEvaluator` now caches computed safety results per type, avoiding redundant recursive traversals of the type graph
- Results are only cached when no cycle was encountered during evaluation, preserving correctness for mutually-recursive type definitions

### Motivation

JFR profiling of `compileConjureObjects` on a large IR (~4600 types) showed that **74% of CPU time** was spent in `SafetyEvaluator`, with the top frames being `TypeDefinitionSafetyVisitor.with()` and `SafetyEvaluator.combine()`. Each call to `evaluate()` created a fresh visitor and re-traversed the entire reachable type graph from scratch. Since many types reference the same shared types, this resulted in massive redundant work — effectively O(n * m) where n is the number of types and m is the average reachable subgraph size.

### What changed

A `HashMap<TypeName, Optional<LogSafety>>` cache is shared across all evaluations on a `SafetyEvaluator` instance. Before traversing a type's subgraph, `with()` checks the cache and returns immediately on a hit. After evaluation, the result is cached — but only if no cycle-breaking occurred in the subtree, since the cycle-breaking heuristic (assume `SAFE`) produces results that depend on evaluation order and should not be persisted.

### Impact

Tested against a production IR with ~4600 types: `compileConjureObjects` dropped from **42s to 20s** in Gradle.

### Test plan
- [x] All existing conjure-java-core tests pass
- [x] Generated output verified identical to baseline against a production IR with ~4600 types

## After this PR

==COMMIT_MSG==
Memoize safety evaluator
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
@@ -38,6 +38,7 @@ import com.palantir.conjure.spec.TypeDefinition;
 import com.palantir.conjure.spec.TypeName;
 import com.palantir.conjure.spec.UnionDefinition;
 import com.palantir.logsafe.Preconditions;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -64,6 +65,7 @@ public final class SafetyEvaluator {
     public static final Optional<LogSafety> UNKNOWN_UNION_VARINT_SAFETY = Optional.empty();
 
     private final Map<TypeName, TypeDefinition> definitionMap;
+    private final Map<TypeName, Optional<LogSafety>> cache = new HashMap<>();
 
     public SafetyEvaluator(ConjureDefinition definition) {
         this(TypeFunctions.toTypesMap(definition));
@@ -75,12 +77,12 @@ public final class SafetyEvaluator {
 
     public Optional<LogSafety> evaluate(TypeDefinition def) {
         return Preconditions.checkNotNull(def, "TypeDefinition is required")
-                .accept(new TypeDefinitionSafetyVisitor(definitionMap, new HashSet<>()));
+                .accept(new TypeDefinitionSafetyVisitor(definitionMap, cache, new HashSet<>()));
     }
 
     public Optional<LogSafety> evaluate(Type type) {
         return Preconditions.checkNotNull(type, "TypeDefinition is required")
-                .accept(new TypeDefinitionSafetyVisitor(definitionMap, new HashSet<>()).fieldVisitor);
+                .accept(new TypeDefinitionSafetyVisitor(definitionMap, cache, new HashSet<>()).fieldVisitor);
     }
 
     public Optional<LogSafety> evaluate(Type type, Optional<LogSafety> declaredSafety) {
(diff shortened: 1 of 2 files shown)


xanderbailey:
Did a dry run to check the size of the cache for this large IR and got:
```
SafetyEvaluator cache: 3437 entries, estimated 268 KB
```

conjure-java-core/src/test/java/com/palantir/conjure/java/types/SafetyEvaluatorTest.java
```diff
@@ -366,6 +366,119 @@ void testEmptyEnum() {
...
+                        .type(Type.reference(BAR))
+                        .build())
+                .build());
+        TypeDefinition bar = TypeDefinition.object(ObjectDefinition.builder()
+                .typeName(BAR)
+                .fields(FieldDefinition.builder()
+                        .fieldName(FieldName.of("foo"))
+                        .type(Type.reference(FOO))
+                        .build())
+                .build());
```
aldexis, line 386: The fields are _not_ marked as safe. Shouldn't we be returning unknown/empty in this case? 🤔 (I don't think your change changes this, but this seems weird to me 🤔 )

conjure-java-core/src/test/java/com/palantir/conjure/java/types/SafetyEvaluatorTest.java
```diff
@@ -366,6 +366,119 @@ void testEmptyEnum() {
...
+        ConjureDefinition conjureDef =
+                ConjureDefinition.builder().version(1).types(foo).types(bar).build();
+        SafetyEvaluator evaluator = new SafetyEvaluator(conjureDef);
+        // Evaluation order should not matter for the result
+        assertThat(evaluator.evaluate(foo)).hasValue(LogSafety.SAFE);
+        assertThat(evaluator.evaluate(bar)).hasValue(LogSafety.SAFE);
```
aldexis, line 392: You're not checking the evaluation order here. Since you're already doing it in a dedicated test, might as well remove the comment?
```suggestion
        SafetyEvaluator evaluator = new SafetyEvaluator(conjureDef);
        assertThat(evaluator.evaluate(foo)).hasValue(LogSafety.SAFE);
        assertThat(evaluator.evaluate(bar)).hasValue(LogSafety.SAFE);
```

conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java
```diff
@@ -170,15 +184,42 @@ public Optional<LogSafety> visitUnknown(String unknownType) {
...
+
+            // Save and reset cycle state so we can detect cycles within this type's subtree only.
+            boolean previousCycleState = encounteredCycle;
+            encounteredCycle = false;
+
             Optional<LogSafety> result = task.get();
+
+            boolean subtreeHadCycle = encounteredCycle;
+            // Propagate cycle detection upward: if this subtree had a cycle, callers should know.
+            encounteredCycle = previousCycleState || subtreeHadCycle;
```
aldexis, line 209: It's slightly confusing (to me) that we store the cyclic state in a field, unset it before calling the method, then reset it afterwards.

How would you feel about returning a pair from this method e.g. `record(Optional<LogSafety> safety, boolean cycleDetected)` and then use that instead here? (and maybe make a second method that extracts the safety result for usage in the rest of the code)

Edit: We'd want to check the memory impact of doing these allocations though

xanderbailey: Yeah I wanted to keep this lean. Want me to try that way?

aldexis: If you have a bit of time to try this out, I'd be curious about the memory impact yeah (especially around the total allocations). If it's low enough, I feel like this would make the cognitive overhead of this piece of code a bit lower, and thus more maintainable.

If it's not, I think this should be fine

aldexis (approved):
Approving to unblock this PR @xanderbailey though I'd still be interested in seeing the impact of https://github.com/palantir/conjure-java/pull/2812#discussion_r2852103993

PR #2818, opened 2026-02-24
Remove expensive safety combination Table

aldexis:
## Before this PR
After https://github.com/palantir/conjure-java/pull/2812, JFRs show that we still spend a large amount of time getting results from the precomputed combination table. This is because it's dealing with objects, so we need to compute their hashcode and get the values in the map. Instead, we can use our own internal table, based on an array and the actual underlying enum values (similar to what an EnumMap would do, but mapping two keys instead).

image
image
image

## After this PR

==COMMIT_MSG==
Remove expensive safety combination Table
==COMMIT_MSG==


aldexis:
This didn't bring the benefits I hoped it would, so closing

PR #2836, opened 2026-03-12
Add service to MDC

pkoenig10:
We added this JWT claim in https://github.com/palantir/auth-tokens/pull/1331.

This will allow us to use include the service in logs.


mpritham (approved):
Thanks!

PR #2839, opened 2026-03-17
Set the exit code when running the Conjure Java CLI

mpritham:
## Before this PR
In #2737 we updated the `main()` method in `ConjureJavaCli` to use a non-deprecated PicoCli method.

Previously we had:

```java
public static void main(String[] args) {
    CommandLine.run(new ConjureJavaCli(), args);
}
```

which we changed to:

```java
public static void main(String[] args) {
    new CommandLine(new ConjureJavaCli()).execute(args);
}
```

The execute() method returns an int exit code and catches exceptions internally (see the "Exception Exit Codes" section of [these PicoCli release notes](https://github.com/remkop/picocli/releases/tag/v4.0.0)), while the previous run() method would propagate exceptions to the caller. This means that when code generation fails (e.g., a validation error in ErrorGenerator), the exception is printed to stderr but main() returns normally.

The first commit on this PR is a small reproducer. It can be run with:

```bash
java -cp "conjure-java/build/classes/java/main:/<path-to-jar-directory>/picocli-4.7.7.jar" com.palantir.conjure.java.cli.PicocliExceptionRepro
```

The output is this:

```
=== execute() without System.exit ===
java.lang.IllegalStateException: boom
	at com.palantir.conjure.java.cli.PicocliExceptionRepro.run(PicocliExceptionRepro.java:10)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2045)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2469)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2461)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2423)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2425)
	at picocli.CommandLine.execute(CommandLine.java:2174)
	at com.palantir.conjure.java.cli.PicocliExceptionRepro.main(PicocliExceptionRepro.java:16)
Returned exit code: 1
Execution continued past execute() — exception was swallowed!

=== CommandLine.run() (deprecated) ===
Exception propagated: boom
```

## After this PR

==COMMIT_MSG==
Set the exit code when running the Conjure Java CLI
==COMMIT_MSG==


mpritham:
We should probably have some tests in `gradle-conjure` that are able to catch this sort of regression.

pkoenig10:
Can we add a test in this repo where we run a new process that invokes `ConjureJavaCli.main` using the classpath of our current process?

pkoenig10 (approved):

PR #2846, opened 2026-03-23
Generate a file with `TypeMarker`s for each errors namespace

mpritham:
## Before this PR
Each Dialogue service interface constructs two `TypeMarker`s per Conjure error:

https://github.com/palantir/conjure-java/blob/8ed34f38d374a747501b74e789476b5cf8a865aa/conjure-java-core/src/integrationInput/java/exceptionthrowingdialogueinterfaces/com/palantir/product/EteServiceBlocking.java#L461-L466

The construction of these `TypeMarker`s are duplicated in every service in defined in the Conjure definition with the declared errors. For some projects with many errors and services, this leads to a sharp increase in the number of anonymous classes generated which creates load on build tooling: e.g. JARs need to be compressed and lots more classes need to be loaded.

## After this PR
A new `{Namespace}ErrorsTypeMarkers` class is generated per error namespace in the Dialogue output, holding
  shared public static final TypeMarker constants:

```java
  public final class ConjureErrorsTypeMarkers {
      public static final TypeMarker<ConjureErrors.ConflictingCauseSafeArgSerializableError>
              CONFLICTING_CAUSE_SAFE_ARG_SERIALIZABLE_ERROR = new TypeMarker<...>() {};
      // ...
      private ConjureErrorsTypeMarkers() {}
  }
```

  Service interfaces now reference these constants instead of creating anonymous classes:

```java
  .exception(
          ConjureErrors.CONFLICTING_CAUSE_SAFE_ARG.name(),
          ConjureErrorsTypeMarkers.CONFLICTING_CAUSE_SAFE_ARG_SERIALIZABLE_ERROR,
          ConjureErrorsTypeMarkers.CONFLICTING_CAUSE_SAFE_ARG_EXCEPTION)
```

The TypeMarkers class is generated into the `-dialogue` project (not the `-objects` output) so it doesn't require adding dialogue-target as a dependency to the objects project.

==COMMIT_MSG==
Generate a file with `TypeMarker`s for each errors namespace
==COMMIT_MSG==

## Possible downsides?
I would've preferred to generate these `TypeMarker` anonymous classes in each of the generated `<Error>SerializableError` and `<Error>Exception` classes, but the `TypeMarker` is defined in `dialogue-target` and I did not want to force all consumers to `-objects` JARs to require a new API dep on `dialogue-target`.

I don't think there is anything specific to dialogue about `TypeMarker`: in the future, we can consider moving this class out to another project.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/another/EndpointSpecificErrorsTypeMarkers.java
+++ b/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/another/EndpointSpecificErrorsTypeMarkers.java
@@ -4,7 +4,10 @@ import com.palantir.dialogue.ExceptionDeserializerArgs;
 import com.palantir.dialogue.TypeMarker;
 import javax.annotation.processing.Generated;
 
-/** Internal utility class used by generated Dialogue interfaces. Not intended for external use. */
+/**
+ * Internal utility class used by generated Dialogue interfaces. Not intended for external use. This class needs to be
+ * public because errors from a certain namespace can be used in Dialogue interfaces defined in any namespace.
+ */
 @Generated("com.palantir.conjure.java.services.dialogue.ErrorTypeMarkersGenerator")
 public final class EndpointSpecificErrorsTypeMarkers {
     private static final TypeMarker<EndpointSpecificErrors.DifferentPackageSerializableError>
--- a/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/another/EndpointSpecificErrorsTypeMarkers.java
+++ b/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/another/EndpointSpecificErrorsTypeMarkers.java
@@ -6,9 +6,11 @@ import javax.annotation.processing.Generated;
 @Generated("com.palantir.conjure.java.services.dialogue.ErrorTypeMarkersGenerator")
 public final class EndpointSpecificErrorsTypeMarkers {
     public static final TypeMarker<EndpointSpecificErrors.DifferentPackageSerializableError>
-            DifferentPackageSerializableError =
+            DIFFERENT_PACKAGE_SERIALIZABLE_ERROR =
                     new TypeMarker<EndpointSpecificErrors.DifferentPackageSerializableError>() {};
 
-    public static final TypeMarker<EndpointSpecificErrors.DifferentPackageException> DifferentPackageException =
+    public static final TypeMarker<EndpointSpecificErrors.DifferentPackageException> DIFFERENT_PACKAGE_EXCEPTION =
             new TypeMarker<EndpointSpecificErrors.DifferentPackageException>() {};
+
+    private EndpointSpecificErrorsTypeMarkers() {}
 }
--- a/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/another/EndpointSpecificErrorsTypeMarkers.java
+++ b/conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/another/EndpointSpecificErrorsTypeMarkers.java
@@ -1,17 +1,25 @@
 package endpointerrors.com.palantir.another;
 
+import com.palantir.dialogue.ExceptionDeserializerArgs;
 import com.palantir.dialogue.TypeMarker;
 import javax.annotation.processing.Generated;
 
 /** Internal utility class used by generated Dialogue interfaces. Not intended for external use. */
 @Generated("com.palantir.conjure.java.services.dialogue.ErrorTypeMarkersGenerator")
 public final class EndpointSpecificErrorsTypeMarkers {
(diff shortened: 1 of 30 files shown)


bjlaub (dismissed):
approving for RC

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/another/EndpointSpecificErrorsTypeMarkers.java
```diff
@@ -0,0 +1,17 @@
+package endpointerrors.com.palantir.another;
+
+import com.palantir.dialogue.TypeMarker;
+import javax.annotation.processing.Generated;
+
+/** Internal utility class used by generated Dialogue interfaces. Not intended for external use. */
+@Generated("com.palantir.conjure.java.services.dialogue.ErrorTypeMarkersGenerator")
+public final class EndpointSpecificErrorsTypeMarkers {
+    public static final TypeMarker<EndpointSpecificErrors.DifferentPackageSerializableError>
```
mpritham: These constants cannot be package-private. Errors can be used in a namespace other than the one they were defined in.

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceAsync.java
```diff
@@ -117,36 +117,34 @@ private static <T> ExceptionDeserializerArgs<T> createExceptionDeserializerArgs(
...
-                        new TypeMarker<
-                                endpointerrors.com.palantir.another.EndpointSpecificErrors
-                                        .DifferentPackageSerializableError>() {},
-                        new TypeMarker<
-                                endpointerrors.com.palantir.another.EndpointSpecificErrors
-                                        .DifferentPackageException>() {});
+                        endpointerrors.com.palantir.another.EndpointSpecificErrorsTypeMarkers
+                                .DIFFERENT_PACKAGE_SERIALIZABLE_ERROR,
+                        endpointerrors.com.palantir.another.EndpointSpecificErrorsTypeMarkers
+                                .DIFFERENT_PACKAGE_EXCEPTION);
```
aldexis: @mpritham Do you think it would make sense to have a `EndpointSpecificErrorsTypeMarkers.registerExceptions(ExceptionDeserializerArgs.Builder<T> args` method?  Among other things, that would allow us to
1. Reduce the duplicated code across the dialogue clients
2. Reduce the public API surface by hiding the type markers

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueServiceGenerator.java
```diff
@@ -86,14 +86,22 @@ public Stream<JavaFile> generate(ConjureDefinition conjureDefinition) {
...
                         endpoints,
                         interfaceGenerator,
                         blockingGenerator,
                         asyncGenerator));
+
+        if (options.generateErrorParameterFormatRespectingDialogueInterfaces()) {
+            Stream<JavaFile> typeMarkerFiles =
+                    new ErrorTypeMarkersGenerator(options).generate(conjureDefinition.getErrors());
+            return Stream.concat(serviceFiles, typeMarkerFiles);
+        }
```
aldexis, line 105: Would you mind adding a comment here repeating what you mention in your PR description around why we're putting this in dialogue as opposed to objects?

conjure-java-core/src/integrationInput/java/endpointerrors/com/palantir/product/ErrorServiceAsync.java
mpritham: Sounds good!

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/ErrorTypeMarkersGenerator.java
```diff
@@ -0,0 +1,124 @@
...
+    private JavaFile generateTypeMarkersClass(NamespacedErrors namespacedErrors) {
+        String conjurePackage = Packages.getPrefixedPackage(namespacedErrors.javaPackage(), options.packagePrefix());
+        String errorsClassName = ErrorGenerationUtils.errorTypesClassName(namespacedErrors.namespace());
+        String typeMarkersClassName = errorsClassName + "TypeMarkers";
+        ClassName errorsClass = ClassName.get(conjurePackage, errorsClassName);
+
+        TypeSpec.Builder classBuilder = TypeSpec.classBuilder(typeMarkersClassName)
+                .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
+                .addJavadoc(
+                        "Internal utility class used by generated Dialogue interfaces. Not intended for external use.")
```
aldexis: nit: Add a comment here to explain why we need to make this public, even when it's not intended for external use

aldexis (dismissed):

aldexis (approved):

PR #2847, opened 2026-03-23
Don't chain `.exception` calls when constructing exception deserializers

mpritham:
## Before this PR

When the `generateErrorParameterFormatRespectingDialogueInterfaces` is set, generated Dialogue interfaces are able to deserialize errors as exceptions. Currently, the exception deserializer is constructed as a single builder call, chaining a call to `.exception` for each error. 

https://github.com/palantir/conjure-java/blob/8ed34f38d374a747501b74e789476b5cf8a865aa/conjure-java-core/src/integrationInput/java/exceptionthrowingdialogueinterfaces/com/palantir/product/EteServiceBlocking.java#L454-L492

Some Conjure definitions have thousands of errors, leading to thousands of chained calls to `.exception`. This can cause the Java compiler stack to overflow: [1](https://issues.apache.org/jira/browse/MCOMPILER-325), [2](https://bugs.openjdk.org/browse/JDK-8177933?focusedId=14066535&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14066535)

## After this PR
==COMMIT_MSG==
Don't chain .exception calls when constructing exception deserializers
==COMMIT_MSG==

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
@@ -148,6 +148,10 @@ public final class DefaultStaticFactoryMethodGenerator implements StaticFactoryM
                 name);
     }
 
+    /**
+     * This method intentionally generates code where each call to the builder is not chained but a separate method call
+     * and assignment. Having many chained method calls causes the Java compiler stack to overflow.
+     */
     private MethodSpec createHelperToConstructExceptionDeserializerArgs() {
         TypeVariableName typeVariableT = TypeVariableName.get("T");
         ParameterizedTypeName builderType =
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
@@ -150,16 +150,20 @@ public final class DefaultStaticFactoryMethodGenerator implements StaticFactoryM
 
     private MethodSpec createHelperToConstructExceptionDeserializerArgs() {
         TypeVariableName typeVariableT = TypeVariableName.get("T");
+        ParameterizedTypeName builderType =
+                ParameterizedTypeName.get(ClassName.get(ExceptionDeserializerArgs.Builder.class), typeVariableT);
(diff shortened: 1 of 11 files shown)


conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -150,16 +150,20 @@ private ClassName getClassNameInErrorsPackage(ErrorDefinition errorDef, String n
...
-        CodeBlock.Builder exceptions = CodeBlock.builder()
-                .add("return $T.<$T>builder()", ExceptionDeserializerArgs.class, typeVariableT)
-                .add(".returnType(returnType)");
-        // Add exceptions from error definitions
+        CodeBlock.Builder code = CodeBlock.builder()
+                .addStatement(
+                        "$T builder = $T.<$T>builder().returnType(returnType)",
+                        builderType,
+                        ExceptionDeserializerArgs.class,
+                        typeVariableT);
```
aldexis, line 170: nit: Maybe add a comment explaining that we explicitly do not want to chain method calls to avoid stack overflows in the compiler (or was it actually at runtime in the jvm? I'm confused by your PR description mentioning the jvm)

aldexis (dismissed):

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
mpritham: Sorry, it is the compiler. Updated the comment and the PR desc.

mpritham:
Aside: there are 2 other places where we dynamically create builders:

1. `DialogueEndpointsGenerator#pathTemplateInitializer` chains `.fixed()`/ `.variable()` for
  each URL path segment. Unlikely to be a problem in practice since URL paths are short.

https://github.com/palantir/conjure-java/blob/09f5c30ca9f3f9d319f401a5ce097a0872d09b21/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DialogueEndpointsGenerator.java#L169-L191

2. `ErrorGenerator#generateSerializableError` chains `.putParameters()` for each error
  parameter. Also unlikely to blow the stack since individual errors typically have few parameters.

https://github.com/palantir/conjure-java/blob/09f5c30ca9f3f9d319f401a5ce097a0872d09b21/conjure-java-core/src/main/java/com/palantir/conjure/java/types/ErrorGenerator.java#L284-L288

aldexis (approved):

PR #2869, opened 2026-04-10
Deduplicate TypeMarker, ExceptionDeserializerArgs, and Deserializer fields in generated dialogue interfaces

schlosna:
## Before this PR

Generated dialogue service interfaces previously created per-endpoint TypeMarker
anonymous classes, ExceptionDeserializerArgs instances, and Deserializer fields,
causing excessive allocations when multiple endpoints share the same return type.

This adds unnecessary overhead for conjure clients that use sticky channels such as ⛰️ clients via e.g. `DataServiceAsync#of(com.palantir.dialogue.EndpointChannelFactory, com.palantir.dialogue.ConjureRuntime)`.

## After this PR
==COMMIT_MSG==
This change modifies DefaultStaticFactoryMethodGenerator to:

1. Generate one `private static final TypeMarker<T>` per distinct return type
   instead of `new TypeMarker<T>() {}` inline per endpoint.
2. Generate one `private static final ExceptionDeserializerArgs<T>` per distinct
   return type (error-respecting path), instead of calling
   createExceptionDeserializerArgs per endpoint.
3. Generate one shared `Deserializer<T>` instance field per distinct return type,
   referenced by all endpoints with the same return type.

For example EteServiceBlocking (36 endpoints, 22 distinct return types) this reduces
anonymous TypeMarker classes, createExceptionDeserializerArgs calls, and
Deserializer allocations from 36 each to 22 each.

==COMMIT_MSG==


conjure-java-core/src/integrationInput/java/dialogue/com/palantir/product/EmptyPathServiceAsync.java
```diff
@@ -28,18 +28,20 @@ public interface EmptyPathServiceAsync {
     /** Creates an asynchronous/non-blocking client for a EmptyPathService service. */
     static EmptyPathServiceAsync of(EndpointChannelFactory _endpointChannelFactory, ConjureRuntime _runtime) {
         return new EmptyPathServiceAsync() {
+            private static final TypeMarker<Boolean> booleanTypeMarker = new TypeMarker<Boolean>() {};
```
pkoenig10, line 31: I probably would prefer if these weren't static. It means that simply loading these classes will requiring initializing and retaining these classes, even if the client is no longer used.

mpritham:
Discussed internally that [this proposed Dialogue change](https://github.com/palantir/dialogue/pull/2647) is probably the more useful change to mitigate the overhead from using sticky clients. Hopefully that change will take the construction of these clients out of a hot path.

I think this is a good idea though. I think the hashes added to the variable names when there are name collisions hurts the readability: I wonder if there's another way to deconflict here.

schlosna:
> Discussed internally that [this proposed Dialogue change](https://github.com/palantir/dialogue/pull/2647) is probably the more useful change to mitigate the overhead from using sticky clients. Hopefully that change will take the construction of these clients out of a hot path.

Fixed up merge conflicts and finished out implementation of https://github.com/palantir/dialogue/pull/2647 in https://github.com/palantir/dialogue/pull/2955 on top of that branch

conjure-java-core/src/integrationInput/java/dialogue/com/palantir/product/EmptyPathServiceAsync.java
aldexis: @pkoenig10 What are your thoughts about making these be weak references instead? (which are reconstructed on the fly when necessary)
This may incur some extra allocations when the values have been GC'ed in between client creations (i.e. no clients are left, values are GC'ed, then we create a new client), but it should also alleviate the concern you raised.

With that said, I would tend to think that if an application is using a dialogue client, it will generally keep using some instances of that client over its whole lifecycle, and thus the retained classes and associated memory isn't really a huge deal

aldexis: Or is the concern about the anonymous classes themselves? I'm not entirely sure about the semantics of retaining classes, but I'd imagine the cost to be relatively small, and better than paying the initialization and allocation cost and every client creation? 🤔

pkoenig10: > With that said, I would tend to think that if an application is using a dialogue client, it will generally keep using some instances of that client over its whole lifecycle, and thus the retained classes and associated memory isn't really a huge deal

Yeah my concern here is a bit pedantic. Making these static fields is almost certainly not going to be a meaningful issue in practice.

Feel free to ignore my comment.

atrocities:
Tested with the dialogue generated packages for mp and premerge tested with a couple services - seems good. However, given the potentially laborious process of recalling generated dialogue modules across the board should there be any issues, we might want to hide this behind a feature flag and apply it to a couple repos for the time being, then change the default later. 

@mpritham - any thoughts?

aldexis:
A feature flag probably makes sense here imo

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -109,36 +115,195 @@ public MethodSpec generate(ServiceDefinition def) {
...
+    @SuppressWarnings("checkstyle:cyclomaticcomplexity")
+    private SerDeFieldsContext collectSerDeFields(ServiceDefinition def, boolean isErrorRespecting) {
+        Map<TypeName, String> fieldBaseNames = new HashMap<>();
+        Map<TypeName, String> typeMarkerFields = new LinkedHashMap<>();
+        Map<TypeName, String> exceptionArgsFields = new LinkedHashMap<>();
+        Map<TypeName, String> deserializerFieldNames = new LinkedHashMap<>();
+        Map<TypeName, DeserializerType> deserializerTypes = new LinkedHashMap<>();
+        Map<TypeName, String> serializerFieldNames = new LinkedHashMap<>();
+        Map<String, String> endpointToDeserializerField = new HashMap<>();
+        Map<String, String> endpointToSerializerField = new HashMap<>();
```
kunalrkak, line 156: Why are we keeping track of these mapping in a bunch of separate maps as opposed to using objects to encode the relationships here? i.e. a binding per type could own its marker, serializer, deserializer, etc., and could be referenced something keeping track of types at an endpoint level

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -179,6 +344,49 @@ private ClassName getClassName(ServiceDefinition def) {
...
+            sb.append(CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_CAMEL, rawName));
+            for (TypeName arg : parameterized.typeArguments()) {
+                String argBase = typeNameToFieldBase(arg);
+                sb.append(CaseFormat.LOWER_CAMEL.to(CaseFormat.UPPER_CAMEL, argBase));
+            }
+            return sb.toString();
+        } else if (typeName instanceof ClassName cn) {
+            return CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_CAMEL, cn.simpleName());
+        } else {
+            return "type" + CaseFormat.LOWER_CAMEL.to(CaseFormat.UPPER_CAMEL, deterministicHash(typeName));
```
kunalrkak: what are the cases where we'd hit this case and use the hash? could we add a test?

conjure-java-core/src/test/java/com/palantir/conjure/java/DialogueServiceGeneratorTests.java
```diff
@@ -67,6 +67,14 @@ public void testConjureImports() throws IOException {
         // Generated files contain imports
         assertThat(compiledFileContent(src, "test/api/with/imports/ImportServiceBlocking.java"))
                 .contains("import com.palantir.product.StringExample;");
+
+        String testService = compiledFileContent(src, "com/palantir/another/TestServiceBlocking.java");
```
kunalrkak: Could we write some tests to assert the special cases we're handling here, i.e. a type used as both req/resp, assertions for empty responses, naming conflicts between parameterized types & flattened names, etc?

Also why aren't we producing serializers for binary bodies?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -109,36 +115,195 @@ public MethodSpec generate(ServiceDefinition def) {
                 .initializer(CodeBlock.of("$L.plainSerDe()", StaticFactoryMethodGenerator.RUNTIME))
                 .build());
 
-        if (options.generateErrorParameterFormatRespectingDialogueInterfaces()) {
+        boolean isErrorRespecting = options.generateErrorParameterFormatRespectingDialogueInterfaces();
```
mpritham: nit:
```suggestion
        boolean isErrorParameterFormatRespecting = options.generateErrorParameterFormatRespectingDialogueInterfaces();
```

The flag controls whether Dialogue interfaces send the `Accept-Conjure-Error-Parameter-Format` header.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
```diff
@@ -109,36 +115,195 @@ public MethodSpec generate(ServiceDefinition def) {
...
                 .addParameter(ConjureRuntime.class, StaticFactoryMethodGenerator.RUNTIME)
                 .addCode(CodeBlock.builder().add("return $L;", impl.build()).build())
                 .build();
-        return method;
+    }
+
+    @SuppressWarnings("checkstyle:cyclomaticcomplexity")
+    private SerDeFieldsContext collectSerDeFields(ServiceDefinition def, boolean isErrorRespecting) {
+        Map<TypeName, String> fieldBaseNames = new HashMap<>();
+        Map<TypeName, String> typeMarkerFields = new LinkedHashMap<>();
```
mpritham, line 150: Why are some of these `LinkedHashMap` while others are `HashMap`? Do we need to use the ordering `LinkedHashMap` gives us?

mpritham: Oh I see, we do `forEach` on the entries with `LinkedHashMap` and we'd like the generation across runs to be the same. Shall we add a small comment?

conjure-java-core/src/test/java/com/palantir/conjure/java/DialogueServiceGeneratorTests.java
atrocities: Yeah, we can add some additional tests for the edge cases. 

The avoided serializers is for binary body _aliases_ is because we never actually end up using the TypeMarkers generated for them. The actual TypeMarker type ends up being InputStream, signaling to DefaultSerDe to use the BinaryDeserializer.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
atrocities: So, this catch-all case would apply to things like arrays, primitives, and anything else that's not a simple class nor a parameterized class... which would essentially be nothing that can be represented in a conjure yml. 

This feels like a kind of overly defensive case that should probably just throw an exception. Let me change that.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
atrocities: So, not being the author here, I don't have a strong opinion, but my feeling is that it's because all of the fields and names are fairly ephemeral, are only really used for the duration of codegen on a single class, aren't persisted, and don't get reused anywhere else. 

The additional amount of code needed to re-implement this might not be worth it. Do you feel particularly strongly about changing it?

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
atrocities: Yes, added a comment explaining it's this way for stable codegen.

conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java
kunalrkak: Yeah that's fair, given the scope I'm okay to leave it as is - I was coming more from the perspective that it'd be easier to understand what's going on for someone later on

kunalrkak (approved):

PR #2878, opened 2026-04-20
ConjureAuthorizationExtractor: exceptions thrown by a custom JsonWebTokenHandler are not thrown as `Conjure:MalformedCredentials`

mpritham:
## Before this PR

In `ConjureAuthorizationExtractor#cookie`, `setState` is called inside the try block that wraps `BearerToken` deserialization:

setState eventually calls the server-supplied `JsonWebTokenHandler#handle`, which can throw. This makes 401s from custom handlers misleading because they're not `Conjure:MalformedCredentials`.

The header path already behaves correctly (it calls setState outside any try/catch).

## After this PR

==COMMIT_MSG==
ConjureAuthorizationExtractor: exceptions thrown by a custom JsonWebTokenHandler are not thrown as `Conjure:MalformedCredentials`
==COMMIT_MSG==

### DIFF
--- a/conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AuthTest.java
+++ b/conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AuthTest.java
@@ -30,6 +30,7 @@ import io.undertow.server.HttpServerExchange;
 import io.undertow.server.handlers.CookieImpl;
 import io.undertow.util.Headers;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.jupiter.api.Test;
 
 public final class AuthTest {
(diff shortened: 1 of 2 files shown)


conjure-java-undertow-runtime/src/test/java/com/palantir/conjure/java/undertow/runtime/AuthTest.java
```diff
@@ -140,6 +140,23 @@ public void testJsonWebHandlerThrowsWhenAuthTokenSet() {
                 .hasType(error);
     }
 
+    @Test
+    public void testJsonWebHandlerThrowsWhenAuthCookieSet() {
```
bjlaub, line 145: should we add a similar test to verify that we see `MalformedCredentials` when the cookie auth path fails to parse a token first?

mpritham: added.

bjlaub (approved):
Thanks!

PR #2917, opened 2026-06-01
Redact `DoNotLog` annotated values in `toString` and introduce `dangerousToString`

mpritham:
## Before this PR

Third party libraries can call `toString` on objects when constructing exceptions, leading to `DoNotLog` annotated fields appearing in logs. Having `toString` not contain any `DoNotLog` data will prevent this class of error.

Conjure generates `toString` methods on objects containing `DoNotLog` field values. 

This PR adds the `redactToStringDoNotLog` flag (default off), that when enabled:
- Objects render `DoNotLog` fields as `<name>: REDACTED` in `toString` and contain a new method `dangerousToString` (annotated `DoNotLog`) that returns the full representation for the rare cases that need it.
- `DoNotLog` aliases return the bare literal `"REDACTED"` from toString.
  - Aliases do not need the `dangerousToString` method because they expose an equivalent method `get`
- Union variants whose member is `DoNotLog` redact the value in the wrapper's toString (both sealed and non-sealed paths)
  - Unions do not need the `dangerousToString` method because they expose an equivalent method `getValue`

There are certain cases where developers are using the `toString` value of Conjure objects with `DoNotLog` fields, so the flag is disabled by default, and will be rolled out repo by repo.

## After this PR

==COMMIT_MSG==
Redact `DoNotLog` annotated values in `toString` and introduce `dangerousToString`
==COMMIT_MSG==

## Possible downsides?

This introduces a new method for every generated object. We can't naively generate the `dangerousToString` only for objects with `DoNotLog` fields: a `DoNotLog` field could be removed from an object definition while a client calls the `dangerousToString` method which would lead to an ABI break. Solutions that try to take the previous state of a Conjure definition when generated code seem complicated and would be a divergence from Conjure's simple IR based code gen.


conjure-java-core/src/integrationInput/java/redactdonotlogobjects/com/palantir/product/ObjectAllDoNotLog.java
```diff
@@ -0,0 +1,156 @@
...
+            memoizedHashCode = result;
+        }
+        return result;
+    }
+
+    @Override
+    @DoNotLog
+    public String toString() {
+        return "ObjectAllDoNotLog{secret: REDACTED" + ", token: REDACTED" + '}';
+    }
```
aldexis, line 77: This is not in fact `DoNotLog` anymore

conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java
```diff
@@ -243,6 +243,15 @@ default boolean sealedUnionVisitors() {
...
 
+    /**
+     * If enabled, {@code toString} methods render {@code @DoNotLog}-annotated fields as "REDACTED" instead of including
+     * the actual value. A separate {@code dangerousToString} method (itself annotated {@code @DoNotLog}) is generated.
+     */
+    @Value.Default
+    default boolean redactToStringDoNotLog() {
+        return false;
+    }
+
```
aldexis, line 254: > There are certain cases where developers are using the toString value of Conjure objects with DoNotLog fields, so the flag is disabled by default, and will be rolled out repo by repo.

Since this will be generated on the producer, but may cause issues on the consumers, I don't think it's safe to simply turn on the flag? We may instead need to start always introducing the `dangerousToString` methods, so consumers can migrate across the board (and we can then write automation to migrate everything), _then_ get the flag turned on to also redact the main `toString`?

conjure-java-core/src/integrationInput/java/redactdonotlogsealedunions/com/palantir/product/UnionWithDoNotLogVariant.java
```diff
@@ -0,0 +1,322 @@
...
+
+        @Override
+        public int hashCode() {
+            return this.value.hashCode();
+        }
+
+        @Override
+        public String toString() {
+            return "UnionWithDoNotLogVariant{value: SecretValueWrapper{value: REDACTED" + "}}";
+        }
```
aldexis, line 164: I wonder if we should still provide a `dangerousToString` method for unions, because `value()` is no quite the same here, and I wonder if this is going to make migration more painful. We can also just not do anything for now and wait until we see migration issues, if any

conjure-java-core/src/integrationInput/java/redactdonotlogunions/com/palantir/product/UnionWithDoNotLogVariant.java
```diff
@@ -0,0 +1,344 @@
...
+
+        @Override
+        public int hashCode() {
+            return this.value.hashCode();
+        }
+
+        @Override
+        public String toString() {
+            return "SafeValueWrapper{value: " + value + '}';
+        }
```
aldexis, line 237: What is the replacement for this/the union's toString method? `getValue()` is private access, so we can't use it, and the visitor seems overkill 🤔

conjure-java-core/src/main/java/com/palantir/conjure/java/types/UnionGenerator.java
```diff
@@ -220,19 +228,19 @@ public static JavaFile generateUnionType(
...
      */
-    private static MethodSpec createLegacyToStringForSealedUnions(ClassName baseClassName, ClassName wrapperClassName) {
+    private static MethodSpec createLegacyToStringForSealedUnions(
+            ClassName baseClassName, ClassName wrapperClassName, boolean redactValue) {
+        String prefix = baseClassName.simpleName() + '{' + VALUE_FIELD_NAME + ": " + wrapperClassName.simpleName() + '{'
+                + VALUE_FIELD_NAME + ": ";
+        CodeBlock.Builder body = CodeBlock.builder().add("return $S\n", prefix + (redactValue ? "REDACTED" : ""));
+        if (!redactValue) {
+            body.add(" + $N", VALUE_FIELD_NAME);
+        }
```
aldexis, line 238: Won't this redact the _entire_ value if it's do-not-log, when we now actually have safe toString for conjure objects, which we could leverage? Probably good enough for a start, but seems like we may want to revisit that approach?

This actually makes me wonder whether the overall object itself should even be marked `DoNotLog` anymore (iirc, this mostly indicates the stringified log safety?)

PR #2922, opened 2026-06-04
Add RequestContext method to get source address

pkoenig10:
There are a number of existing use cases for wanting to know the source address of a request. For example, to include in audit logs.

Today we use `HttpServerExchange` directly for these code paths, but it would be nice to be able to use the read-only `ReqeustContext` instead. This information feels aligned with the purpose of `RequestContext` (to provide read-only information about a request) and seems reasonable to include here.

### DIFF
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -1,0 +1,6 @@
+acceptedBreaks:
+  "8.78.0":
+    com.palantir.conjure.java:conjure-undertow-lib:
+    - code: "java.method.addedToInterface"
+      new: "method java.net.InetSocketAddress com.palantir.conjure.java.undertow.lib.RequestContext::sourceAddress()"
+      justification: "Added method to interface that should not have external implementations"
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -2,5 +2,5 @@ acceptedBreaks:
   "8.78.0":
     com.palantir.conjure.java:conjure-undertow-lib:
     - code: "java.method.addedToInterface"
-      new: "method java.net.InetSocketAddress com.palantir.conjure.java.undertow.lib.RequestContext::sourceAddress()"
+      new: "method java.util.Optional<java.net.SocketAddress> com.palantir.conjure.java.undertow.lib.RequestContext::peerAddress()"
       justification: "Added method to interface that should not have external implementations"
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
@@ -29,6 +29,7 @@ import com.palantir.logsafe.logger.SafeLoggerFactory;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.server.handlers.Cookie;
 import io.undertow.util.HeaderValues;
+import java.net.InetSocketAddress;
 import java.security.cert.Certificate;
 import java.util.Collections;
 import java.util.Deque;
@@ -109,6 +110,11 @@ final class ConjureContexts implements Contexts {
         @Override
         public void requestArg(Arg<?> arg) {
             requestArgHandler.arg(exchange, arg);
+        }
+
+        @Override
+        public InetSocketAddress sourceAddress() {
+            return exchange.getSourceAddress();
         }
 
         @Override
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
@@ -29,7 +29,7 @@ import com.palantir.logsafe.logger.SafeLoggerFactory;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.server.handlers.Cookie;
 import io.undertow.util.HeaderValues;
-import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.security.cert.Certificate;
 import java.util.Collections;
 import java.util.Deque;
@@ -113,8 +113,8 @@ final class ConjureContexts implements Contexts {
         }
 
         @Override
-        public InetSocketAddress sourceAddress() {
-            return exchange.getSourceAddress();
+        public Optional<SocketAddress> peerAddress() {
+            return Optional.ofNullable(exchange.getConnection().getPeerAddress());
         }
 
         @Override
--- a/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
+++ b/conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
@@ -20,6 +20,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ListMultimap;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.Unsafe;
+import java.net.InetSocketAddress;
 import java.security.cert.Certificate;
 import java.util.List;
 import java.util.Optional;
@@ -66,6 +67,8 @@ public interface RequestContext {
      * Implementations may choose to include this data in the request log.
      */
     void requestArg(Arg<?> arg);
+
+    InetSocketAddress sourceAddress();
 
     /**
      * Returns the client certificates associated with the connection used to make the current request.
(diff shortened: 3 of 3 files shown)


.palantir/revapi.yml
```diff
@@ -0,0 +1,6 @@
+acceptedBreaks:
+  "8.78.0":
+    com.palantir.conjure.java:conjure-undertow-lib:
+    - code: "java.method.addedToInterface"
+      new: "method java.net.InetSocketAddress com.palantir.conjure.java.undertow.lib.RequestContext::sourceAddress()"
+      justification: "Added method to interface that should not have external implementations"
```
aldexis: @pkoenig10 Just for SA, we have ~20 internal repos that implement this interface (mostly to create fake/test contexts) https://pl.ntr/2BN

This will unfortunately prevent them from upgrading without breaks (I don't disagree with your statement that these should _not_ exist though)

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
```diff
@@ -67,6 +68,8 @@ public interface RequestContext {
      */
     void requestArg(Arg<?> arg);
 
+    InetSocketAddress sourceAddress();
```
aldexis: Could you add some javadoc to this? (every other method has it, and it feels a bit weird not to have one - one thing I would potentially specify is that this is the IP address, as opposed to the hostname?)

pkoenig10: Ah yup, sorry, I intended to add Javadoc here.

.palantir/revapi.yml
pkoenig10: Hmm that is unfortunate. Do you think this prevents us from merging this?

aldexis: How about making `sourceAddress()` return an `Optional` and default it to empty instead?

Reading through the `HttpServerExchange#getSourceAddress` code (and things it calls), it appears this _could_ return null (notably, if the underlying connection is a `org.xnio.nio.NioPipeStreamConnection`, though I don't think we'd hit that case internally).

We could also add a convenience method that returns the value or throws if absent (e.g. `maybeSourceAddress()` and `sourceAddress()` or something)

pkoenig10: > How about making sourceAddress() return an Optional and default it to empty instead?

Yeah this is objectively more correct. Not all socket connections have a source IP address - for example, Unix domain socket.

> We could also add a convenience method that returns the value or throws if absent (e.g. maybeSourceAddress() and sourceAddress() or something)

I would prefer not to do this. Consumers can easily handle absent optional values however they choose.

aldexis: > I would prefer not to do this. Consumers can easily handle absent optional values however they choose.

I agree - it was in case we were strongly opinionated that this should always exist

pkoenig10: Actually we probably want to return `Optional<SocketAddress>` here to correctly report the socket address for thing that are no IP-based socket addresses. But we'll still make it optional as it appears there are some implementations that return `null`.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
```diff
@@ -111,6 +112,11 @@ public void requestArg(Arg<?> arg) {
             requestArgHandler.arg(exchange, arg);
         }
 
+        @Override
+        public Optional<SocketAddress> sourceAddress() {
+            return Optional.ofNullable(exchange.getConnection().getPeerAddress());
```
aldexis, line 117: Undertow's `HttpServerExchange` seems to also support setting the `sourceAddress` directly, and `getSourceAddress()` actually does
```
    public InetSocketAddress getSourceAddress() {
        if (sourceAddress != null) {
            return sourceAddress;
        }
        return connection.getPeerAddress(InetSocketAddress.class);
    }
```
I'm unclear in which cases this gets set, but there seem to be a few places in Undertow that set this directly. What do you think should happen when/if these end up being different for some reason? 🤔

pkoenig10: The places where this gets set are:
1. [`ProxyPeerAddressHandler`](https://github.com/undertow-io/undertow/blob/f3da6fc173365dbe4d6f2cdc55362eaf3b90a7df/core/src/main/java/io/undertow/server/handlers/ProxyPeerAddressHandler.java), which sets the source addressed based on the [`X-Forwarded-For`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For) header.
1. [`ForwardedHandler`](https://github.com/undertow-io/undertow/blob/f3da6fc173365dbe4d6f2cdc55362eaf3b90a7df/core/src/main/java/io/undertow/server/handlers/ForwardedHandler.java), which sets the source addressed based on the [`Forwarded`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Forwarded) header.
1. [`PeerNameResolvingHandler`](https://github.com/undertow-io/undertow/blob/f3da6fc173365dbe4d6f2cdc55362eaf3b90a7df/core/src/main/java/io/undertow/server/handlers/PeerNameResolvingHandler.java), which can perform forward and/or reverse DNS lookups to provide fully resolved source addresses.

We don't use any of these handlers in this repo or any of our internal infrastructure libraries. So not considering the explicitly set `sourceAddresses` should not have a material impact on behavior.

Regarding (1) and (2), I think we explicitly do not want this layer to infer alternative source addresses based on information from the HTTP layer. For one, it muddies the semantics of this API. Second, consumers can always derive this information themselves from the header values in the `RequestContext`. But if we override the underlying source address with the values from these headers, then consumers have no way to obtain the actual peer address.

Regarding (3), we're almost certainly never going to force reverse DNS lookups on requests paths like this. Again, this is something that consumers can choose to do if they want - this API should provide the actual peer address from the connection.

The `HttpServerExchange` `sourceAddress` value is a sort of "HTTP layer" source address. But we want to provide the actual socket peer address in this API. We want to own the implementation for transforming an HTTP request into a "source address". And this may differ depending on the use case - some places we may want to respect the forwarded headers, other places we may not.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
```diff
@@ -111,6 +112,11 @@ public void requestArg(Arg<?> arg) {
             requestArgHandler.arg(exchange, arg);
         }
 
+        @Override
+        public Optional<SocketAddress> peerAddress() {
```
pkoenig10, line 116: Changed the name to `peerAddress` to be consistent with the Undertow API and with `peerCertificates` below.

conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
aldexis: This all makes sense to me, and I agree that in this case, what we want to expose is actually the underlying peerAddress 👍

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/RequestContext.java
```diff
@@ -67,6 +68,11 @@ public interface RequestContext {
      */
     void requestArg(Arg<?> arg);
 
+    /**
+     * Returns the socket address associated with the connection used to make the current request.
+     */
+    Optional<SocketAddress> peerAddress();
```
aldexis, line 74: Can you make this have a default implementation and return `Optional.empty()`? (and maybe add a comment indicating that this is for backwards compatibility to avoid breaking existing clients that implement this interface)

pkoenig10: Can we just require the handful of implementations to implement this? I would hope we're not exposing custom implementations in libraries, so the risk should be minimal.

I'm concerned about providing a default implementation that will return incorrect information.

aldexis: In this case, we're back to https://github.com/palantir/conjure-java/pull/2922#discussion_r3362332229. We'd need to go and fix the ~20 implementations by hand

(I agree that this is probably nicer)

pkoenig10: I can take responsibility for fixing those implementations.

aldexis: Sounds good in this case 👍 (and thanks for doing so)

aldexis (approved):

PR #2928, opened 2026-06-12
Add a handler for including standard endpoint request args

tom-s-powell:
## Before this PR
Request logs contain `method` and `path` and the combination of those can be used to identify the endpoint.

## After this PR
Request logs additional contain `_endpointName` and `_serviceName` for easier transparency into the Conjure endpoint that corresponds to this request. The intent is to make it easier to map these request logs back to the code and easier to filter logs on specific endpoints. 

I did also consider updating the code gen for the endpoints and leveraging the request context there instead. This felt simpler.

==COMMIT_MSG==
Add a handler for including standard endpoint request args.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
+++ b/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
@@ -28,6 +28,7 @@ import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.server.handlers.Cookie;
+import io.undertow.util.AttachmentKey;
 import io.undertow.util.HeaderValues;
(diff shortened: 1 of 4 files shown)


conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureContexts.java
```diff
@@ -49,7 +53,12 @@ final class ConjureContexts implements Contexts {
 
     @Override
     public RequestContext createContext(HttpServerExchange exchange, Endpoint _endpoint) {
-        return new ConjureServerRequestContext(exchange, requestArgHandler);
+        RequestContext context = exchange.getAttachment(ATTACHMENT);
```
tom-s-powell, line 56: Perhaps overkill but thought it was perhaps better to have a single `RequestContext` per exchange.

pkoenig10 (changes requested):
We already have a handler that does this in our internal server library. Custom request args are only added for non-successful responses.

PR #3029, opened 2026-08-21
Optimize generated Jackson deserialization

schlosna:
## Summary

- Add repository guidance for coding agents and Java stream conventions.
- Generate a type-first union deserialization fast path with mapper-local, lazy, thread-safe wrapper deserializer caching.
- Use prefix-only buffering for out-of-order union discriminators and a single discriminator switch.
- Populate generated builder maps directly during Jackson deserialization and declare Jackson Core dependencies explicitly.
- Regenerate the integration input fixtures.

## Performance

Midpoints from two JVM forks with seven measured trials per fork:

| Scenario | Baseline ns/op | Final ns/op | Baseline B/op | Final B/op |
|---|---:|---:|---:|---:|
| Union, type first | 508 | 314 | 1,552 | 1,076 |
| Union, out of order | 492 | 489 | 1,552 | 1,576 |
| Sealed union, type first | 446 | 319 | 1,536 | 1,040 |
| 101-variant union, type first | 480 | 328 | 1,552 | 1,056 |
| 101-variant union, out of order | 510 | 517 | 1,552 | 1,640 |
| 32-entry map | 3,019 | 2,474 | 8,224 | 6,552 |

Cold reader creation plus first read for the 101-variant union improved from about 490 microseconds and 134 KB to 300 microseconds and 22 KB. Serialization allocation remained unchanged; timing did not show a consistent shift.

The generated 101-variant deserializer decreased from 26,191 to 21,214 bytes compared with the initial reviewed implementation.

## Testing

- ./gradlew :conjure-java-core:test
- ./gradlew build -x :conjure-java-server-verifier:test

The server verifier test is excluded because the local native verifier requires unavailable libssl.so.1.1.

### DIFF
--- a/conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
+++ b/conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
@@ -6,14 +6,27 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSetter;
-import com.fasterxml.jackson.annotation.JsonSubTypes;
-import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.util.JsonParserSequence;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.deser.ResolvableDeserializer;
+import com.fasterxml.jackson.databind.util.TokenBuffer;
 import com.palantir.conjure.java.lib.SafeLong;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.Safe;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -25,17 +38,8 @@ import javax.annotation.Nullable;
 import javax.annotation.processing.Generated;
 
 @Generated("com.palantir.conjure.java.types.UnionGenerator")
-@JsonTypeInfo(
-        use = JsonTypeInfo.Id.NAME,
-        include = JsonTypeInfo.As.EXISTING_PROPERTY,
-        property = "type",
-        visible = true,
-        defaultImpl = SimpleUnion.Unknown.class)
-@JsonSubTypes({
-    @JsonSubTypes.Type(value = SimpleUnion.Foo.class, name = "foo"),
-    @JsonSubTypes.Type(value = SimpleUnion.Bar.class, name = "bar"),
-    @JsonSubTypes.Type(value = SimpleUnion.Baz.class, name = "baz")
-})
+@JsonDeserialize(using = SimpleUnion.Deserializer.class)
+@JsonSerialize(using = SimpleUnion.Serializer.class)
 @JsonIgnoreProperties(ignoreUnknown = true)
 public abstract sealed class SimpleUnion
         permits SimpleUnion.Foo, SimpleUnion.Bar, SimpleUnion.Baz, SimpleUnion.Unknown {
@@ -81,6 +85,9 @@ public abstract sealed class SimpleUnion
     public sealed interface Known permits Foo, Bar, Baz {}
 
     @JsonTypeName("foo")
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    @JsonDeserialize
+    @JsonSerialize
     public static final class Foo extends SimpleUnion implements Known {
         private final String value;
 
@@ -126,6 +133,9 @@ public abstract sealed class SimpleUnion
     }
 
     @JsonTypeName("bar")
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    @JsonDeserialize
+    @JsonSerialize
     public static final class Bar extends SimpleUnion implements Known {
         private final int value;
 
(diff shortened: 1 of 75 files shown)


conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
```diff
@@ -275,6 +290,138 @@ public String toString() {
...
 
+    static final class Serializer extends JsonSerializer<SimpleUnion> {
+        @Override
+        public void serialize(SimpleUnion value, JsonGenerator generator, SerializerProvider serializers)
+                throws IOException {
+            serializers.findValueSerializer(value.getClass()).serialize(value, generator, serializers);
+        }
+    }
+
+    static final class Deserializer extends JsonDeserializer<SimpleUnion> implements ResolvableDeserializer {
```
kunalrkak: what benefit does `ResolvableDeserializer` give us here?

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
```diff
@@ -275,6 +290,138 @@ public String toString() {
...
+        private static final Class<?>[] VARIANT_TYPES = new Class<?>[] {Foo.class, Bar.class, Baz.class};
+
+        private volatile JsonDeserializer<?>[] deserializers;
```
kunalrkak: nit:
```suggestion
        private volatile JsonDeserializer<?>[] variantDeserializers;
```

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
```diff
@@ -275,6 +290,138 @@ public String toString() {
         }
     }
 
+    static final class Serializer extends JsonSerializer<SimpleUnion> {
```
kunalrkak, line 293: is this different than what Jackson does out of the box? i.e. why does this need to be custom

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
```diff
@@ -81,6 +85,9 @@ public Known throwOnUnknown() {
     public sealed interface Known permits Foo, Bar, Baz {}
 
     @JsonTypeName("foo")
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    @JsonDeserialize
+    @JsonSerialize
```
kunalrkak, line 90: why do we need these extra annotations on each subvariant?

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
```diff
@@ -275,6 +290,138 @@ public String toString() {
...
+                    return context.reportInputMismatch(
+                            SimpleUnion.class, "Expected the end of a JSON object while deserializing a union");
+                }
+            }
+            return context.reportInputMismatch(SimpleUnion.class, "Union discriminator 'type' is required");
+        }
+
+        private static boolean isTypeField(String fieldName, DeserializationContext context) {
+            return "type".equals(fieldName)
+                    || (context.isEnabled(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES)
```
kunalrkak: nit: given this flag is constant can we look it up once and pass it into this method?

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
```diff
@@ -275,6 +290,138 @@ public String toString() {
...
+        }
+
+        private synchronized JsonDeserializer<?> resolveDeserializer(DeserializationContext context, int variantIndex)
+                throws JsonMappingException {
+            JsonDeserializer<?> deserializer = deserializers[variantIndex];
+            if (deserializer == null) {
+                deserializer = context.findRootValueDeserializer(context.constructType(VARIANT_TYPES[variantIndex]));
+                JsonDeserializer<?>[] updated = deserializers.clone();
+                updated[variantIndex] = deserializer;
+                deserializers = updated;
```
kunalrkak: why are we handling for a race condition here (shouldn't `findRootValueDeserializer` be idempotent?)

kunalrkak: also why this particular method? I'm comparing it to the jackson code for polymorphic deser and don't see it used, but do see `findContextualValueDeserializer` used here:
https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.21.1/src/main/java/com/fasterxml/jackson/databind/jsontype/impl/AsPropertyTypeDeserializer.java#L217

kunalrkak (commented):
commented on the generated code itself, ignored the actual codgen piece for now

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
schlosna: Removed ResolvableDeserializer. It only initialized the variant cache, so the cache is now initialized directly when the generated deserializer is constructed.

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
schlosna: Done: renamed this to variantDeserializers.

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
schlosna: Jackson discovers the union-level custom serializer and deserializer through the concrete variant supertypes. The empty annotations mask those inherited handlers; otherwise the runtime dispatch would select the union handler again and recurse. They are only needed for sealed variants because those variants extend the annotated union base.

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
schlosna: Removed the synchronized double-check and clone. The generated cache now uses AtomicReferenceArray: concurrent misses may perform the idempotent Jackson lookup, while compareAndSet safely publishes the cached result without serializing unrelated variant lookups.

schlosna: Changed this to findContextualValueDeserializer, matching Jackson's polymorphic deserializer lookup. The generated wrapper dispatch has no containing BeanProperty to propagate, so it passes null for the property.

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
schlosna: Done. The mapper feature is now read once per deserialize call and passed to both the type-first and buffered discriminator checks. I added coverage for both paths.

conjure-java-core/src/integrationInput/java/sealedunions/com/palantir/product/SimpleUnion.java
schlosna: This differs for statically typed serialization, such as writerFor(SimpleUnion.class). Jackson cannot derive the concrete variant serializer from the abstract union type in that case, so this serializer performs one runtime-type dispatch. The existing writerFor tests cover both known and unknown sealed variants.

PR #3034, opened 2026-08-26
[dnm] Prototype for ofType and ofGeneric TypeMarkers

atrocities:
Generated Conjure code creates anon TypeMarker subclasses liberally. This is done in order to capture generic type information for parameterized classes. Since type erasure prevents directly passing in generic type information as a constructor argument for TypeMarker, the full paramaterized type must come from a subtype, and obtained with getGenericSuperclass(). 

As far as I can tell, the anon TypeMarker subclasses are also generated for non-paramaterized types for the sake of uniformity.

The major drawback is that these TypeMarker classes pollute the class space, and end up consuming material amounts of metaspace. 

Introduce an ofGeneric factory for TypeMarker that accepts the generic type arguments for the paramaterized type as a list, and use the of factory method for TypeMarkers that encapsulate non-paramaterized types. It is, however, not actually typesafe at compile time.


atrocities:
Closing in favor of #3035

PR #3035, opened 2026-08-26
Eliminate TypeMarkers for simple and one level parameterized types

atrocities:
## Before this PR
Anon TypeMarker implementations were created for all types that need to be serialized/deserialized. This results in pollution of jars at rest, and meta/symbol/class space at runtime.

The reason for this is because of the erasure of the parameterized type arguments for generic types, which makes it not possible to express something like List<String>.class as an argument to a factory method. The workaround was to capture this type information in a subclass instead. While this approach was not strictly required for simple, non parameterized types, it was chosen due to uniformity.

## After this PR
Add 5 factory methods to TypeMarker and an implementation of ParameterizedType for simple types and the 4 base parameterized types that conjure generates. This reduces the need for most TypeMarkers, while also being reasonably readable, and type safe, so that no casting of TypeMarkers is needed by the caller.

TypeMarkers are still produced for parameterized types that have more than one level of parameterization. In practice, this comprises a vanishingly small number of types. While conjure-java does not generate any parameterized classes outside of Set, List, Map, and Optional, it's possible that more types may be added in the future, for which anon TypeMarkers will be generated for those types unless this collection of factory methods is also updated.

## Possible downsides?
This only applies to the Undertow side. Changes to Dialogue will need to be made separately.

### DIFF
--- a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
+++ b/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
@@ -287,10 +287,10 @@ final class UndertowServiceHandlerGenerator {
                             FieldSpec.builder(type, DESERIALIZER_VAR_NAME, Modifier.PRIVATE, Modifier.FINAL)
                                     .build());
                     ctorBuilder.addStatement(
-                            "this.$1N = $2N.bodySerDe().deserializer(new $3T() {}, this)",
+                            "this.$1N = $2N.bodySerDe().deserializer($3L, this)",
                             DESERIALIZER_VAR_NAME,
                             RUNTIME_VAR_NAME,
-                            ParameterizedTypeName.get(ClassName.get(TypeMarker.class), typeName));
+                            typeMarker(typeName));
                 });
 
         endpointDefinition.getReturns().ifPresent(returnType -> {
@@ -301,10 +301,10 @@ final class UndertowServiceHandlerGenerator {
                 endpointBuilder.addField(FieldSpec.builder(type, SERIALIZER_VAR_NAME, Modifier.PRIVATE, Modifier.FINAL)
                         .build());
                 ctorBuilder.addStatement(
-                        "this.$1N = $2N.bodySerDe().serializer(new $3T() {}, this)",
+                        "this.$1N = $2N.bodySerDe().serializer($3L, this)",
                         SERIALIZER_VAR_NAME,
                         RUNTIME_VAR_NAME,
-                        ParameterizedTypeName.get(ClassName.get(TypeMarker.class), typeName));
+                        typeMarker(typeName));
             }
         });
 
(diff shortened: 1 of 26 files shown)


conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
```diff
@@ -813,6 +811,43 @@ private static Optional<CodeBlock> getSafeLogging(
         };
     }
 
+    private static final ClassName MAP_NAME = ClassName.get(Map.class);
```
kunalrkak, line 814: is there no shared location we can put all the below code to avoid the code duplication?

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/TypeMarker.java
```diff
@@ -81,12 +87,95 @@ public final String toString() {
...
-        return new WrappingTypeMarker(type);
+        return new WrappingTypeMarker<>(type);
     }
 
-    private static final class WrappingTypeMarker extends TypeMarker<Object> {
+    public static <T> TypeMarker<T> of(Class<T> type) {
+        return new WrappingTypeMarker<>(type);
+    }
+
+    public static <T> TypeMarker<List<T>> listOf(Class<T> elementType) {
```
kunalrkak, line 97: could we add some simple equality tests for the new parameterized wrapped type factories? i.e.
```
assertThat(TypeMarker.listOf(String.class)).isEqualTo(new TypeMarker<List<String>>() {});
```
would be good to also confirm equal hashcodes and equivalent values from `getType()` so we know there won't be any issues replacing the existing anon TypeMarkers

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
```diff
@@ -404,6 +404,35 @@ private static void addTags(EndpointDefinition endpointDefinition, TypeSpec.Buil
...
+                    result.add(", ");
+                }
+                result.add("$T.class", parameterized.typeArguments().get(index));
+            }
+            return result.add(")").build();
+        }
+        return CodeBlock.of("new $T<$T>() {}", TypeMarker.class, type);
+    }
+
+    private static boolean hasExpectedSimpleTypeArguments(ParameterizedTypeName type) {
```
kunalrkak, line 431: I couldn't find one while poking, can we also add a case confirming this behavior leaves "complex" wrapped types as is? I see some examples use `ImmutableList` which would presumably fail at `TYPE_MARKER_FACTORIES#containsKey`.

kunalrkak (commented):
overall lgtm, few nits

conjure-undertow-lib/src/main/java/com/palantir/conjure/java/undertow/lib/TypeMarker.java
atrocities: Hmm, I thought I did add a test for this case, but it turns out that was in dialogue (https://github.com/palantir/dialogue/pull/3172/changes#diff-cb25112e7cfce5ab658e57af4dc192e2860ec203e757d96c3aaf773953b65519R29). Will add!

conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java
atrocities: Will add a case for this.

conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/generate/ConjureUndertowEndpointsGenerator.java
atrocities: Yeah, the code is identical between the endpoint and handler generators. The problem is that one of these is an annotation processor whereas the other one is in codegen, and that this little bit of code happens to need javapoet. 

Pulling this code up into conjure-undertow-lib or conjure-lib would introduce javapoet, which is only used for code generation, as a dependency, and that isn't great because these happen to be used at service runtime. 

That leaves the option of making a new module with a single util class that contains the typemarker generation. Let me think about that. It would likely be part of a broader deduplication refactor.

atrocities: Ended up doing the option where we're just pulling it up into a new module.

kunalrkak (approved):
Thanks!

PR #3064, opened 2026-09-16
Add comparator for uuid type

ericwang1409:
## Before this PR
UUID did not have a compareTo method, resulting in clients creating custom Comparator objects for Conjure objects that are aliases of UUID that delegate to UUID.compareTo anyways. 

## After this PR

==COMMIT_MSG==
Add a compareTo for aliases of UUID so devs can avoid writing these helpers themselves.
==COMMIT_MSG==

## Possible downsides?

There is a documented [bug](https://bugs.openjdk.org/browse/JDK-7025832) with UUID's comparator, but it doesn't seem like they will be changing this anytime soon so this should still be used.


conjure-java-core/src/test/java/com/palantir/conjure/java/types/AliasTests.java
```diff
@@ -156,4 +157,15 @@ void aliasWithDefensiveCollectionsIsImmutable_map() {
         Map<String, Boolean> internal = alias.get();
         assertThatExceptionOfType(UnsupportedOperationException.class).isThrownBy(() -> internal.put("bar", true));
     }
+
+    @Test
+    public void testUuidAliasComparison() {
```
mpritham, line 162: Should we add a test exercising [this bug](https://bugs.openjdk.org/browse/JDK-7025832) just to document it?

conjure-java-core/src/test/java/com/palantir/conjure/java/types/AliasTests.java
```diff
@@ -156,4 +157,15 @@ void aliasWithDefensiveCollectionsIsImmutable_map() {
...
+        UuidAliasExample lower = UuidAliasExample.of(new UUID(0L, 1L));
+        UuidAliasExample higher = UuidAliasExample.of(new UUID(0L, 2L));
+        UuidAliasExample equalToLower = UuidAliasExample.of(new UUID(0L, 1L));
+
+        assertThat(lower.compareTo(higher)).isNegative();
```
mpritham, line 167: Should we add a test for aliases of aliases? I see this is used in a few places.

ericwang1409: We can make aliases of aliases but aliases of aliases do not have a comparator

aldexis: Thoughts on adding comparators to aliases of aliases in another PR in this case?

mpritham (approved):
lgtm. Thanks!

PR #3065, opened 2026-09-18
Add support for comparing reference types that resolve to base comparable types

ericwang1409:
## Before this PR
Aliases of aliases did not implement Comparable or have a compareTo method, even when their underlying type supported comparison (i.e. int, doubles, etc.). 

## After this PR

==COMMIT_MSG==
References types that resolve to base comparable types now have a compareTo function. Same reference types are able to be compared.   Comparisons delegate to the wrapped alias and require both values to have the same alias type. For example, AliasOfUuidAlias can now be compared with another AliasOfUuidAlias. This supports multiple levels of alias nesting.
==COMMIT_MSG==

## Possible downsides?
No known downsides. Comparisons preserve the underlying type’s existing ordering.


conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -412,6 +412,23 @@ && hasValueOfFactory(reference.getExternalReference())
...
+        // If it resolves down to a primitive, return the comparable function. If the comparable function is present
+        // after the recursion, ignore the return value and then add compareTo
+        if (conjureType.accept(MoreVisitors.IS_INTERNAL_REFERENCE)) {
+            return typeMapper
+                    .getType(conjureType.accept(TypeVisitor.REFERENCE))
+                    .filter(type -> type.accept(TypeDefinitionVisitor.IS_ALIAS))
+                    .map(type -> type.accept(TypeDefinitionVisitor.ALIAS))
+                    .flatMap(type -> referenceToComparable(
+                                    typeMapper.getClassName(type.getAlias()), typeMapper, type.getAlias())
+                            .map(ignored -> createCompareTo(aliasName)));
```
aldexis: Is this so we only add comparable if the underlying type itself is comparable?

Could we add an integration test for the negative case as well? (where it's not comparable in the end) I see you have a unit test for this, but feel like it would be nice to have a test covering that for integration tests as well

conjure-java-core/src/main/java/com/palantir/conjure/java/types/AliasGenerator.java
```diff
@@ -412,6 +412,23 @@ && hasValueOfFactory(reference.getExternalReference())
         return Optional.empty();
     }
 
+    private static Optional<MethodSpec> referenceToComparable(
```
aldexis: nit: Since it's only used below, it might be nicer to move it into the ComparableVisitor class?

conjure-java-core/src/test/resources/example-types.yml
```diff
@@ -143,6 +145,8 @@ types:
         alias: bearertoken
       UuidAliasExample:
         alias: uuid
+      AliasOfUuidAliasExample:
```
mpritham, line 152: Let's also add a test that is another level (or a few more levels) deeper? I think we don't need to worry about any recursive types here because those should be caught when parsing the Conjure def.

mpritham: I think let's do `AliasOfAliasOfDouble` or something where the comparator is not just `compareTo` but a different method like `Double.compare`

